PR #1, opened 2016-05-31
Support for enumerating key handles

kevinawalsh:
This is useful for debugging and is needed for graceful cleanup or recovery of TPM resources. Specifically, some TPMs seem to run out of key handles, causing LoadKey to fail, and the key handles persist even across reboots. Apparently, the only way to clear them short of resetting the TPM is to enumerate and explicitly unload old keys.

### DIFF
--- a/tpm/encoding.go
+++ b/tpm/encoding.go
@@ -140,6 +140,25 @@ func packType(buf io.Writer, elts []interface{}) error {
 	return nil
 }
 
+func unpackKeyHandleList(b []byte) ([]Handle, error) {
+	// TODO(kwalsh): handle pack/unpack of TPM_KEY_HANDLE_LIST more gracefully
+	buf := bytes.NewBuffer(b)
+	var n uint16
+	if err := unpackType(buf, []interface{}{&n}); err != nil {
+		return nil, err
+	}
+	if n == 0 {
+		return nil, nil
+	}
+	h := make([]Handle, int(n))
+	for i, _ := range h {
+		if err := unpackType(buf, []interface{}{&h[i]}); err != nil {
+			return nil, err
+		}
+	}
+	return h, nil
+}
+
 // unpack performs the inverse operation from pack.
 func unpack(b []byte, elts []interface{}) error {
 	buf := bytes.NewBuffer(b)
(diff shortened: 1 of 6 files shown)


kevinawalsh:
I signed it!

tpm/encoding.go
```diff
@@ -140,6 +140,25 @@ func packType(buf io.Writer, elts []interface{}) error {
...
+	// TODO(kwalsh): handle pack/unpack of TPM_KEY_HANDLE_LIST more gracefully
+	buf := bytes.NewBuffer(b)
+	var n uint16
+	if err := unpackType(buf, []interface{}{&n}); err != nil {
+		return nil, err
+	}
+	if n == 0 {
+		return nil, nil
+	}
+	h := make([]Handle, int(n))
```
tmroeder: I believe that make accepts any integer type in its size argument, so I don't think you need to cast to int here.

tpm/tpm.go
```diff
@@ -61,6 +61,21 @@ func OpenTPM(path string) (io.ReadWriteCloser, error) {
 	return rwc, nil
 }
 
+// GetKeys gets the list of handles for currently-loaded TPM keys.
+func GetKeys(f *os.File) ([]Handle, error) {
```
tmroeder: In the other functions in this file, I've been using the more generalized io.ReadWriter interface instead of *os.File. This makes it easier to build a fake TPM for testing.

tpm/encoding.go
```diff
@@ -140,6 +140,25 @@ func packType(buf io.Writer, elts []interface{}) error {
 	return nil
 }
 
+func unpackKeyHandleList(b []byte) ([]Handle, error) {
```
tmroeder: Please add a test in encoding_test.go for the new encoding function.

tpm/tpm.go
tmroeder: Please also add an example/test in tpm_test.go for GetKeys. It will Skip() if openTPMOrSkip can't reach the TPM.

kevinawalsh:
I think the latest commits address all of your concerns Tom.

PR #3, opened 2018-01-27
Add methods for generating keys and doing sign operation.

JackOfMostTrades:
Add and example for generating a keyblob, getting its public key, generating a signature, and verifying it. Add method for reseting PCRs so that the tpm-sign example can demonstrate binding to PCR values.

### DIFF
--- a/examples/tpm-sign/extend_pcr.go
+++ b/examples/tpm-sign/extend_pcr.go
@@ -1,0 +1,62 @@
+// Copyright (c) 2018, Ian Haken. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"crypto/sha1"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/google/go-tpm/tpm"
+)
+
+func extendPcrAction() {
+	var tpmname = flag.String("tpm", "/dev/tpm0", "The path to the TPM device to use")
+	var pcrNum = flag.Int("pcr", 16, "PCR number to extend")
+	var reset = flag.Bool("reset", false, "Reset the PCR rather than extending it")
+	var dataPath = flag.String("data", "", "Path to the data that will be used to extend the PCR. If empty or omitted, the data will be read from stdin.")
+	flag.CommandLine.Parse(os.Args[2:])
+
+	rwc, err := tpm.OpenTPM(*tpmname)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Couldn't open the TPM file %s: %s\n", *tpmname, err)
+		return
+	}
+	defer rwc.Close()
+
+	if *reset {
+		if err = tpm.PcrReset(rwc, []int{*pcrNum}); err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to reset PCR: %s\n", err)
+			return
+		}
+	} else {
+		var data []byte
+		if *dataPath == "" {
+			data, err = ioutil.ReadAll(os.Stdin)
+		} else {
+			data, err = ioutil.ReadFile(*dataPath)
+		}
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to read input: %s\n", err)
+			return
+		}
+		if _, err = tpm.PcrExtend(rwc, uint32(*pcrNum), sha1.Sum(data)); err != nil {
+			fmt.Fprintf(os.Stderr, "Error extending PCR: %s\n", err)
+			return
+		}
+	}
+}
--- a/examples/tpm-sign/generate.go
+++ b/examples/tpm-sign/generate.go
@@ -31,7 +31,7 @@ func generateAction() {
 	var tpmname = flag.String("tpm", "/dev/tpm0", "The path to the TPM device to use")
 	var keyblobPath = flag.String("keyblob", "keyblob", "Output path of the generated keyblob")
 	var pubKeyPath = flag.String("publicKey", "publickey", "Output path of the generated keyblob's public key")
-	var pcrsStr = flag.String("pcrs", "", "A comma-separated list of PCR numbers againt which the generated key will be bound. If blank, it will not be bound to any PCR values.")
+	var pcrsStr = flag.String("pcrs", "", "A comma-separated list of PCR numbers against which the generated key will be bound. If blank, it will not be bound to any PCR values.")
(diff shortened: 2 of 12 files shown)


tpm/commands.go
```diff
@@ -324,3 +324,38 @@ func takeOwnership(rw io.ReadWriter, encOwnerAuth []byte, encSRKAuth []byte, srk
 
 	return &k, &ra, ret, nil
 }
+
+// Creates a wrapped key under the SRK
```
tmroeder: nit: please end comments with punctuation.

examples/tpm-sign/tpm-sign.go
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2014, Google Inc. All rights reserved.
```
tmroeder: Generally, the style here would be copyright Whoever with the year 2018, followed by the Apache license, IIUC.

examples/tpm-sign/tpm-sign.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+var (
+	ownerAuthEnvVar     = "TPM_OWNER_AUTH"
+	srkAuthEnvVar       = "TPM_SRK_AUTH"
+	usageAuthEnvVar     = "TPM_USAGE_AUTH"
+	migrationAuthEnvVar = "TPM_MIGRATION_AUTH"
+)
+
+func main() {
```
tmroeder: This looks very similar to other example files. I'd suggest doing a git cp to make it clear how this is the same and how it differs from other examples. It makes the history cleaner that way and makes the review easier.

examples/tpm-sign/tpm-sign.go
```diff
@@ -0,0 +1,148 @@
...
+	fmt.Printf("Public Key: %s\n", base64.StdEncoding.EncodeToString(pubKeyBytes))
+
+	keyHandle, err := tpm.LoadKey2(rwc, keyblob, srkAuth[:])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Could not load keyblob: %s\n", err)
+		return
+	}
+	defer keyHandle.CloseKey(rwc)
+
+	hashed := sha256.Sum256([]byte("Hello, World!"))
```
tmroeder: Instead of using fixed strings here, how about taking in a file or a string to sign? That would make this a useful utility in addition to being a useful example. Same comment about extending the PCR.

examples/tpm-sign/tpm-sign.go
```diff
@@ -0,0 +1,148 @@
...
+	fmt.Printf("Signature: %s\n", base64.StdEncoding.EncodeToString(signature))
+
+	err = rsa.VerifyPKCS1v15(pubKey, crypto.SHA256, hashed[:], signature)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error from verification: %s\n", err)
+		return
+	}
+	fmt.Printf("Signature valid.\n")
+
+	// Extend PCR 16
```
tmroeder: nit: please end comments with punctuation.

examples/tpm-sign/tpm-sign.go
```diff
@@ -0,0 +1,148 @@
...
+	_, err = tpm.PcrExtend(rwc, 16, sha1.Sum([]byte("xyz")))
+	if err != nil {
```
tmroeder: It would be a little more idiomatic Go in this code to merge lines that don't create variables (i.e., that use "=" instead of ":=") with their following if statments. E.g., in this case

```go
if _, err = tpm.PcrExtend(rwc, 16, sha1.Sum([]byte("xyz"))); err != nil {
```

Please change that here and elsewhere in this file.

tpm/tpm.go
```diff
@@ -872,3 +872,157 @@ func TakeOwnership(rw io.ReadWriter, newOwnerAuth digest, newSRKAuth digest, pub
...
+		return nil, err
+	}
+
+	k, ra, ret, err := createWrapKey(rw, encUsageAuth, encMigrationAuth, keyInfo, ca)
+	if err != nil {
+		return nil, err
+	}
+
+	raIn := []interface{}{ret, ordCreateWrapKey, k}
+	err = ra.verify(ca.NonceOdd, sharedSecret[:], raIn)
```
tmroeder: Here's another opportunity to merge an assignment with the following test

JackOfMostTrades:
@tmroeder Thanks for the feedback! I believe I've fixed everything you suggested. As part of cleaning up the tpm-sign example I made it in to a more general utility, so please take a look at the new README in the tpm-sign example and let me know if you have any more suggestions.

examples/tpm-sign/tpm-sign.go
```diff
@@ -0,0 +1,42 @@
...
+
+func showUsage() {
+	fmt.Println("Usage: ./tpm-sign <sign|verify|generate|extendPcr> [...]")
+}
+
+func main() {
+	if len(os.Args) < 2 {
+		showUsage()
+	} else {
+		if os.Args[1] == "sign" {
```
tmroeder: [optional]: I think this would be more idiomatic in Go as a switch, but it's OK like this, too.

tpm/tpm.go
```diff
@@ -872,3 +872,156 @@ func TakeOwnership(rw io.ReadWriter, newOwnerAuth digest, newSRKAuth digest, pub
 	raIn := []interface{}{ret, ordTakeOwnership, k}
 	return ra.verify(ca.NonceOdd, newOwnerAuth[:], raIn)
 }
+
+// CreateWrapKey creates a new RSA key for signatures inside the TPM. It is wrapped by the SRK (which is to say,
```
tmroeder: nit: The style in this repo is to wrap comments to 80 columns.

JackOfMostTrades: :+1:

examples/tpm-sign/tpm-sign.go
JackOfMostTrades: Yep, better as a switch. Updated.

examples/tpm-sign/generate.go
```diff
@@ -0,0 +1,105 @@
...
+	"strings"
+
+	"github.com/google/go-tpm/tpm"
+)
+
+func generateAction() {
+	var tpmname = flag.String("tpm", "/dev/tpm0", "The path to the TPM device to use")
+	var keyblobPath = flag.String("keyblob", "keyblob", "Output path of the generated keyblob")
+	var pubKeyPath = flag.String("publicKey", "publickey", "Output path of the generated keyblob's public key")
+	var pcrsStr = flag.String("pcrs", "", "A comma-separated list of PCR numbers againt which the generated key will be bound. If blank, it will not be bound to any PCR values.")
```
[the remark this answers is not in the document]
JackOfMostTrades: :+1:

examples/tpm-sign/extend_pcr.go
```diff
@@ -0,0 +1,62 @@
...
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/google/go-tpm/tpm"
+)
+
+func extendPcrAction() {
+	var tpmname = flag.String("tpm", "/dev/tpm0", "The path to the TPM device to use")
```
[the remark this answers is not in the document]
JackOfMostTrades: In this case the lower-case `tpmname` is copied from the other examples, so I've left it as-is for consistency. But of course there's no trouble changing it if you want the proper style for new code.

PR #4, opened 2018-02-01
tpm2 files

jlmucb:


tmroeder:
Please submit this as a PR to https://github.com/awly/go-tpm rather than here, since awly is going to work on these files before submitting a PR to this repo.

PR #5, opened 2018-02-06
TPM 2.0 partial implementation

awly:
Contains code from https://github.com/jlmucb/cloudproxy/tree/master/go/tpm2 plus refactoring.

This is NOT a full TPM 2.0 spec implementation. Only a select handful of commands are implemented, along with their required constants and data structures.
Most of the constant groups are not the full set of constants from that category in the spec.
Certain commands are not implemented with their full potential flexibility (e.g. TPM2_Create only creates RSA keys).


tpm2/constants.go
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) 2014, Google Inc. All rights reserved.
```
tmroeder: 2018 now

tpm2/constants.go
```diff
@@ -0,0 +1,181 @@
...
+	AlgECDH      Algorithm = 0x0019
+	AlgECDAA     Algorithm = 0x001A
+	AlgECC       Algorithm = 0x0023
+	AlgCTR       Algorithm = 0x0040
+	AlgOFB       Algorithm = 0x0041
+	AlgCBC       Algorithm = 0x0042
+	AlgCFB       Algorithm = 0x0043
+	AlgECB       Algorithm = 0x0044
+	AlgLAST      Algorithm = 0x0044
+	AlgKEYEDHASH Algorithm = 0x0008
```
tmroeder: Out of order. I'd suggest moving this up to just after `AlgAES`.

Also, for the Alg* entries that are not acronyms, Go style prefers a different casing. I think this would be `AlgKeyedHash` and `AlgLast`.

tpm2/constants.go
```diff
@@ -0,0 +1,181 @@
...
+	HandleTransport       Handle = 0x40000003
+	HandleOperator        Handle = 0x40000004
+	HandleAdmin           Handle = 0x40000005
+	HandleEK              Handle = 0x40000006
+	HandleNull            Handle = 0x40000007
+	HandleUnassigned      Handle = 0x40000008
+	HandleLockout         Handle = 0x4000000A
+	HandleEndorsement     Handle = 0x4000000B
+	HandlePlatform        Handle = 0x4000000C
+	PasswordSessionHandle Handle = 0x40000009
```
tmroeder: Out of order and a changed naming scheme. I think this should come after `HandleUnassigned`. Also, please consider using [`iota`](https://golang.org/doc/effective_go.html#constants) for these constant blocks, since it makes them much clearer: readers don't have to try to figure out if there are any gaps in the list.

tpm2/constants.go
```diff
@@ -0,0 +1,181 @@
...
+	CapabilityPPCommands    Capability = 0x00000003
+	CapabilityAuditCommands Capability = 0x00000004
+	CapabilityPCRs          Capability = 0x00000005
+	CapabilityTPMProperties Capability = 0x00000006
+	CapabilityPCRProperties Capability = 0x00000007
+	CapabilityECCCurves     Capability = 0x00000008
+	CapabilityAuthPolicies  Capability = 0x00000009
+)
+
+type structureTag uint16
```
tmroeder: Why is `structureTag` not exported and the others are? Even though it's not supported, I'd like to have a comment about the type so that readers can have some context. That comment could also explain why it's not exported.

tpm2/constants.go
```diff
@@ -0,0 +1,181 @@
...
+
+type command uint32
+
+// Supported TPM operations.
+const (
+	cmdEvictControl       command = 0x00000120
+	cmdClockSet           command = 0x00000128
+	cmdPCRAllocate        command = 0x0000012B
+	cmdCreatePrimary      command = 0x00000131
+	cmdCreate             command = 0x00000153
```
tmroeder: Please put these in numeric order.

tpm2/encoding.go
```diff
@@ -0,0 +1,246 @@
...
+	var size uint32
+	var status uint32
+
+	out := []interface{}{&tag, &size, &status}
+	err := unpack(in, out)
+	if err != nil {
+		return 0, 0, 0, fmt.Errorf("decoding response header: %v", err)
+	}
+
+	return tag, size, responseCode(status), nil
```
tmroeder: I don't have the details of packing and unpacking in my head right now, but is it possible to allow unpack to unpack into a `responseCode`? If it's a matter of the type not being exported, then this would be a good argument for exporting it.

tpm2/encoding.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// packedSize computes the size of a sequence of types that can be passed to
+// binary.Read or binary.Write.
+func packedSize(elts ...interface{}) (int, error) {
```
tmroeder: Is there a way to share this code with the tpm package? Maybe there could be a tpmutil package that had packing and unpacking so the code doesn't need to be duplicated.

tmroeder:
I started looking at this, then got to the point where I saw a lot of code that looked similar to code in the tpm package. Is it possible to refactor this first into a utility package so that there doesn't need to be code duplication?

I'll send out my initial comments, but I don't want to go to far with them if there's going to be significant refactoring.

awly:
PTAL

Extracted common encoding/decoding code into package tpmutil. Also added helper func to execute the command (write, read, parse response header).
Both libraries are updated to use it now.

Also addressed comments in constants.go. 

Regarding iota: many constant tables in https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf do have gaps, which is why i stayed away from it initially.
But added iota to some constant blocks that indeed are gap-less.

tmroeder:
I like how the refactored code looks, though I haven't finished going over it.

Would it be possible to separate this into two PRs for ease of review? I'm a little overloaded with the number of lines of changes here. I'd like to have a review that was just the refactoring of tpm into tpmutil (without any of the new tpm2 code). Then a review that is the tpm2 code, now using the tpmutil library.

Also, it looks like there's a problem with the CLA for one or more the commits. Maybe it's the jlmucb one?

awly:
Yeah, the CLA bot also marks my commits for some reason.
@jlmucb would you mind checking at https://cla.developers.google.com/clas if it requires you to sign one?

Will split out tpmutil parts into separate PR, good idea.

awly:
Rebased on master. PTAL

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+
+	err := tpmutil.Unpack(in, &publicBlob, &name, &qualifiedName)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	return publicBlob, name, qualifiedName, nil
+}
+
+// ReadPublic reads the public part of the object under handle.
+// Returns the public data, name and qualified name.
```
tmroeder: Here, too, it would be nice to have a reference for the formats of these byte slices.

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+func encodeCreateKey(owner Handle, pcrNums []int, parentPassword, ownerPassword string, parms RSAParams) ([]byte, error) {
+	b1, err := tpmutil.Pack(Handle(owner))
+	if err != nil {
+		return nil, err
+	}
+	b2, err := tpmutil.Pack([]byte(nil))
+	if err != nil {
+		return nil, err
+	}
+	b3, err := encodePasswordAuthArea(parentPassword, HandlePasswordSession)
```
tmroeder: It looks like this sequence (`encodePasswordAuthArea`, `encodePasswordData`, `encodeSensitiveArea`, `encodeRSAParams`, `tpmutil.Pack`) occurs at least twice. Consider refactoring to make this a well-known sequence of operations so the code doesn't have to be repeated each time. This will also make the semantics of the operation clearer.

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	}
+	return privateBlob, publicBlob, nil
+}
+
+func encodeLoad(parentHandle Handle, parentAuth, ownerAuth string, publicBlob, privateBlob []byte) ([]byte, error) {
+	b1, err := tpmutil.Pack(parentHandle)
+	if err != nil {
+		return nil, err
+	}
+	b3, err := encodePasswordData(parentAuth)
```
tmroeder: What happened to `b2`? Is this a copy-paste issue? These two calls (`encodePasswordData`, `encodePasswordAuthArea`) seem like good candidates for refactoring as well.

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	_, err := runCommand(rw, tagNoSessions, cmdPolicyPassword, handle)
+	return err
+}
+
+func encodePolicyPCR(handle Handle, expectedDigest []byte, pcrNums []int) ([]byte, error) {
+	b1, err := tpmutil.Pack(handle, expectedDigest)
+	if err != nil {
+		return nil, err
+	}
+	b2, err := encodeLongPCR(1, pcrNums)
```
tmroeder: Please comment on the meaning of `1` in the call to `encodeLongPCR`.

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	}
+	b2, err := tpmutil.Pack(bindKey)
+	if err != nil {
+		return nil, err
+	}
+	b3, err := tpmutil.Pack(nonceCaller, secret)
+	if err != nil {
+		return nil, err
+	}
+	b5, err := tpmutil.Pack(sym, hashAlg)
```
tmroeder: `b4` is missing

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	}
+
+	err = tpmutil.Unpack(buf, &attest, &s1, &s2, &signature)
+	if err != nil {
+		return nil, 0, 0, nil, err
+	}
+	return attest, s1, s2, signature, nil
+}
+
+// Quote returns a quote of PCR values.
```
tmroeder: Maybe mention here what a Quote is. Even a reference to the TPM docs would be useful for readers.

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	}
+	err = tpmutil.Unpack(buf, &certInfo)
+	if err != nil {
+		return nil, err
+	}
+	return certInfo, nil
+}
+
+// ActivateCredential associates an object with a credential.
+// Returns decrypted certificate information.
```
tmroeder: From just reading this comment, it's not clear what the certificate is that is getting decrypted. Please add a little more text here to make it clear to readers.

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	}
+	args := append(b1, b2...)
+	args = append(args, b3...)
+	args = append(args, b4...)
+	args = append(args, b5...)
+	return args, nil
+}
+
+// EvictControl toggles persistence of an object within the TPM.
+func EvictControl(rw io.ReadWriter, owner, tmpHandle, persistantHandle Handle) error {
```
tmroeder: spelling nit: `persistantHandle` -> `persistentHandle` (change `ant` to `ent`)

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	cmd, err := encodeEvictControl(owner, tmpHandle, persistantHandle)
+	if err != nil {
+		return err
+	}
+	_, err = runCommand(rw, tagSessions, cmdEvictControl, tpmutil.RawBytes(cmd))
+	return err
+}
+
+// ContextSave returns an encrypted version of the session, object or sequence
+// context for storage outside of the TPM.
```
tmroeder: Please add a little more context to the comment. I think it must be the case that the choice of session, object, or sequence depends on what the Handle refers to. Is that right?

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+		return 0, err
+	}
+	handle, err := decodeLoadContext(resp)
+	if err != nil {
+		return 0, fmt.Errorf("decoding LoadContext response: %v", err)
+	}
+	return handle, nil
+}
+
+func encodeIncrementNv(handle Handle, authString string) ([]byte, error) {
```
tmroeder: Go style would prefer "NV" (capital V) instead of "Nv".

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	}
+	out, err := tpmutil.Pack(handle, handle, []byte(nil))
+	if err != nil {
+		return nil, err
+	}
+	out = append(out, auth...)
+	return out, nil
+}
+
+// NVIncrement increments a counter NV index.
```
tmroeder: For clarity, maybe replace "a counter NV index" with "a counter in NVRAM".

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	auth, err := encodePasswordAuthArea("", HandlePasswordSession)
+	if err != nil {
+		return nil, err
+	}
+	out1, err := tpmutil.Pack(owner, []byte(nil))
+	if err != nil {
+		return nil, err
+	}
+	hashAlg := AlgSHA1
+	sizeNvArea := uint16(2*int(unsafe.Sizeof(owner)) + 3*int(unsafe.Sizeof(dataSize)) + len(policy))
```
tmroeder: `sizeNvArea` -> `sizeNVArea` (fully capitalize initialisms for Go style)

Also, please add a comment about why this computation is the way it is.

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	out = append(out, auth...)
+	params, err := tpmutil.Pack(dataSize, offset)
+	if err != nil {
+		return nil, err
+	}
+	out = append(out, params...)
+	return out, nil
+}
+
+// NVRead reads full data blob from NV index.
```
tmroeder: grammar nit: "reads full data blob from NV index" -> "reads a full data blob from an NV index" (add the indefinite article twice)

tpm2/tpm2_test.go
```diff
@@ -0,0 +1,318 @@
...
+func TestMain(m *testing.M) {
+	flag.Parse()
+	os.Exit(m.Run())
+}
+
+func TestGetRandom(t *testing.T) {
+	if !*runIntegration {
+		t.SkipNow()
+	}
+	rw, err := OpenTPM("/dev/tpm0")
```
tmroeder: [optional]: consider making the path a flag so it can be used to point to, e.g., a Unix domain socket backed by a TPM 2.0 software emulator.

tpm2/tpm2_test.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		t.Fatalf("OpenTPM failed: %s", err)
+	}
+	defer rw.Close()
+
+	pcr := []byte{0x03, 0x80, 0x00, 0x00}
+	_, _, _, _, err = ReadPCRs(rw, pcr)
```
tmroeder: For Go style: collapse this statement into the next if statement. Same comment for other tests in this file.

tpm2/tpm2_test.go
```diff
@@ -0,0 +1,318 @@
...
+		1024,
+		uint32(0x00010001),
+		[]byte(nil),
+	}
+	parentHandle, publicBlob, err := CreatePrimary(rw, HandleOwner, []int{0x7}, "", "01020304", primaryparms)
+	if err != nil {
+		t.Fatalf("CreatePrimary failed: %s", err)
+	}
+
+	keyparms := RSAParams{
```
tmroeder: This looks like it's identical to primaryparms. Can that object be reused?

tpm2/tpm2_test.go
```diff
@@ -0,0 +1,318 @@
...
+		AlgAES,
+		128,
+		AlgCFB,
+		AlgNull,
+		0,
+		1024,
+		uint32(0x00010001),
+		[]byte(nil),
+	}
+	parentHandle, publicBlob, err := CreatePrimary(rw, HandleOwner, []int{0x7}, "", "01020304", primaryparms)
```
tmroeder: Consider putting this into a variable/const, given that it's reused in test later.

tpm2/tpm2_test.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		t.Fatalf("OpenTPM failed: %v", err)
+	}
+	defer rw.Close()
+
+	pcrs := []int{7}
+	keySize := 2048
+	quotePassword := ""
+
+	primaryparms := RSAParams{
```
tmroeder: If most of the parameters are the same across all uses of RSAParams in this test file, consider making a template version at the top of this file and filling in only the parts that need to change for each test.

tpm2/tpm2_test.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	FlushContext(rw, quoteHandle)
+
+	quoteHandle, err = ContextLoad(rw, saveArea)
+	if err != nil {
+		t.Fatalf("Load failed: %v", err)
+	}
+
+	FlushContext(rw, quoteHandle)
+}
```
tmroeder: Not in this PR, but in future ones, please add some negative tests that show that the TPM fails appropriately, e.g., when the password auth is wrong.

tpm2/encoding_test.go
```diff
@@ -0,0 +1,436 @@
...
+package tpm2
+
+import (
+	"bytes"
+	"encoding/hex"
+	"testing"
+)
+
+func TestDecodeGetRandom(t *testing.T) {
+	testRespBytes, err := hex.DecodeString("80010000001c00000000001024357dadbf82ec9f245d1fcdcda33ed7")
```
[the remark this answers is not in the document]
awly: These values are from https://github.com/jlmucb/cloudproxy/blob/master/go/tpm2/tpm2_command_encoding_test.go and I'm not entirely sure how they got created.

tpm2/encoding_test.go
```diff
@@ -0,0 +1,436 @@
...
+		"2244fedc13fe0abb526e64b10b2de030b6f02be278e23365ef663febe7e" +
+		"b4ddae935ca627ce4c40af9f5244dafbc7f47ceb84de87e72a75c7f1032" +
+		"d3e7faddde0036000800040000001200140debb4cc9d2158cf7051a19ca" +
+		"24b31e35d53b64d001000140b0758c7e4ce32c9d249151e91b72e35a6372fed")
+	if err != nil {
+		t.Fatal(err)
+	}
+	privateBlob := testCmdBytes[33:123]
+	publicBlob := testCmdBytes[125:]
+	cmdBytes, err := encodeLoad(Handle(0x80000000), "", "01020304", publicBlob, privateBlob)
```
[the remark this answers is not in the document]
awly: Done

tpm2/tpm2_test.go
```diff
@@ -0,0 +1,318 @@
...
+		FlushContext(rw, parentHandle)
+		t.Fatalf("ActivateCredential failed: %s", err)
+	}
+	if bytes.Compare(credential, recoveredCredential1) != 0 {
+		FlushContext(rw, keyHandle)
+		FlushContext(rw, parentHandle)
+		t.Fatalf("Credential and recovered credential differ: got %v, want %v", recoveredCredential1, credential)
+	}
+
+	FlushContext(rw, keyHandle)
```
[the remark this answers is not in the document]
awly: Done

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	"os"
+	"unsafe"
+
+	"github.com/google/go-tpm/tpmutil"
+)
+
+// OpenTPM opens a channel to the TPM at the given path. If the file is a
+// device, then it treats it like a normal TPM device, and if the file is a
+// Unix domain socket, then it opens a connection to the socket.
+func OpenTPM(path string) (io.ReadWriteCloser, error) {
```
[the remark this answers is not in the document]
awly: Moved and added aliases in each tpm/tpm2 for backward-compatibility

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	var rsaBuf []byte
+
+	err := tpmutil.Unpack(in, &rsaBuf)
+	if err != nil {
+		return nil, err
+	}
+	return decodeRSABuf(rsaBuf)
+}
+
+func encodeKeyedHashParams(parms KeyedHashParams) ([]byte, error) {
```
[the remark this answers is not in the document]

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return rand, nil
+}
+
+// FlushContext removes an object or session under handle to be removed from
+// the TPM. This must be called for any loaded handle to avoid out-of-memory
+// errors in TPM.
+func FlushContext(rw io.ReadWriter, handle Handle) error {
```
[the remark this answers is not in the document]

tpm2/tpm2.go
```diff
@@ -0,0 +1,1247 @@
...
+
+func decodeGetCapability(in []byte) (Capability, []Handle, error) {
+	var numHandles uint32
+	var capReported Capability
+
+	err := tpmutil.Unpack(in[1:9], &capReported, &numHandles)
+	if err != nil {
+		return 0, nil, err
+	}
+	// only TPM_CAP_HANDLES handled
```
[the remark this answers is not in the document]

tpm2/tpm2.go
awly: Done. Looks like TPM2_CreatePrimary and TPM2_Create take the exact same set of parameters.

tpm2/tpm2.go
awly: Added a small clarification, but this package is full of TPM lingo (PCR, Handle, Session etc).
I'm not sure how much detail should the package provide, it should be assumed that reader has some understanding of TPMs.

That being said, there is no good concise learning material for TPM 2.0, so perhaps more detail is good.

tpm2/tpm2.go
awly: Yes, handle points to _something_ that is being saved.

tpm2/tpm2.go

tpm2/tpm2.go

tpm2/tpm2.go
awly: Done for naming

tpm2/tpm2.go

tpm2/tpm2_test.go
awly: Replaced bool flag with TPM path flag.
Empty flag for disabling tests. Also moved OpenTPM into a helper func.

tpm2/tpm2_test.go
awly: Done

tpm2/tpm2_test.go
awly: Done

tpm2/tpm2_test.go
awly: Done

tpm2/tpm2_test.go
awly: Done.

tpm2/tpm2_test.go
awly: Acknowledged.

awly:
Fixed most straightforward comments.

Regarding math and naming in encode.*/decode.* funcs: those were copied over from cloudproxy repo and it will be tricky for me to explain most of them.
Also, I feel like most of them can be simplified by defining structs for parameters, instead of manually decoding each field.

I need to do several things things:
- actually run integration tests in a VM and make sure they work (not confident enough in them to run locally)
(ideally in followup CL, but up to you):
- add integration tests for any commands that miss them
- replace most/all encoding/decoding funcs with proper structs

Let me know if you'd be ok with last 2 items being done in separate PRs.

tmroeder:
Yes, I'm fine with last two items (more integration tests and encoding/decoding structs) waiting for a later PR

jlmucb:
Just signed it again

awly:
Thanks John.

Tom, fixed ReadPCRs test, everything else now passes against a real device.
PTAL.

tmroeder:
Thanks. BTW, this will need to be rebased against master again, because I ran across a bug caused by the tpmutil refactoring. It is fixed now on master in commit c5c955955f4946a192b77e31952338b3ba36bc54

awly:
Rebased to master

awly:
Ping @tmroeder

tpm2/tpm2.go
```diff
@@ -0,0 +1,1202 @@
...
+	// the selected PCR index.
+	// Size of the bitmask in bytes is pre-pended. It should be at least
+	// sizeOfPCRSelect.
+	//
+	// For example, selecting PCRs 3 and 9 looks like:
+	// size(3)  mask     mask     mask
+	// 00000011 00000000 00000001 00000100
+	pcr := make([]byte, sizeOfPCRSelect+1)
+	pcr[0] = sizeOfPCRSelect
+	// pcrNums parameter is indexes of PCRs, convert hat to set bits.
```
tmroeder: nit: "convert hat" -> "convert that" ?

tpm/tpm.go
```diff
@@ -24,44 +24,15 @@ import (
...
-			return nil, err
-		}
-		rwc = io.ReadWriteCloser(uc)
-	} else {
-		return nil, fmt.Errorf("unsupported TPM file mode %s", fi.Mode().String())
-	}
-
-	return rwc, nil
-}
+var OpenTPM = tpmutil.OpenTPM
```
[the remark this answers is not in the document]
awly: This would break any upstream users of this package.
I'm happy to remove it, but just making sure you're OK with that.

tpm2/structures.go
```diff
@@ -0,0 +1,77 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tpm2
+
+import "github.com/google/go-tpm/tpmutil"
+
+// RSAParams us a template for an RSA key.
```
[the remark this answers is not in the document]

tpm2/tpm2.go
```diff
@@ -0,0 +1,1202 @@
...
+
+func decodeRSABuf(rsaBuf []byte) (*RSAParams, error) {
+	params := new(RSAParams)
+	current := 0
+	err := tpmutil.Unpack(rsaBuf[current:], &params.EncAlg, &params.HashAlg, &params.Attributes, &params.AuthPolicy)
+	if err != nil {
+		return nil, err
+	}
+	current += 10 + len(params.AuthPolicy)
+	err = tpmutil.Unpack(rsaBuf[current:], &params.SymAlg)
```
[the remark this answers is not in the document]
awly: Good idea, added a TODO in Unpack

tpm2/tpm2.go

tpm/tpm.go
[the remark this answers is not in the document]
tmroeder: Good point. I guess it's OK to leave it like this for now.

jlmucb:
This acct

tmroeder:
jlmucb@: sorry to ask for one more thing, but the googlebot says to get you to respond in this pull request to say that you are OK with having your code contributed to this project (go-tpm). I think it's obviously the case that you are OK with contributing your code, given the work you've done to get this code to us, but I haven't had this situation before, and I want to make sure that we are not only correct but are obviously correct.

tmroeder:
@jlmucb: never mind, actually. On reflection, I think it's extremely clear from this PR that you are both aware that your code is being contributed to this repository and that you are OK with that, since you've jumped through multiple hoops for the CLA, and you originally sent a PR with that original code to this repo. So, I think there's no need for you to make it any clearer.

PR #6, opened 2018-02-08
Extract common encoding code into separate package

awly:
Package tpmutil can be re-used for TPM 2.0 implementation.

Note that tpmutil has some mentions of TPM 2.0 already. Specifically LengthPrefixSize.


tpm/commands.go
```diff
@@ -18,61 +18,23 @@ import (
 	"encoding/binary"
 	"errors"
 	"io"
-	"strconv"
+
+	"github.com/awly/go-tpm/tpmutil"
```
tmroeder: Is it possible to make this github.com/google/go-tpm/tpmutil, since this PR will cause that directory to exist? The way I usually do this for testing is to check out my repo (in this case, it would be github.com/awly/go-tpm) into "${GOPATH}/src/github.com/google/go-tpm", even though it's really under awly. Then the go compiler is happy with the paths, and the checked-in version will work. @

tpm/constants.go
```diff
@@ -14,6 +14,12 @@
 
 package tpm
 
+import "github.com/awly/go-tpm/tpmutil"
+
+func init() {
+	tpmutil.LengthPrefixSize = 4
```
tmroeder: Please add a comment about why this is necessary.

tpm/tpm.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package tpm supports direct communication with a tpm device under Linux.
+// tpmutil.Package tpm supports direct communication with a tpm device under Linux.
```
tmroeder: I think this is a copy-paste error (s/pack/tpmutil.Pack/gi ?). This comment should start with "Package tpm"

tpm/tpm.go
```diff
@@ -561,9 +567,9 @@ func MakeIdentity(rw io.ReadWriter, srkAuth []byte, ownerAuth []byte, aikAuth []
 			return nil, err
 		}
 
-		// We can't pack the pair of values directly, since the label is
+		// We can't tpmutil.Pack the pair of values directly, since the label is
```
tmroeder: This looks like another copy-paste error. I think the comment should probably just say "pack", not "tpmutil.Pack"

tpmutil/encoding.go
```diff
@@ -0,0 +1,266 @@
...
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"reflect"
+)
+
+// LengthPrefixSize is the size in bytes of length prefix for byte slices.
+//
+// In TPM 1.2 this is 4 bytes.
```
tmroeder: In cases like this, I find it useful to simplify the job of the user of this package. Since this code knows that this value should either be 4 bytes or 2 bytes, consider adding a function to set it and making this variable internal to the package. This makes client code clearer and reduces the chances of an error. E.g.,

```go
func SetTPM12LengthPrefixSize() {
  LengthPrefixSize = 4
}

func SetTPM20LengthPrefixSize() {
  LengthPrefixSize = 2
}
```

tpmutil/encoding.go
```diff
@@ -0,0 +1,266 @@
...
+// In TPM 2.0 this is 2 bytes.
+var LengthPrefixSize int
+
+// packedSize computes the size of a sequence of types that can be passed to
+// binary.Read or binary.Write.
+func packedSize(elts ...interface{}) (int, error) {
```
tmroeder: Is it possible to make this a git mv so that this code review reflects differences between the tpmutil version of the functions and the original TPM 1.2 version?

tpmutil/encoding.go
```diff
@@ -0,0 +1,266 @@
...
+	in = append(in, cmd...)
+	return Pack(in...)
+}
+
+// Pack encodes a set of elements into a single byte array, using
+// encoding/binary. This means that all the elements must be encodeable
+// according to the rules of encoding/binary.
+//
+// It has one difference from encoding/binary: it encodes byte slices with a
+// prepended uint16 length, to match how the TPM encodes variable-length
```
tmroeder: I think this comment has gone stale (and probably other comments). Instead of a `uint16` for the size, this now depends on `LengthPrefixSize`, and this comment should say that.

tpmutil/encoding_test.go
```diff
@@ -90,85 +92,71 @@ type nestedSlice struct {
...
+		{[]byte(nil), 4},
+	}
+	for _, tt := range tests {
+		if s, err := packedSize(tt.in); err != nil || s != tt.want {
+			t.Errorf("packedSize(%#v): %d, want %d", tt.in, s, tt.want)
+		}
 	}
 }
 
 func TestEncodingPackType(t *testing.T) {
```
tmroeder: Can this test also be turned into a table-based test?

tpmutil/run.go
```diff
@@ -0,0 +1,64 @@
...
+	"encoding/binary"
+	"errors"
+	"io"
+)
+
+const maxTPMResponse = 4096
+
+// RunCommand executes cmd with given tag and arguments. Returns TPM response
+// body (without response header) and response code from the header. Returned
+// error may be nil if response code is not RCSuccess, caller should check
```
tmroeder: tiny language nit: ", caller should check both" is a comma splice. I'd prefer a semi-colon instead of a comma. But I recognize that this is already a comma splice of sentence fragments and is in a highly stylized form, so I'm OK if you won't want to change this.

tpmutil/structures.go
```diff
@@ -0,0 +1,50 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tpmutil
+
+// RawBytes is for Pack and RunCommand arguments that are already encoded.
+// Compared to []byte, RawBytes will not be pre-pended with slice length during
```
tmroeder: tiny nit: "pre-prended" is a single word: "prepended"

tpm/tpm.go
awly: Done

tpmutil/encoding.go
awly: Good idea, done

tpmutil/encoding.go
awly: I'm not sure how to do this retroactively. Is there some history shuffle magic you would suggest?

tpmutil/encoding.go
awly: Removed length type from comment. LengthPrefixSize is not exported now.
Added package comment about UseTPMXXLengthPrefixSize functions instead.

tpmutil/structures.go
awly: Done

tpmutil/run.go
awly: Done

tpmutil/encoding_test.go
awly: Done

awly:
PTAL

tpmutil/encoding.go
tmroeder: Actually, I'm not sure why git doesn't detect it in this commit, since encoding.go is removed from tpm/ and is created in tpmutil/, and they're very similar. I was going to say that it was because the changes were in different commits, but that doesn't seem to be the case. I'll just deal with it...

tpmutil/encoding.go
```diff
@@ -0,0 +1,283 @@
...
+
+// Unpack performs the inverse operation from Pack.
+func Unpack(b []byte, elts ...interface{}) error {
+	buf := bytes.NewBuffer(b)
+	return unpackType(buf, elts...)
+}
+
+// unpackType recursively unpacks types from a reader just as encoding/binary
+// does under binary.BigEndian, but with one difference: it unpacks a byte slice
+// by first reading a uint16, then reading that many bytes. It assumes that
```
tmroeder: "a uint16" -> "an integer with lengthPrefixSize bytes"

awly: Done

PR #10, opened 2018-03-05
Support for ECC keys in CreateKey and LoadExternal

awly:
Also move some encoding/decoding code to use tpmutil.UnpackBuf (bytes.Buffer alternative to tpmutil.Unpack).

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -31,11 +31,12 @@ type Algorithm uint16
 
 // IsNull returns true if a is AlgNull or zero (unset).
 func (a Algorithm) IsNull() bool {
-	return a == AlgNull || a == 0
+	return a == AlgNull || a == AlgUnknown
 }
 
 // Supported Algorithms.
 const (
+	AlgUnknown   Algorithm = 0x0000
 	AlgRSA       Algorithm = 0x0001
 	AlgSHA1      Algorithm = 0x0004
 	AlgAES       Algorithm = 0x0006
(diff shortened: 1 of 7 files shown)


tpm2/constants.go
```diff
@@ -29,6 +29,11 @@ func init() {
 // Algorithm represents a TPM_ALG_ID value.
 type Algorithm uint16
 
+// IsNull returns true if a is AlgNull or zero (unset).
+func (a Algorithm) IsNull() bool {
+	return a == AlgNull || a == 0
```
tmroeder: For clarity in the code, how about naming the 0 case AlgUnset or AlgUnknown?

tpm2/constants.go
```diff
@@ -136,6 +142,23 @@ const (
 	StartupState
 )
 
+// ECCCurve identifies specific ECC curves.
```
tmroeder: ECCCurve == "Elliptic Curve Cryptography Curve", which seems redundant to me. How about just EllipticCurve?

tpm2/structures.go
```diff
@@ -74,15 +51,264 @@ type tpmsPCRSelection struct {
...
+		return nil, err
+	}
+	return concat(sym, sig, rest)
+}
+
+func decodeRSAParams(in *bytes.Buffer) (*RSAParams, error) {
+	var params RSAParams
+	var err error
+
+	params.Symmetric, err = decodeSymScheme(in)
```
tmroeder: style nit: This line can merge with the next if statement, and the same is true for line 146

tpm2/structures.go
```diff
@@ -74,15 +51,264 @@ type tpmsPCRSelection struct {
...
+type ECCParams struct {
+	Symmetric *SymScheme
+	Sign      *SigScheme
+	CurveID   ECCCurve
+	KDF       *KDFScheme
+	Point     ECCPoint
+}
+
+// ECCPoint represents a ECC coordinates for a point.
+type ECCPoint struct {
```
tmroeder: How about ECPoint instead, since this is just a point on a curve. It happens to be used for cryptography in this case, but I don't think the extra "C" is meaningful in the name.

tpm2/tpm2.go
```diff
@@ -80,28 +79,23 @@ func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
...
-	read, err = tpmutil.Unpack(buf, &ts.Hash, &ts.Size)
-	if err != nil {
-		return 0, sel, err
+	if err := tpmutil.UnpackBuf(buf, &ts.Hash, &ts.Size); err != nil {
+		return sel, err
 	}
-	buf = buf[read:]
-	ts.PCRs, buf = buf[:ts.Size], buf[ts.Size:]
+	ts.PCRs = make([]byte, ts.Size)
+	buf.Read(ts.PCRs)
```
tmroeder: Please check the return value for errors here.

tpm2/constants.go

tpm2/constants.go
awly: Done.
I was trying to keep this close to name in the spec, which is TPM_ECC_CURVE.
But agree, it's more awkward and not necessary

tpm2/structures.go

tpm2/structures.go

tpm2/tpm2.go

awly:
PTAL

awly:
Ah, forgot to run all tests.
Fixed.

PR #11, opened 2018-03-22
Add tests and ECC support to Sign

awly:
Also fix response parsing.

### DIFF
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -306,10 +306,49 @@ func decodeKDFScheme(in *bytes.Buffer) (*KDFScheme, error) {
 	return &scheme, nil
 }
 
-type tpmtSignatureRSA struct {
-	SigAlg    Algorithm
+// Signature combines all possible signatures from RSA and RCC keys. Only one
+// of RSA or ECC will be populated.
+type Signature struct {
+	Alg Algorithm
+	RSA *SignatureRSA
+	ECC *SignatureECC
+}
+
+func decodeSignature(in *bytes.Buffer) (Signature, error) {
+	var sig Signature
+	if err := tpmutil.UnpackBuf(in, &sig.Alg); err != nil {
+		return sig, err
+	}
+	switch sig.Alg {
+	case AlgRSASSA:
+		sig.RSA = new(SignatureRSA)
+		if err := tpmutil.UnpackBuf(in, sig.RSA); err != nil {
+			return sig, err
+		}
+	case AlgECDSA:
+		sig.ECC = new(SignatureECC)
+		var r, s []byte
+		if err := tpmutil.UnpackBuf(in, &sig.ECC.HashAlg, &r, &s); err != nil {
+			return sig, err
+		}
+		sig.ECC.R = big.NewInt(0).SetBytes(r)
+		sig.ECC.S = big.NewInt(0).SetBytes(s)
+	default:
+		return sig, fmt.Errorf("unsupported signature algorithm 0x%x", sig.Alg)
+	}
+	return sig, nil
+}
+
+// SignatureRSA is an RSA-specific signature value.
+type SignatureRSA struct {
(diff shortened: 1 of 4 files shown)


tpm2/tpm2.go
```diff
@@ -347,15 +349,28 @@ func decodeCreatePrimary(in []byte) (tpmutil.Handle, *rsa.PublicKey, error) {
...
+		// Endianness of big.Int.Bytes/SetBytes and modulus in the TPM is the same
+		// (big-endian).
+		pubKey = &rsa.PublicKey{N: pub.RSAParameters.Modulus, E: int(pub.RSAParameters.Exponent)}
+	case AlgECC:
+		pubKey = &ecdsa.PublicKey{X: pub.ECCParameters.Point.X, Y: pub.ECCParameters.Point.Y}
+		curve, ok := toGoCurve[pub.ECCParameters.CurveID]
+		if !ok {
+			return 0, nil, fmt.Errorf("can't map TPM EC curve ID 0x%x to Go elliptic.Curve value", pub.ECCParameters.CurveID)
+		}
+		pubKey.(*ecdsa.PublicKey).Curve = curve
```
tmroeder: [optional]: Instead of having to cast, why not try to convert the curve value first? Then the Curve field can be filled in directly in the assignment to pubKey:

```go
		curve, ok := toGoCurve[pub.ECCParameters.CurveID]
		if !ok {
			return 0, nil, fmt.Errorf("can't map TPM EC curve ID 0x%x to Go elliptic.Curve value", pub.ECCParameters.CurveID)
		}
		pubKey = &ecdsa.PublicKey{
                      X: pub.ECCParameters.Point.X,
                      Y: pub.ECCParameters.Point.Y,
                      Curve: curve,
                }
```

tpm2/tpm2.go
```diff
@@ -667,14 +682,14 @@ func decodeQuote(in []byte) ([]byte, []byte, error) {
 // Returns attestation data and the signature.
 //
 // Note: currently only RSA signatures are supported.
-func Quote(rw io.ReadWriter, signingHandle tpmutil.Handle, parentPassword, ownerPassword string, toQuote []byte, sel PCRSelection, sigAlg Algorithm) ([]byte, []byte, error) {
+func Quote(rw io.ReadWriter, signingHandle tpmutil.Handle, parentPassword, ownerPassword string, toQuote []byte, sel PCRSelection, sigAlg Algorithm) ([]byte, Signature, error) {
```
tmroeder: Why return Signature instead of *Signature? If the latter, then this code (and similar calls) could return nil in the error case instead of a zero Signature struct.

tpm2/structures.go
```diff
@@ -306,12 +306,51 @@ func decodeKDFScheme(in *bytes.Buffer) (*KDFScheme, error) {
 	return &scheme, nil
 }
 
-type tpmtSignatureRSA struct {
-	SigAlg    Algorithm
+// Signature combines all possible signatures from RSA and RCC keys. Only one
```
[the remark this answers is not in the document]

tpm2/structures.go
```diff
@@ -306,12 +306,51 @@ func decodeKDFScheme(in *bytes.Buffer) (*KDFScheme, error) {
...
+// SignatureRSA is an RSA-specific signature value.
+type SignatureRSA struct {
 	HashAlg   Algorithm
 	Signature []byte
 }
 
+// SignatureECC is an ECC-specific signature value.
+type SignatureECC struct {
+	HashAlg Algorithm
+	R, S    *big.Int
```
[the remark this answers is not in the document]

tpm2/tpm2.go
awly: Good call, done

tpm2/tpm2.go
awly: Was following pattern with Public, but agree - this make it cleaner. Done

awly:
PTAL

PR #12, opened 2018-04-10
Decode attest data

awly:
During remote attestation, remote process needs to check that the
attested key pair matches attestation data. Attestation data is what's
actually signed during TPM2_Certify.

Also added MatchesPublic helper to Name.

Remote process will have to:
- verify signature of attestDataRaw using attestation certificate
- attestData = DecodeAttestationData(attestDataRaw)
- attestData.AttestedCertifyInfo.Name.MatchesPublic(Public{publicKey})
  where publicKey is the attested key

Remote process doesn't need a TPM device to do the above.

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -133,7 +133,7 @@ const (
 	tagNoSessions    tpmutil.Tag = 0x8001
 	tagSessions      tpmutil.Tag = 0x8002
 	tagAttestCertify tpmutil.Tag = 0x8017
-	tagHashcheck     tpmutil.Tag = 0x8024
+	tagHashCheck     tpmutil.Tag = 0x8024
 )
 
 // StartupType instructs the TPM on how to handle its state during Shutdown or
@@ -231,7 +231,7 @@ func digestSize(alg Algorithm) int {
 
 const defaultRSAExponent = 1<<16 + 1
 
-var toGoHashes = map[Algorithm]func() hash.Hash{
+var hashConstructors = map[Algorithm]func() hash.Hash{
 	AlgSHA1:   sha1.New,
 	AlgSHA256: sha256.New,
 	AlgSHA384: sha512.New384,
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -128,10 +128,11 @@ const (
 )
 
 const (
-	tagNull       tpmutil.Tag = 0x8000
-	tagNoSessions tpmutil.Tag = 0x8001
-	tagSessions   tpmutil.Tag = 0x8002
-	tagHashcheck  tpmutil.Tag = 0x8024
+	tagNull          tpmutil.Tag = 0x8000
+	tagNoSessions    tpmutil.Tag = 0x8001
+	tagSessions      tpmutil.Tag = 0x8002
+	tagAttestCertify tpmutil.Tag = 0x8017
+	tagHashcheck     tpmutil.Tag = 0x8024
 )
 
 // StartupType instructs the TPM on how to handle its state during Shutdown or
@@ -228,3 +229,11 @@ func digestSize(alg Algorithm) int {
 }
 
 const defaultRSAExponent = 1<<16 + 1
+
+var hashSizes = map[Algorithm]int{
+	AlgNull:   0,
+	AlgSHA1:   20,
+	AlgSHA256: 32,
+	AlgSHA384: 48,
+	AlgSHA512: 64,
+}
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -19,6 +19,7 @@ import (
 	"crypto/sha1"
 	"crypto/sha256"
 	"crypto/sha512"
+	"hash"
 
 	"github.com/google/go-tpm/tpmutil"
 )
@@ -230,10 +231,9 @@ func digestSize(alg Algorithm) int {
 
 const defaultRSAExponent = 1<<16 + 1
 
-var hashSizes = map[Algorithm]int{
-	AlgNull:   0,
-	AlgSHA1:   20,
-	AlgSHA256: 32,
-	AlgSHA384: 48,
-	AlgSHA512: 64,
+var toGoHashes = map[Algorithm]func() hash.Hash{
+	AlgSHA1:   sha1.New,
+	AlgSHA256: sha256.New,
+	AlgSHA384: sha512.New384,
+	AlgSHA512: sha512.New,
 }
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -384,15 +384,12 @@ type AttestationData struct {
 }
 
 // DecodeAttestationData decode a TPMS_ATTEST message. No error is returned if
-// in has extra trailing data.
+// the input has extra trailing data.
 func DecodeAttestationData(in []byte) (*AttestationData, error) {
 	buf := bytes.NewBuffer(in)
 
 	var ad AttestationData
-	if err := tpmutil.UnpackBuf(buf,
-		&ad.Magic,
-		&ad.Type,
-	); err != nil {
+	if err := tpmutil.UnpackBuf(buf, &ad.Magic, &ad.Type); err != nil {
 		return nil, fmt.Errorf("decoding Magic/Type: %v", err)
 	}
 	n, err := decodeName(buf)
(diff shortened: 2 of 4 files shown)


awly:
@tmroeder could you take a look at this when you have some time?

tpm2/constants.go
```diff
@@ -128,10 +129,11 @@ const (
...
 const (
-	tagNull       tpmutil.Tag = 0x8000
-	tagNoSessions tpmutil.Tag = 0x8001
-	tagSessions   tpmutil.Tag = 0x8002
-	tagHashcheck  tpmutil.Tag = 0x8024
+	tagNull          tpmutil.Tag = 0x8000
+	tagNoSessions    tpmutil.Tag = 0x8001
+	tagSessions      tpmutil.Tag = 0x8002
+	tagAttestCertify tpmutil.Tag = 0x8017
+	tagHashcheck     tpmutil.Tag = 0x8024
```
tmroeder: nit: I see that I missed this in previous reviews: shouldn't this be `tagHashCheck`? (capitalized `C`)

tpm2/constants.go
```diff
@@ -228,3 +230,10 @@ func digestSize(alg Algorithm) int {
 }
 
 const defaultRSAExponent = 1<<16 + 1
+
+var toGoHashes = map[Algorithm]func() hash.Hash{
```
tmroeder: I find the name a little ambiguous, since it says what the hash maps to, but not what it maps from. I'm not sure that it needs to say that it's converting to the Go programming language, given that it's written in Go.  How about `hashConstructors`?

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+	Type                tpmutil.Tag
+	QualifiedSigner     Name
+	ExtraData           []byte
+	ClockInfo           ClockInfo
+	FirmwareVersion     uint64
+	AttestedCertifyInfo *CertifyInfo
+}
+
+// DecodeAttestationData decode a TPMS_ATTEST message. No error is returned if
+// in has extra trailing data.
```
tmroeder: I originally thought that "in has trailing data" was a mistake, but then I saw that the name of the input argument was `in`. That suggests that while the comment is technically correct, it could use revision for clarity. How about "No error is return if the input has extra trailing data"?

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+func DecodeAttestationData(in []byte) (*AttestationData, error) {
+	buf := bytes.NewBuffer(in)
+
+	var ad AttestationData
+	if err := tpmutil.UnpackBuf(buf,
```
tmroeder: style: this doesn't seem consistent to me with other code so far: normally, UnpackBuf has been on a single line. I find this compound if statement hard to read like this. Can this (and similar lines) be pulled back up into a single line? If the line gets really long, then I don't mind breaking it up, but in this case, it's

```go
if err := tpmutil.UnpackBuf(buf, &ad.Magic, &ad.Type); err != nil {
```
and that seems reasonable; it even fits on an 80-char line.

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+	ad.QualifiedSigner = *n
+	if err := tpmutil.UnpackBuf(buf,
+		&ad.ExtraData,
+		&ad.ClockInfo,
+		&ad.FirmwareVersion,
+	); err != nil {
+		return nil, fmt.Errorf("decoding ExtraData/ClockInfo/FirmwareVersion: %v", err)
+	}
+
+	// Add support for more types as needed.
```
tmroeder: Please clarify this comment. Is this meant to be a TODO? Or is this meant to point out to the reader that this is one of the places that would need to be changed if more types were needed?

The reason I have trouble with this comment is that it's written in the imperative, which is not what I normally expect from a non-TODO comment.

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+	if ad.Type != tagAttestCertify {
+		return nil, fmt.Errorf("only Certify attestation structure is supported, got type 0x%x", ad.Type)
+	}
+	ad.AttestedCertifyInfo, err = decodeCertifyInfo(buf)
```
tmroeder: Given that this is `=` and not `:=`, this line can be folded into the following if statement

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("decoding AttestedCertifyInfo: %v", err)
+	}
+	return &ad, nil
+}
+
+// CertifyInfo contains Certify-specific data for TPMS_ATTEST.
```
tmroeder: I'd prefer to have this defined before it is used in the definition of the AttestationData structure, since that makes the code easier for readers to follow when going through the file linearly.

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+		return nil, fmt.Errorf("decoding QualifiedName: %v", err)
+	}
+	ci.QualifiedName = *n
+
+	return &ci, nil
+}
+
+// Name contains a Name for TPM entities. This is either a handle or a digest.
+// Only one of Handle/Digest will be set.
+type Name struct {
```
tmroeder: Same comment as for CertifyInfo; I'd like this to be defined before it is used.

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+
+// AttestationData contains data attested by TPM commands (like Certify).
+type AttestationData struct {
+	Magic               uint32
+	Type                tpmutil.Tag
+	QualifiedSigner     Name
+	ExtraData           []byte
+	ClockInfo           ClockInfo
+	FirmwareVersion     uint64
+	AttestedCertifyInfo *CertifyInfo
```
tmroeder: Why is QualifiedSigner (a Name struct) inline while AttestedCertifyInfo (a CertifyInfo) is a pointer?

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+	n, err = decodeName(in)
+	if err != nil {
+		return nil, fmt.Errorf("decoding QualifiedName: %v", err)
+	}
+	ci.QualifiedName = *n
+
+	return &ci, nil
+}
+
+// Name contains a Name for TPM entities. This is either a handle or a digest.
```
tmroeder: I don't think the second "Name" should be capitalized here, since it's not referring to the Name struct, but just the common noun "name" with its standard meaning. Also, please clarify for the reader: given the "either", does that mean that this is exclusive: only one of Handle and Digest should be set?

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+func decodeName(in *bytes.Buffer) (*Name, error) {
+	var nameBuf []byte
+	if err := tpmutil.UnpackBuf(in, &nameBuf); err != nil {
+		return nil, err
+	}
+
+	name := new(Name)
+	switch len(nameBuf) {
+	case 0:
+		// no name is present
```
tmroeder: style nit: Please capitalize the comment and add punctuation at the end.

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+}
+
+func decodeName(in *bytes.Buffer) (*Name, error) {
+	var nameBuf []byte
+	if err := tpmutil.UnpackBuf(in, &nameBuf); err != nil {
+		return nil, err
+	}
+
+	name := new(Name)
+	switch len(nameBuf) {
```
tmroeder: I don't know the spec well, but I'm surprised at this: is the only way to distinguish the two cases based on the size?

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+	Alg   Algorithm
+	Value []byte
+}
+
+func decodeHashValue(in *bytes.Buffer) (*HashValue, error) {
+	var hv HashValue
+	if err := tpmutil.UnpackBuf(in, &hv.Alg); err != nil {
+		return nil, fmt.Errorf("decoding Alg: %v", err)
+	}
+	h, ok := toGoHashes[hv.Alg]
```
tmroeder: For consistency with the previous code, this should be hfn, since h was used there as the result of calling the function returned from toGoHashes

tpm2/structures.go
```diff
@@ -371,3 +371,158 @@ type tpmtSigScheme struct {
...
+		return nil, fmt.Errorf("unsupported hash algorithm type 0x%x", hv.Alg)
+	}
+	hv.Value = make([]byte, h().Size())
+	if _, err := in.Read(hv.Value); err != nil {
+		return nil, fmt.Errorf("decoding Value: %v", err)
+	}
+	return &hv, nil
+}
+
+// ClockInfo contains TPM state info included in AttestationData.
```
tmroeder: Please also move this up to before it is used.

tpm2/constants.go

tpm2/constants.go

tpm2/structures.go
awly: Yep, that's nicer. Done

tpm2/structures.go

tpm2/structures.go
awly: Expanded the comment a bit.
It's not a TODO because I don't actually plan to implement it. But it's a suggestion for future reader, in case the need arises.

tpm2/structures.go

tpm2/structures.go
awly: This is personal preference, one where our views diverge :)
I prefer to see the higher-level code first and dive into details as needed.

The way it is now is consistent with the rest of this file (see Public and its fields). I'd like to keep it as is. If you feel strongly about the order, I can make a follow up PR moving things around.

tpm2/structures.go
awly: QualifiedSigner is always present in TPM response.
AttestedCertifyInfo is only populated if Type == tagAttestCertify. If we added support for other Type values, we'd add corresponding pointer Attested* fields. Only one of those fields would be set.

tpm2/structures.go

tpm2/structures.go

tpm2/structures.go
awly: Yes, unfortunately that's how it's described in the spec.

From https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf
"The type of Name in the structure is determined by context and the size parameter. If size is four, then
the Name is a handle. If size is zero, then no Name is present. Otherwise, the size shall be the size of a
TPM_ALG_ID plus the size of the digest produced by the indicated hash algorithm."

tpm2/structures.go

awly:
PTAL

tpm2/structures.go
tmroeder: no, that's fine

PR #13, opened 2018-05-28
Add TPM_ForceClear command

zaolin:
* Needed for some platforms.
* Normally only used in physical presence mode.

### DIFF
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -1008,11 +1008,13 @@ func PcrReset(rw io.ReadWriter, pcrs []int) error {
 
 // ForceClear is normally used by firmware but on some platforms
 // vendors got it wrong and didn't call TPM_DisableForceClear.
-func ForceClear(rw io.ReadWriter) (uint32, error) {
-	ret, err := submitTPMRequest(rw, tagRQUCommand, ordForceClear)
+func ForceClear(rw io.ReadWriter) error {
+	in := []interface{}{}
+	out := []interface{}{}
+	_, err := submitTPMRequest(rw, tagRQUCommand, ordForceClear, in, out)
 	if err != nil {
-		return 0, err
+		return err
 	}
 
-	return ret, nil
+	return nil
 }
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -1008,13 +1008,11 @@ func PcrReset(rw io.ReadWriter, pcrs []int) error {
 
 // ForceClear is normally used by firmware but on some platforms
 // vendors got it wrong and didn't call TPM_DisableForceClear.
+// It removes forcefully the ownership of the TPM.
 func ForceClear(rw io.ReadWriter) error {
 	in := []interface{}{}
 	out := []interface{}{}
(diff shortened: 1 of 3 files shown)


tpm/tpm.go
```diff
@@ -1005,3 +1005,16 @@ func PcrReset(rw io.ReadWriter, pcrs []int) error {
...
 	return nil
 }
+
+// ForceClear is normally used by firmware but on some platforms
+// vendors got it wrong and didn't call TPM_DisableForceClear.
+func ForceClear(rw io.ReadWriter) error {
+	in := []interface{}{}
+	out := []interface{}{}
+	_, err := submitTPMRequest(rw, tagRQUCommand, ordForceClear, in, out)
+	if err != nil {
```
insomniacslk: this line and the following could simply be `return err`

tmroeder: Agreed. Please fix.

tpm/tpm_test.go
```diff
@@ -498,3 +498,15 @@ func TestTakeOwnership(t *testing.T) {
 		t.Fatal("Couldn't take ownership of the TPM:", err)
 	}
 }
+
+func TestForceClear(t *testing.T) {
+	// Only enable this if you know what you're doing.
```
tmroeder: Please add this descriptive comment and warning to the Skip statement to make it clear that users generally will not want to enable this.

tpm/tpm.go
```diff
@@ -1005,3 +1005,16 @@ func PcrReset(rw io.ReadWriter, pcrs []int) error {
 	}
 	return nil
 }
+
+// ForceClear is normally used by firmware but on some platforms
```
tmroeder: In addition to the explanation of why this command is useful to have in this library, please mention briefly the expected effect of calling this function (for completeness, since the caller can probably guess from the name).

tpm/tpm_test.go
```diff
@@ -498,3 +498,15 @@ func TestTakeOwnership(t *testing.T) {
...
+	t.Skip()
+	rwc := openTPMOrSkip(t)
+	defer rwc.Close()
+
+	// This test code assumes that the owner auth is the well-known value.
```
tmroeder: I don't think this comment is accurate: TPM_ForceClear always succeeds if TPM_DisableForceClear has not been called, and it doesn't perform any owner auth checks, according to Section 6.3 of [the TPM 1.2 spec, part 3](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Main-Part-3-Commands_v1.2_rev116_01032011.pdf).

So, maybe something like "Note that ForceClear does not check owner auth." Or just drop the comment here and make sure the description in the Skip is good.

tpm/tpm.go
zaolin: Done

tpm/tpm.go
zaolin: Done

tpm/tpm_test.go
zaolin: Done

tpm/tpm_test.go
zaolin: Done, leftover

PR #14, opened 2018-06-05
Add an example binary for reading EK cert from NVRAM

awly:


awly:
@tmroeder could you take a quick look?

cc @joshzimmerman

examples/tpm2-ekcert/main.go
```diff
@@ -0,0 +1,59 @@
...
+	"os"
+
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
+)
+
+var (
+	tpmPath   = flag.String("tpm-path", "/dev/tpm0", "Path to the TPM device (character device or a Unix socket)")
+	certIndex = flag.Uint("cert-index", 0, "NVRAM index of the certificate file")
+	outPath   = flag.String("output", "", "File path to for output. Leave blank to write to stdout.")
```
tmroeder: "to for" -> "for"

examples/tpm2-ekcert/main.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func main() {
+	flag.Parse()
+
+	if *certIndex == 0 {
+		fmt.Println("--cert-index must be set")
+		os.Exit(2)
```
tmroeder: Why 2 instead of 1? Is there a convention that this is meant to follow? AFAIK, 1 is the standard generic failure exit, and 2 has special meanings in some cases but not most.

examples/tpm2-ekcert/main.go
```diff
@@ -0,0 +1,59 @@
...
+	tpmPath   = flag.String("tpm-path", "/dev/tpm0", "Path to the TPM device (character device or a Unix socket)")
+	certIndex = flag.Uint("cert-index", 0, "NVRAM index of the certificate file")
+	outPath   = flag.String("output", "", "File path to for output. Leave blank to write to stdout.")
+)
+
+func main() {
+	flag.Parse()
+
+	if *certIndex == 0 {
+		fmt.Println("--cert-index must be set")
```
tmroeder: Consider printing errors to stderr:

fmt.Fprintln(os.stderr, "--cert-index must be set")

examples/tpm2-ekcert/main.go
awly: Done

awly:
Also set the default NVRAM index for EK cert, per https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf

PTAL

examples/tpm2-ekcert/main.go
```diff
@@ -0,0 +1,61 @@
...
+	// Default value is defined in section 7.8
+	// https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf
+	certIndex = flag.Uint("cert-index", 0x01C00002, "NVRAM index of the certificate file")
+	outPath   = flag.String("output", "", "File path for output. Leave blank to write to stdout.")
+)
+
+func main() {
+	flag.Parse()
+
+	if *certIndex == 0 {
```
tmroeder: Now that a sensible default has been set, I'm not sure this check is relevant anymore. It was checking for 0 in the past because that was the default in the flag. How about just removing this?

examples/tpm2-ekcert/main.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
+)
+
+var (
+	tpmPath = flag.String("tpm-path", "/dev/tpm0", "Path to the TPM device (character device or a Unix socket)")
+	// Default value is defined in section 7.8
+	// https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf
+	certIndex = flag.Uint("cert-index", 0x01C00002, "NVRAM index of the certificate file")
+	outPath   = flag.String("output", "", "File path for output. Leave blank to write to stdout.")
```
tmroeder: tiny nit: inconsistent flag help text: there's a period at the end of the text for outPath but not for the others. I think the standard style is to not have punctuation at the end of flag help text.

examples/tpm2-ekcert/main.go
awly: You're right, it's unnecessary now. Removed

examples/tpm2-ekcert/main.go
awly: Done

PR #15, opened 2018-06-06
Update link to TCG documentation

joshzimmerman:
This links to the landing page for the "Provisioning Guidance" document rather than the version-specific PDF. This way, if the TCG makes updates to the document, the comment will point people to the most recent version.


examples/tpm2-ekcert/main.go
```diff
@@ -15,7 +15,7 @@ import (
 var (
 	tpmPath = flag.String("tpm-path", "/dev/tpm0", "Path to the TPM device (character device or a Unix socket)")
 	// Default value is defined in section 7.8
```
tmroeder: nit: please also update this line to make it clear that 7.8 is the section not of the linked page but of the sole version-specific PDF that can be found at this page.

PR #16, opened 2018-06-11
explicit signature scheme and hash algorithm

rautammi:
At least my Infineon OPTIGA SLB9670 requires signature scheme and hash algorithm to be stated explicitly.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -976,7 +976,11 @@ func Shutdown(rw io.ReadWriter, typ StartupType) error {
 	return err
 }
 
-func encodeSign(key tpmutil.Handle, password string, digest []byte) ([]byte, error) {
+func encodeSign(key tpmutil.Handle, password string, digest []byte, sigScheme *SigScheme) ([]byte, error) {
+	if sigScheme == nil {
+		sigScheme = &SigScheme{Alg: AlgNull}
+	}
+
 	ha, err := tpmutil.Pack(key)
 	if err != nil {
 		return nil, err
@@ -985,11 +989,20 @@ func encodeSign(key tpmutil.Handle, password string, digest []byte) ([]byte, err
 	if err != nil {
 		return nil, err
 	}
-	params, err := tpmutil.Pack(digest, AlgNull, tagHashCheck, HandleNull, []byte(nil))
+	d, err := tpmutil.Pack(digest)
 	if err != nil {
 		return nil, err
 	}
-	return concat(ha, auth, params)
+	s, err := sigScheme.encode()
+	if err != nil {
+		return nil, err
+	}
+	params, err := tpmutil.Pack(HandleNull, []byte(nil))
+	if err != nil {
+		return nil, err
+	}
+
+	return concat(ha, auth, d, s, params)
 }
 
 func decodeSign(buf []byte) (*Signature, error) {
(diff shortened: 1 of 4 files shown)


rautammi:
I signed it!

awly:
@rautammi is the email you used for commit (`rauno.tamminen@insta.fi`) same as the one you used to sign the CLA and used for your GitHub account?

tpm2/tpm2.go
```diff
@@ -985,7 +985,7 @@ func encodeSign(key tpmutil.Handle, password string, digest []byte) ([]byte, err
 	if err != nil {
 		return nil, err
 	}
-	params, err := tpmutil.Pack(digest, AlgNull, tagHashCheck, HandleNull, []byte(nil))
+	params, err := tpmutil.Pack(digest, sigScheme, hashAlg, tagHashCheck, HandleNull, []byte(nil))
```
awly: This only works for RSA, HMAC and some ECC signing schemes.
Anonymous ECC schemes also take a count (see section 11.2.1.3 in https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf).

Ideally, switch on `sigScheme`:
- for `AlgNull` do nothing
- for `AlgECDAA` add hash and count
- for the rest add a hash

rautammi: So, sigScheme should be an interface, like crypto.SignerOpts, which implements HashFunc() and , in case of AlgECDAA, also Count() ?

awly: If you want to abstract it, make it like https://github.com/google/go-tpm/blob/master/tpm2/structures.go#L72
Have a dedicated type and switch within to decide how to encode.

Or just add a switch here. Up to you.

rautammi:
My bad with messing up with author email addresses. I may have to close this pull request and create a new one with correct email addresses.

But now there's the switch added. I am not convenient with bloating the function interface with seldom needed parameters, but I am lacking a vision of a really elegant solution.

tpm2/tpm2.go
```diff
@@ -1003,8 +1013,8 @@ func decodeSign(buf []byte) (*Signature, error) {
 
 // Sign computes a signature for digest using a given loaded key. Signature
 // algorithm depends on the key type.
-func Sign(rw io.ReadWriter, key tpmutil.Handle, password string, digest []byte) (*Signature, error) {
-	cmd, err := encodeSign(key, password, digest)
+func Sign(rw io.ReadWriter, key tpmutil.Handle, password string, digest []byte, sigScheme, hashAlg Algorithm, count uint32) (*Signature, error) {
```
awly: Ah, I totally forgot, `TPMT_SIG_SCHEME` is already (almost) implemented: https://github.com/google/go-tpm/blob/master/tpm2/structures.go#L256

Could you add `count` field (and handling of ECDAA) to it instead and use a `*SigScheme` argument for `Sign`?

awly:
To fix the email you can squash your commits instead and use the correct email there.
Something like
```
git reset --soft 4e28e10ca34602b0156c61a478b61499bce142d5
git config --global user.email "email@example.com"
git add .
git commit
git push --force
```

tpm2/tpm2.go
tmroeder: I agree. Using SigScheme in Sign sounds like the right choice for the API.

tpm2/tpm2.go
```diff
@@ -976,7 +976,11 @@ func Shutdown(rw io.ReadWriter, typ StartupType) error {
 	return err
 }
 
-func encodeSign(key tpmutil.Handle, password string, digest []byte, sigScheme, hashAlg Algorithm, count uint32) ([]byte, error) {
+func encodeSign(key tpmutil.Handle, password string, digest []byte, sigScheme *SigScheme) ([]byte, error) {
+	if sigScheme == nil {
```
awly: This shouldn't be necessary. `SigScheme.encode` handles nil

tpm2/tpm2.go
```diff
@@ -976,7 +976,11 @@ func Shutdown(rw io.ReadWriter, typ StartupType) error {
 	return err
 }
 
-func encodeSign(key tpmutil.Handle, password string, digest []byte) ([]byte, error) {
+func encodeSign(key tpmutil.Handle, password string, digest []byte, sigScheme *SigScheme) ([]byte, error) {
+	if sigScheme == nil {
```
awly: Remove this check, it's handled by SigScheme.encode

awly:
Thanks for fixing the CLA. One last comment and LGTM

PR #18, opened 2018-06-22
add tagHashCheck

atotto:
We need TPMT_TK_HASHCHECK for TPM_ALG_NULL scheme.

refs #16 cc @rautammi 

example our tpm2 persistent (SLB9670):

```
$ sudo -E tpm2_listpersistent
persistent-handle[0]:0x81000001 key-alg:rsa hash-alg:sha256 object-attr:sensitivedataorigin|userwithauth|sign
```

I wrote some test for encodeSign, but AlgSHA256 test behavior does not check on our tpm device. 

Thanks.


tpm2/encoding_test.go
```diff
@@ -371,3 +371,38 @@ func TestECCParamsEncodeDecode(t *testing.T) {
 		t.Fatalf("got: %+v\nwant: %+v", got, params)
 	}
 }
+
+func TestSignEncode(t *testing.T) {
+	t.Run("AlgNone", func(t *testing.T) {
+		testCmdBytes, err := hex.DecodeString("800000010000000d40000009000001000401020304000301020300108024400000070000")
```
awly: How were these strings generated? Could you add a comment in case we need to update it in the future?

tpm2/encoding_test.go
```diff
@@ -371,3 +371,38 @@ func TestECCParamsEncodeDecode(t *testing.T) {
 		t.Fatalf("got: %+v\nwant: %+v", got, params)
 	}
 }
+
+func TestSignEncode(t *testing.T) {
+	t.Run("AlgNone", func(t *testing.T) {
```
awly: `AlgNull`

atotto:
also I can trace write string with strace tool,

```
sudo strace -x -s 1024 -e trace=write tpm2_sign -k 0x81000001 -m example.txt -s out.bin -g sha256
```

tpm2/encoding_test.go
```diff
@@ -371,3 +371,39 @@ func TestECCParamsEncodeDecode(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !bytes.Equal(cmdBytes, testCmdBytes) {
+			t.Fatalf("got: %s, want: %s", hex.EncodeToString(cmdBytes), hex.EncodeToString(testCmdBytes))
+		}
+	})
+
+	t.Run("AlgSHA256", func(t *testing.T) {
+		// TODO: this test string generated by own test. please see https://www.trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf 20.2.2 Command and Respons.
```
awly: Not sure that we need "TODO", there's nothing actionable here.
How about: `// The expected output blobs were recorded while successfully running Sign against a real TPM device.`

awly:
LGTM
Ping @tmroeder for merge

tpm2/encoding_test.go
```diff
@@ -371,3 +371,39 @@ func TestECCParamsEncodeDecode(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		digest := []byte{0x01, 0x02, 0x03}
+		cmdBytes, err := encodeSign(tpmutil.Handle(0x80000001), defaultPassword, digest, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !bytes.Equal(cmdBytes, testCmdBytes) {
+			t.Fatalf("got: %s, want: %s", hex.EncodeToString(cmdBytes), hex.EncodeToString(testCmdBytes))
```
tmroeder: Please add a little more context to the error string here and below. I definitely want to know what the code got and what it wanted, but the context can be invaluable for debugging. E.g., in this case, something like

"encoded Sign message did not match golden value. got: ...."

tpm2/encoding_test.go
```diff
@@ -371,3 +371,39 @@ func TestECCParamsEncodeDecode(t *testing.T) {
...
+		}
+	})
+
+	t.Run("AlgSHA256", func(t *testing.T) {
+		// The expected output blobs were recorded while successfully running Sign against a real TPM device.
```
tmroeder: This comment should probably be on the first output blob rather than waiting until the second.

PR #19, opened 2018-06-22
Check that EK cert matches EK via public key

awly:


awly:
@tmroeder please take a look, not urgent

awly:
@joshzimmerman please take a look, not urgent

awly:
Added default RSA EK template if `-template-index` isn't specified

joshzimmerman:
LGTM. Thanks.

examples/tpm2-ekcert/main.go
```diff
@@ -38,19 +42,70 @@ func main() {
...
+		ekTemplate, err := tpm2.NVRead(rwc, tpmutil.Handle(tmplIdx))
+		if err != nil {
+			return nil, fmt.Errorf("reading EK template: %v", err)
+		}
+		ekh, ekPub, err = tpm2.CreatePrimaryRawTemplate(rwc, tpm2.HandleEndorsement, tpm2.PCRSelection{}, "", "", ekTemplate)
+		if err != nil {
+			return nil, fmt.Errorf("creating EK: %v", err)
+		}
+	} else {
+		// Default EK template defined in:
```
tmroeder: I think it would be clearer to pull the default template out to the top as a var declaration, since it doesn't need to be embedded in the code here and is of independent interest if it's part of the spec.

tpm2/structures.go
```diff
@@ -104,13 +104,18 @@ func decodePublic(in *bytes.Buffer) (Public, error) {
 // RSAParams represents parameters of an RSA key pair.
 //
 // Symmetric and Sign may be nil, depending on key Attributes in Public.
-// Modulus must always be non-nil.
+//
+// One of Modulus and ModulusRaw must always be non-nil. Modulus takes
+// precedence.
```
tmroeder: nit: I don't the extra newline is needed here.

tpm2/structures.go
```diff
@@ -104,13 +104,18 @@ func decodePublic(in *bytes.Buffer) (Public, error) {
...
+// ModulusRaw is used for key templates where "unique" field must be a byte
```
tmroeder: Please clarify the meaning of the quotes here. Is the name of the field "unique"? If so, then consider rewriting the text as 'where the field named "unique" must be...'. 

Or if this is talking about key templates that have exactly one field, then please drop the quotes and write 'where the unique field must be...'. It might be clearer in this case to write 'where there must be only a single field, consisting of a byte array of all zeroes' or something along those lines.

tpm2/structures.go
```diff
@@ -125,14 +130,24 @@ func (p *RSAParams) encode() ([]byte, error) {
...
-	if p.Modulus == nil {
-		return nil, errors.New("RSAParams.Modulus must be set")
+	rest, err := tpmutil.Pack(p.KeyBits, p.Exponent)
+	if err != nil {
+		return nil, err
 	}
-	rest, err := tpmutil.Pack(p.KeyBits, p.Exponent, p.Modulus.Bytes())
+
+	if p.Modulus == nil && len(p.ModulusRaw) == 0 {
+		return nil, errors.New("RSAParams.Modulus or RSAParams.ModulusRaw must be set")
```
tmroeder: Is it also an error for both to be set? If so, then please check for this (instead of clobbering mod in the code below).

examples/tpm2-ekcert/main.go
awly: Done

tpm2/structures.go

tpm2/structures.go
awly: Done, it was a field named `unique` in the spec

tpm2/structures.go
awly: Yep, it's an error.
Not sure how i can clean up mod below.

examples/tpm2-ekcert/main.go
```diff
@@ -17,13 +20,41 @@ var (
 	// Default value is defined in section 7.8, "NV Memory" of the latest version pdf on:
 	// https://trustedcomputinggroup.org/resource/tcg-tpm-v2-0-provisioning-guidance/
 	certIndex = flag.Uint("cert-index", 0x01C00002, "NVRAM index of the certificate file")
+	tmplIndex = flag.Uint("template-index", 0, "NVRAM index of the EK template; if empty, default RSA EK template is used")
```
tmroeder: One more minor nit: "empty" doesn't make sense to me in the context of an integer flag. I was going to suggest "if not set", but that's not true either, according to the code. It looks like it should say "if non-zero"

awly: Done

PR #20, opened 2018-06-27
Cleanup maxTPMResponse variable

josephlr:
The maxTPMResponse variable is only used by the tpmutil package, so its comments should be there as well.

I looked through the TCG specification and was unable to find any firm statement that the maxiumum response was 4096. Does anyone have the citation to the TPM1 or TPM2 spec?

### DIFF
--- a/tpm/constants.go
+++ b/tpm/constants.go
@@ -158,12 +158,6 @@ const (
 	authPrivUseOnly byte = 0x03
 )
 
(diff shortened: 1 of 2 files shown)


awly:
AFAIK 4096 is the maximum size of the TPM I/O buffer.
But I can't find it defined anywhere in the spec either.

PR #21, opened 2018-06-27
Add support for TPM devices over TCP sockets

awly:
One such example is the IBM emulator: https://sourceforge.net/projects/ibmswtpm2
Limiting this to localhost:port only, out of paranoia.

### DIFF
--- a/tpmutil/run.go
+++ b/tpmutil/run.go
@@ -25,37 +25,48 @@ import (
 	"io"
 	"net"
 	"os"
+	"strings"
 )
 
-// OpenTPM opens a channel to the TPM at the given path. If the file is a
-// device, then it treats it like a normal TPM device, and if the file is a
-// Unix domain socket, then it opens a connection to the socket.
+const localSocketPrefix = "localhost:"
+
+// OpenTPM opens a channel to the TPM at the given path.
+//
+// If path is a local TCP socket (i.e. starts with "localhost:"), it's dialed
+// and returned.
+// If path is a unix socket, it's dialed and returned.
+// If path is a device file, then it's opened and returned.
 func OpenTPM(path string) (io.ReadWriteCloser, error) {
-	// If it's a regular file, then open it
-	var rwc io.ReadWriteCloser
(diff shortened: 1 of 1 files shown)


awly:
@tmroeder

tpmutil/run.go
```diff
@@ -25,37 +25,48 @@ import (
...
+// If path is a local TCP socket (i.e. starts with "localhost:"), it's dialed
+// and returned.
+// If path is a unix socket, it's dialed and returned.
+// If path is a device file, then it's opened and returned.
 func OpenTPM(path string) (io.ReadWriteCloser, error) {
-	// If it's a regular file, then open it
-	var rwc io.ReadWriteCloser
+	// If it's a local TCP socket, dial it.
+	if strings.HasPrefix(path, localSocketPrefix) {
+		con, err := net.Dial("tcp", path)
```
tmroeder: Go is all for short names, so maybe this could be `c`. But otherwise, I'd like to use `conn` for this (two n's), since that's the spelling used, e.g., in net.TCPConn. Or see the first example in the net package (https://golang.org/pkg/net/)

tpmutil/run.go
```diff
@@ -25,37 +25,48 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+		return con, nil
```
tmroeder: I don't think this is enough to communicate with the TCP socket provided by the IBM SW TPM 2.0. There is also a framing protocol that wraps communication with this port. See the file TpmTcpProtocol.h in the source code for the SW TPM.

Lines 64-65 say:

```c
/* TPM commands are communicated as BYTE streams on a TCP connection.  The TPM command protocol is
   enveloped with the interface protocol described in this file.
```

awly:
Since this is a custom protocol and not part of the spec, I'm inclined to abandon this.
Users can:
- use an emulator that can expose a Unix socket
- implement a custom `io.ReadWriteCloser` doing the emulator-specific wrapping

PR #22, opened 2018-07-03
Remove extraneous null handle argument to commands.

nlehrer1:


tpm2/tpm2.go
```diff
@@ -251,7 +251,7 @@ func encodePasswordAuthArea(passwords ...string) ([]byte, error) {
 }
 
 func encodePCREvent(pcr tpmutil.Handle, eventData []byte) ([]byte, error) {
-	ha, err := tpmutil.Pack(pcr, HandleNull)
+	ha, err := tpmutil.Pack(pcr)
```
tmroeder: Please add some more detail to the commit message. What was the bug and how does this fix it?

Also, I realize that the tests on this repo are not very robust, since they mostly require access to a TPM 1.2 and TPM 2.0 to make them work. But if there's a regression test you can add here, then please do.

nlehrer1: I'm going to re-submit this under a different github account...i'll re-create your comment when I do

PR #24, opened 2018-07-12
Fix ECC Sign test for tpm2

awly:
SigScheme passed to Sign must match the scheme in signing key (in pub).
ECC test was failing because of this.


awly:
@tmroeder please take a look
cc @nathaniellehrer

PR #25, opened 2018-07-13
Remove extraneous null handle argument.

nathaniellehrer:
This extraneous null handle argument is not present in the TPM2_PCR_Event
command spec and causes the command to return a 0x95 response code
(TPM_RC_SIZE).

### DIFF
--- a/tpm2/tpm2_test.go
+++ b/tpm2/tpm2_test.go
@@ -558,3 +558,14 @@ func TestSign(t *testing.T) {
 		})
 	})
 }
+
+func TestPCREvent(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+	debugPCR := uint32(16)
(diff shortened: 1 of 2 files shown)


tpm2/tpm2_test.go
```diff
@@ -558,3 +558,14 @@ func TestSign(t *testing.T) {
 		})
 	})
 }
+
+func TestPCREvent(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+	debugPCR := uint32(16)
+	arbitraryBytes := []byte{1}
+	err := PCREvent(rw, tpmutil.Handle(debugPCR), arbitraryBytes)
```
awly: merge this with error check

awly:
small nit, otherwise LGTM
@tmroeder for merge (we should probably have more people able to merge PRs)

PR #33, opened 2018-07-27
Allow nil slices to be used with the TBS library

josephlr:
The existing code will panic if any nil slices are passed into the tpmutil/tbs library. This is especially bad as GetTCGLog has useful behavior if a zero length buffer is passed.

This change allows nil slices to be passed for all tbs methods. It also tests that we can get the TCG log by either
  - passing in a progressivly larger buffer
  - passing in nil to get the exact size

We also update the documentation to describe exactly what happens when a nil slice is passed. Microsoft's documentation is incomple, all the relevant parameters just say "TBD", see https://github.com/MicrosoftDocs/feedback/issues/551.

Finally, I added tests to make sure that the documented behavior actually happens.
Here's the tpmutil/tbs test output for a GCE VM:
```
PS C:\Users\magendanz\Documents\joe-go-tpm\tpmutil\tbs> go test -v
=== RUN   TestGetLogLargeBuffer
--- PASS: TestGetLogLargeBuffer (0.00s)         
=== RUN   TestGetLogWithNilSlice                
--- PASS: TestGetLogWithNilSlice (0.00s)        
=== RUN   TestSubmitCommandNilCommand           
--- PASS: TestSubmitCommandNilCommand (0.00s)   
=== RUN   TestSubmitCommandNilResponse          
--- PASS: TestSubmitCommandNilResponse (0.00s)  
=== RUN   TestSubmitCommandShortResponse        
--- PASS: TestSubmitCommandShortResponse (0.00s)
=== RUN   TestSubmitCommandLongResponse         
--- PASS: TestSubmitCommandLongResponse (0.03s) 
PASS
ok      _/C_/Users/magendanz/Documents/joe-go-tpm/tpmutil/tbs   0.114s 
```

### DIFF
--- a/tpmutil/tbs/tbs_windows.go
+++ b/tpmutil/tbs/tbs_windows.go
@@ -173,9 +173,10 @@ func (context Context) Close() error {
 	return getError(result)
 }
 
-// SubmitCommand sends commandBuffer to the TPM, returning the number of bytes written
-// to responseBuffer. If responseBuffer is too small (ErrInsufficientBuffer), the required
-// size is returned.
+// SubmitCommand sends commandBuffer to the TPM, returning the number of bytes
+// written to responseBuffer. ErrInsufficientBuffer is returned if the
+// responseBuffer is too short. ErrInvalidOutputPointer is returned if the
+// responseBuffer is nil. On failure, the returned length is unspecified.
 // https://docs.microsoft.com/en-us/windows/desktop/api/Tbs/nf-tbs-tbsip_submit_command
 func (context Context) SubmitCommand(
 	priority CommandPriority,
@@ -206,8 +207,9 @@ func (context Context) SubmitCommand(
 }
 
 // GetTCGLog gets the system event log, returning the number of bytes written
-// to responseBuffer. If logBuffer is nil, the required size and
-// ErrInsufficientBuffer are returned.
+// to logBuffer. If logBuffer is nil, the size of the TCG log is returned.
+// ErrInsufficientBuffer is returned if the logBuffer is too short. On failure,
+// the returned length is unspecified.
 // https://docs.microsoft.com/en-us/windows/desktop/api/Tbs/nf-tbs-tbsi_get_tcg_log
 func (context Context) GetTCGLog(logBuffer []byte) (uint32, error) {
 	logBufferLen := uint32(len(logBuffer))
--- a/tpmutil/tbs/tbs_windows.go
+++ b/tpmutil/tbs/tbs_windows.go
@@ -136,6 +136,14 @@ var (
 	tbsGetTCGLog     = tbsDLL.NewProc("Tbsi_Get_TCG_Log")
 )
 
+// Returns the address of the beginning of a slice or 0 for a nil slice.
+func sliceAddress(s []byte) uintptr {
+	if len(s) == 0 {
+		return 0
+	}
+	return uintptr(unsafe.Pointer(&(s[0])))
+}
+
 // CreateContext creates a new TPM context:
 // https://docs.microsoft.com/en-us/windows/desktop/api/Tbs/nf-tbs-tbsi_context_create
 func CreateContext(version Version, flag Flag) (Context, error) {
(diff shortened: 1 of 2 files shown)


tpmutil/tbs/tbs_windows_test.go
```diff
@@ -0,0 +1,54 @@
...
+		if err == nil {
+			if logLen == 0 {
+				t.Fatalf("Expected positive TCGLog length")
+			}
+			return
+		}
+		if err != ErrInsufficientBuffer {
+			t.Fatalf("GetTCGLog failed: err = %v", err)
+		}
+		log = append(log, make([]byte, len(log))...)
```
awly: `log = make([]byte, 2*len(log))`

tpmutil/tbs/tbs_windows.go
```diff
@@ -189,17 +197,17 @@ func (context Context) SubmitCommand(
 		uintptr(context),
 		uintptr(commandLocalityZero),
 		uintptr(priority),
-		uintptr(unsafe.Pointer(&(commandBuffer[0]))),
+		sliceAddress(commandBuffer),
```
awly: Is there any command where passing nil commandBuffer is valid?
If not, I'd rather check that before calling the command and return the error directly

tpmutil/tbs/tbs_windows_test.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// Get the log by passing in nil, checking the size, and then getting the log.
+func TestGetLogWithNilSlice(t *testing.T) {
+	ctx := getContext(t)
+	defer ctx.Close()
+
+	logLen, err := ctx.GetTCGLog(nil)
+	if err != ErrInsufficientBuffer {
+		t.Fatalf("First GetTCGLog failed: err = %v", err)
```
awly: Add `expected ErrInsufficientBuffer`

tpmutil/tbs/tbs_windows_test.go
```diff
@@ -0,0 +1,54 @@
+package tbs
+
+import (
+	"os"
+	"testing"
+)
+
+func getContext(t *testing.T) Context {
+	ctx, err := CreateContext(TPMVersion20, IncludeTPM12|IncludeTPM20)
```
awly: Are all Windows machines guaranteed to support this?
We should maybe not fail the test when running on a device without a TPM, like all the other integration tests

quinnmagendanz:
Can you add a text file showing the tests passing?

tpmutil/tbs/tbs_windows_test.go
josephlr: Done

tpmutil/tbs/tbs_windows_test.go
josephlr: Done (here and elsewhere)

tpmutil/tbs/tbs_windows_test.go
josephlr: `getContext()` now skips instead of failing (this matches what we do for the tpm library's integration tests)

tpmutil/tbs/tbs_windows.go
josephlr: The comments and tests should clarify what happens when any of the slices in question are nil. I think it's fine for us to just pass along whatever buffer we are given, and let the syscall do the error checking. The TBS library is supposed to just be a safe wrapper around the DLL.

tpmutil/tbs/tbs_windows_test.go
```diff
@@ -1,54 +1,114 @@
...
+import (
+	"bytes"
+	"os"
+	"testing"
+)
+
+
+var (
+	// Encodes a call to Getrandom with a length of 0
+	encodedCommand = []byte{128, 1, 0, 0, 0, 12, 0, 0, 1, 123, 0, 0}
```
awly: Move these vars into the test func that uses them

tpmutil/tbs/tbs_windows_test.go
```diff
@@ -1,54 +1,114 @@
...
+
+// Make sure SubmitCommand can handle a nil command buffer.
+func TestSubmitCommandNilCommand(t *testing.T) {
+	ctx := getContext(t)
+	defer ctx.Close()
+
+	response := make([]byte, os.Getpagesize())
+	_, err := ctx.SubmitCommand(NormalPriority, nil, response)
+	if err != ErrBadParameter {
+		t.Fatalf("Expected ErrBadParameter from Submit Command: got %v", err)
```
awly: Output "got" before "expected"

tpmutil/tbs/tbs_windows_test.go
```diff
@@ -1,54 +1,114 @@
...
+
+import (
+	"bytes"
+	"os"
+	"testing"
+)
+
+
+var (
+	// Encodes a call to Getrandom with a length of 0
```
awly: So this always outputs "random" value 0?

josephlr: I changed the comments to make this clearer. Essentially it always outputs a "random" buffer of `[]`.

tpmutil/tbs/tbs_windows_test.go
awly: The encoded response is only used in 1 test, move it there.
And could you also renamed `encodedTestCommand` to `getRandomRawCommand` to be more obvious?

josephlr: Done

PR #34, opened 2018-08-03
Add a readme in examples directory

awly:
Explain that examples are specific to TPM device versions.
Rename genaik to tpm-genaik to match the rest.


examples/README.md
```diff
@@ -0,0 +1,12 @@
+# go-tpm usage examples
+
+This directory contains binaries demonstrating usage of go-tpm.
```
tmroeder: [optional nit]: This sentence could be simplified and made clearer at the same time:

"This directory contains binaries that show how to use go-tpm."

examples/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Versions
+
+All directories that start with `tpm-` are for TPM 1.x devices.
```
tmroeder: [optional]: another simplification: drop the "All" from each sentence:

"Directories that start with "...

examples/README.md
awly: Done

examples/README.md
awly: Done

PR #37, opened 2018-10-11
Implement TPM2_EncryptDecrypt2

awly:
Symmetric encryption and decryption. Add convenience chopping logic in
case processed data can't fit into one command.

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -28,6 +28,11 @@ func init() {
 	tpmutil.UseTPM20LengthPrefixSize()
 }
 
+// MAX_DIGEST_BUFFER is the maximum size of []byte request or response fields.
+// Typically used for chunking of big blobs of data (such as for hashing or
+// encryption).
+const maxDigestBuffer = 1024
+
 // Algorithm represents a TPM_ALG_ID value.
 type Algorithm uint16
 
@@ -61,6 +66,7 @@ const (
 	AlgECDAA     Algorithm = 0x001A
 	AlgKDF2      Algorithm = 0x0021
 	AlgECC       Algorithm = 0x0023
+	AlgSymCipher Algorithm = 0x0025
 	AlgCTR       Algorithm = 0x0040
 	AlgOFB       Algorithm = 0x0041
 	AlgCBC       Algorithm = 0x0042
@@ -216,6 +222,7 @@ const (
 	cmdUnseal             tpmutil.Command = 0x0000015E
 	cmdContextLoad        tpmutil.Command = 0x00000161
 	cmdContextSave        tpmutil.Command = 0x00000162
+	cmdEncryptDecrypt     tpmutil.Command = 0x00000164
 	cmdFlushContext       tpmutil.Command = 0x00000165
 	cmdLoadExternal       tpmutil.Command = 0x00000167
 	cmdMakeCredential     tpmutil.Command = 0x00000168
@@ -231,6 +238,7 @@ const (
 	cmdPCRExtend          tpmutil.Command = 0x00000182
 	cmdPolicyGetDigest    tpmutil.Command = 0x00000189
 	cmdPolicyPassword     tpmutil.Command = 0x0000018C
+	cmdEncryptDecrypt2    tpmutil.Command = 0x00000193
 )
 
 // Regular TPM 2.0 devices use 24-bit mask (3 bytes) for PCR selection.
@@ -259,3 +267,5 @@ var hashConstructors = map[Algorithm]func() hash.Hash{
 	AlgSHA384: sha512.New384,
 	AlgSHA512: sha512.New,
 }
+
+const rcCommandNotSupported = tpmutil.ResponseCode(0x143)
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -60,8 +60,9 @@ type Public struct {
 	// If Type is AlgKeyedHash, then do not set these.
 	// Otherwise, only one of the Parameters fields should be set. When encoding/decoding,
 	// one will be picked based on Type.
-	RSAParameters *RSAParams
-	ECCParameters *ECCParams
+	RSAParameters       *RSAParams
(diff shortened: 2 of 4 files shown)


tpm2/tpm2.go
```diff
@@ -1131,3 +1131,85 @@ func ReadPCR(rw io.ReadWriteCloser, pcr int, hashAlg Algorithm) ([]byte, error)
...
+	var out, block []byte
+	var err error
+
+	for rest := data; len(rest) > 0; {
+		if len(rest) > maxDigestBuffer {
+			block, rest = rest[:maxDigestBuffer], rest[maxDigestBuffer:]
+		} else {
+			block, rest = rest, nil
+		}
+		block, iv, err = encryptDecryptBlockSymmetric(rw, defaultPassword, key, iv, block, decrypt)
```
tmroeder: The Kokoro presubmit says:

github/go/src/github.com/google/go-tpm/tpm2/tpm2.go:1207:53: undefined: defaultPassword

Please fix.

awly: Oops, fixed

tpm2/tpm2.go
```diff
@@ -1131,3 +1131,85 @@ func ReadPCR(rw io.ReadWriteCloser, pcr int, hashAlg Algorithm) ([]byte, error)
...
+
+// EncryptSymmetric encrypts data using a symmetric key. iv is the initial
+// value and must match the cipher block size.
+//
+// data may be longer than block size, EncryptSymmetric will chain multiple TPM
+// calls to encrypt the entire blob.
+//
+// Key handle should point at SymCipher object which is a child of the key (and
+// not e.g. RSA key itself).
+func EncryptSymmetric(rw io.ReadWriteCloser, keyAuth string, key tpmutil.Handle, iv, data []byte) ([]byte, error) {
```
tmroeder: I realize that this is just exposing a function that is exposed by the TPM, and maybe we should expose that for libraries to build on. However, as someone who regularly deals in cryptographic APIs, this worries me, because these are very low-level primitives, and they're generally not safe to use directly. At the very least, I'd like for these primitives to not be in the same part of the library as the primitives like sealing and attestation, which can be used safely directly.

E.g., see the common python cryptography library (cryptography.io), which exposes low-level primitives but puts them in a module called cryptography.hazmat and puts large warnings on the documentation pages.

tpm2/constants.go
```diff
@@ -28,6 +28,11 @@ func init() {
 	tpmutil.UseTPM20LengthPrefixSize()
 }
 
+// MAX_DIGEST_BUFFER is the maximum size of []byte request or response fields.
+// Typically used for chunking of big blobs of data (such as for hashing,
```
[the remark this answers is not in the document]

tpm2/structures.go
```diff
@@ -248,6 +253,37 @@ func decodeECCParams(in *bytes.Buffer) (*ECCParams, error) {
 	return &params, nil
 }
 
+// SymCipherParams represent parameters of a symmetric block cipher TPM object.
```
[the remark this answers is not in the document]

tpm2/tpm2.go
```diff
@@ -1131,3 +1131,85 @@ func ReadPCR(rw io.ReadWriteCloser, pcr int, hashAlg Algorithm) ([]byte, error)
 	}
 	return pcrVal, nil
 }
+
+// EncryptSymmetric encrypts data using a symmetric key. iv is the initial
```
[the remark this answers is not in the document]
awly: Agree, changed

tpm2/tpm2.go
```diff
@@ -1131,3 +1131,85 @@ func ReadPCR(rw io.ReadWriteCloser, pcr int, hashAlg Algorithm) ([]byte, error)
...
+// value and must match the cipher block size.
```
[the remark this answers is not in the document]

tpm2/tpm2.go
awly: Not sure I understand. Encryption is usually under the same tree as other cryptographic primitives (like hashing or RNG). For example Go stdlib makes no notable distinction (https://golang.org/pkg/crypto/cipher/).
Why is this not safe to use directly?

IMO, any TPM interaction is tricky and possible to get wrong. I think we should keep these libraries close to the actual commands.
One option is to add a typical disclaimer in the repo README about dangers of playing with crypto.

I'm also open to splitting this entire lib into `tpm2raw` which simply exposes method-per-command and `tpm2` which provides Go-friendly API with common use cases simplified. But symmetric encryption should still be available via higher-level lib.

tmroeder: TL;DR: I think the differentiation suggested between tpm2raw and tpm2 is good, and if you're OK with that split, then let's go that way. If we're going to expose a high-level API for encryption and decryption in the tpm2 library, it needs to be authenticated encryption, not unauthenticated AES-*, but I'm OK with exposing the raw TPM calls in a library called something like tpm2raw.

Sorry for the long discussion about this; this is something that I've spent a long time dealing with for the past decade at Microsoft and Google, so I'm very sensitized to the problems that can come up with these APIs.

Painful details:

Unauthenticated encryption primitives when used alone almost always fail to provide the properties that their users expect. E.g., a protocol that uses AES-CBC without some kind of MAC almost certainly fails to provide even basic confidentiality of data (except under extremely limited threat models), given known attacks on AES-CBC. So, providing an interface to let callers use AES-CBC directly, without a MAC, is bad API design: the function says it's encrypting the data, but the result is very likely not providing any of the properties a caller might expect from encryption.

The properties that users expect from encryption are provided by Authenticated Encryption (e.g., the AEAD primitives given in the Go standard library). Adam Langley (of BoringSSL) has a great discussion about this. See https://www.imperialviolet.org/2015/05/16/aeads.html. The TL;DR is that historical APIs provided this low-level access, and that turned out to be a serious mistake (though most people didn't realize it at the time).

awly:
This isn't needed anymore. Dropping the PR, rebasing and addressing comments isn't worth the effort right now.

PR #38, opened 2018-10-22
Implement support for Creation attestation, expose DecodePublic().

twitchy-jsonp:


ericchiang:
cc @awly @tmroeder

tpm2/structures.go
```diff
@@ -90,7 +90,9 @@ func (p Public) Encode() ([]byte, error) {
 	return concat(head, params)
 }
 
-func decodePublic(in *bytes.Buffer) (Public, error) {
+// DecodePublic decode a TPMT_PUBLIC message. No error is returned if
+// the input has extra trailing data.
+func DecodePublic(in *bytes.Buffer) (Public, error) {
```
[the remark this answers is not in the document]
twitchyliquid64: Will do!

tpm2/structures.go
```diff
@@ -90,18 +90,21 @@ func (p Public) Encode() ([]byte, error) {
 	return concat(head, params)
 }
 
-func decodePublic(in *bytes.Buffer) (Public, error) {
+// DecodePublic decode a TPMT_PUBLIC message. No error is returned if
+// the input has extra trailing data.
+func DecodePublic(in []byte) (Public, error) {
+	buf := bytes.NewBuffer(in)
```
awly: Name bytes.Buffer `in` to minimize the diff

tpm2/structures.go
```diff
@@ -438,20 +442,31 @@ func DecodeAttestationData(in []byte) (*AttestationData, error) {
...
 	}
+
 	return &ad, nil
 }
 
 // Encode serializes an AttestationData structure in TPM wire format.
 func (ad AttestationData) Encode() ([]byte, error) {
-	if ad.Type != TagAttestCertify {
-		return nil, fmt.Errorf("only Certify attestation structure is supported, got type 0x%x", ad.Type)
+	switch ad.Type {
```
awly: Remove this and add default case to the other switch below

tpm2/tpm2_test.go
```diff
@@ -644,3 +644,45 @@ func TestEncodeDecodeAttestationData(t *testing.T) {
...
+		},
+		ExtraData: []byte("foo"),
+		ClockInfo: ClockInfo{
+			Clock:        3,
+			ResetCount:   4,
+			RestartCount: 5,
+			Safe:         6,
+		},
+		FirmwareVersion:     7,
+		AttestedCertifyInfo: nil,
```
awly: Remove this line

tpm2/structures.go
```diff
@@ -464,13 +479,60 @@ func (ad AttestationData) Encode() ([]byte, error) {
...
+		}
 	}
+
 	return concat(head, signer, tail, info)
 }
 
+// CreationInfo contains Creation-specific data for TPMS_ATTEST.
+type CreationInfo struct {
+	Name         Name
+	OpaqueDigest []byte
```
awly: Should this be *HashValue?

twitchyliquid64: Negative. While most `TPM2B_Digest` structures contain a `TPMU_HA` structure, there is no requirement for them to, just the requirement that the structure cannot be longer than the longest digest size. (I spent 4 hours this morning finding this out the hard way 💀💀lol whelp)

tpm2/tpm2_test.go
twitchyliquid64: Will do!

tpm2/structures.go
twitchyliquid64: Will do!

tpm2/structures.go
twitchyliquid64: Will do!

twitchyliquid64:
PTAL!

tpm2/structures.go
ericchiang: Can you add a comment here explaining that? Do you need access to this field when it's something else?

Maybe this could be something like

```
type CreationInfo struct {
    Name Name

    // Digest of the creation info. While common to many implementations, a digest isn't required, and may
    // be nil if the digest data isn't of the correct form.
    Digest *HashValue
    // RawDigest is the unparsed digest. This will always be present and can be used when implementations
    // don't use a TPMU_HA for the digest to parse alternative structures.
    RawDigest []byte
}
```

tpm2/structures.go
```diff
@@ -464,13 +472,62 @@ func (ad AttestationData) Encode() ([]byte, error) {
...
 		return nil, fmt.Errorf("encoding ExtraData, ClockInfo, FirmwareVersion: %v", err)
 	}
-	info, err := ad.AttestedCertifyInfo.encode()
-	if err != nil {
-		return nil, fmt.Errorf("encoding AttestedCertifyInfo: %v", err)
+
+	var info []byte
+	switch ad.Type {
+	case TagAttestCertify:
+		info, err = ad.AttestedCertifyInfo.encode()
```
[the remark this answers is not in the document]

twitchy-jsonp:
PTAL! 🔍 🦄

twitchy-jsonp:
Friendly ping :)

awly:
LGTM

PR #39, opened 2018-10-25
Implement TPM2-spec KDFa() function

twitchyliquid64:
This is needed to compute keys for Credential Activation.


twitchyliquid64:
Wrong repo.

PR #40, opened 2018-10-25
Implement TPM2-spec KDFa() function

twitchy-jsonp:
Needed to compute keys for Credential Activation.


twitchy-jsonp:
I've added unit tests based on IBM's open-source soft TPM. PTAL! 🔍

twitchy-jsonp:
To answer your earlier question, the only post-test change added to KDFa() was:

```
go

if len(out) > bits/8 {
    out = out[:bits/8]
}
```

PTAL! 🌈

tpm2/kdf.go
```diff
@@ -0,0 +1,57 @@
...
+// The key & label parameters must not be zero length, but contextU &
+// contextV may be.
+func KDFa(hashAlg Algorithm, key []byte, label string, contextU, contextV []byte, bits int) ([]byte, error) {
+	var counter uint32
+	remaining := (bits + 7) / 8 // As per note at the bottom of page 44.
+	var out []byte
+
+	for remaining > 0 {
+		counter++
+		var mac hash.Hash
```
awly: Create it outside the loop and use `mac.Reset()` on each iteration

tpm2/kdf.go
```diff
@@ -0,0 +1,57 @@
...
+		d.WriteString(label)
+		d.WriteByte(0) // Terminating null chacter for C-string.
+		d.Write(contextU)
+		d.Write(contextV)
+		if err := binary.Write(&d, binary.BigEndian, uint32(bits)); err != nil {
+			return nil, fmt.Errorf("pack bits: %v", err)
+		}
+
+		mac.Write(d.Bytes())
+		out = append(out, mac.Sum(nil)...)
```
awly: `out = mac.Sum(out)`

tpm2/kdf.go
```diff
@@ -0,0 +1,57 @@
...
+	"encoding/binary"
+	"fmt"
+	"hash"
+)
+
+// KDFa implements TPM 2.0's default key derivation function, as defined in
+// section 11.4.9.2 of the TPM revision 2 specification part 1.
+// The key & label parameters must not be zero length, but contextU &
+// contextV may be.
+func KDFa(hashAlg Algorithm, key []byte, label string, contextU, contextV []byte, bits int) ([]byte, error) {
```
awly: Are contextU and contextV expected to be 0-terminated strings? Should we verify that? Or should they be strings instead of []byte?

tpm2/kdf_test.go
```diff
@@ -0,0 +1,54 @@
...
+			label:    "IDENTITY",
+			bits:     256,
+			expected: []byte{0x83, 0xf3, 0x54, 0xaf, 0xcf, 0x92, 0x3d, 0xe2, 0x11, 0x2e, 0x08, 0x91, 0x43, 0x4c, 0xd0, 0xbd, 0xc8, 0xac, 0xbf, 0x01, 0xb8, 0x11, 0xc0, 0xe8, 0xcd, 0x06, 0x2d, 0xed, 0x39, 0xe3, 0x1f, 0x7f},
+		},
+	}
+
+	for i, tc := range tcs {
+		o, err := KDFa(tc.hashAlg, tc.key, tc.label, tc.contextU, tc.contextV, tc.bits)
+		if err != nil {
+			t.Fatal(err)
```
awly: Print input args in the error

tpm2/kdf_test.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		if bytes.Compare(tc.expected, o) != 0 {
+			t.Errorf("Test %d got %v, want %v", i, o, tc.expected)
```
awly: Print input args instead of test index
Output got/want values on separate lines so they line up on top of each other and can be scanned easily

awly:
Regarding squashing - don't worry about it, we can squash at merge time via github UI.

tpm2/kdf.go
```diff
@@ -0,0 +1,57 @@
...
+	"crypto/hmac"
+	"crypto/sha1"
+	"crypto/sha256"
+	"encoding/binary"
+	"fmt"
+	"hash"
+)
+
+// KDFa implements TPM 2.0's default key derivation function, as defined in
+// section 11.4.9.2 of the TPM revision 2 specification part 1.
```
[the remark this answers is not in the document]
twitchy-jsonp: I'm hesitant to do this; its a PDF so I can only link to the whole thing rather than the specific section, and I've already had the URLs break once since ive been working with them. Also, given the number of places we reference the spec, this could get heavy real fast. Are you sure?

tpm2/kdf.go
```diff
@@ -0,0 +1,57 @@
...
+// contextV may be.
+func KDFa(hashAlg Algorithm, key []byte, label string, contextU, contextV []byte, bits int) ([]byte, error) {
+	var counter uint32
+	remaining := (bits + 7) / 8 // As per note at the bottom of page 44.
+	var out []byte
+
+	for remaining > 0 {
+		counter++
+		var mac hash.Hash
+		switch hashAlg {
```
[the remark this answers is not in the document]
twitchy-jsonp: I'm following the pattern of only implementing whats needed / testable, as observed elsewhere in the package.

KDFa() as specified can support also SHA386 & SHA512, but I dont have any test data or real-data validation that this works for those.

tpm2/kdf.go
```diff
@@ -0,0 +1,57 @@
...
+		case AlgSHA1:
+			mac = hmac.New(sha1.New, key)
+		case AlgSHA256:
+			mac = hmac.New(sha256.New, key)
+		default:
+			return nil, fmt.Errorf("hash algorithm 0x%x is not supported", hashAlg)
+		}
+
+		var d bytes.Buffer
```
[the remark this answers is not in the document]
twitchy-jsonp: As opposed to building up a buffer?

I'm not sure at all, I would hazard a guess not. I think hash(b) + hash(a) is a different op to hash(b + a).

tpm2/kdf.go
twitchy-jsonp: contextU/contextV are a bunch of different things depending on the command/context its being used for (we cant enforce anything here). The only one that is guaranteed to be a 0-terminated string is label, and we handle that already by accepting the string type.

tpm2/kdf.go
twitchy-jsonp: Done.

tpm2/kdf.go
twitchy-jsonp: Done.

tpm2/kdf_test.go
twitchy-jsonp: Done.

tpm2/kdf_test.go
twitchy-jsonp: Doneskies

twitchy-jsonp:
Okies, I think this one is ready to go. PTAL! 🦈

twitchy-jsonp:
Third times the charm xD

PR #41, opened 2018-10-25
Implement encode/decode of CreationData (TPMS_CREATION_DATA)

twitchy-jsonp:
In addition, support decode of TPML_PCR_SELECTION structure of length 0.

DecodeCreationData has been tested on real hardware, compliant with Microsoft requirements for Windows 10.

### DIFF
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -545,31 +545,31 @@ func (cd *CreationData) encode() ([]byte, error) {
 
 // DecodeCreationData decodes a TPMS_CREATION_DATA message. No error is
 // returned if the input has extra trailing data.
-func DecodeCreationData(in []byte) (*CreationData, error) {
-	buf := bytes.NewBuffer(in)
+func DecodeCreationData(buf []byte) (*CreationData, error) {
+	in := bytes.NewBuffer(buf)
 	var out CreationData
 
-	sel, err := decodeTPMLPCRSelection(buf)
+	sel, err := decodeTPMLPCRSelection(in)
 	if err != nil {
 		return nil, fmt.Errorf("decoding PCRSelection: %v", err)
 	}
 	out.PCRSelection = sel
 
-	if err := tpmutil.UnpackBuf(buf, &out.PCRDigest, &out.Locality, &out.ParentNameAlg); err != nil {
+	if err := tpmutil.UnpackBuf(in, &out.PCRDigest, &out.Locality, &out.ParentNameAlg); err != nil {
 		return nil, fmt.Errorf("decoding PCRDigest, Locality, ParentNameAlg: %v", err)
 	}
 
-	n, err := decodeName(buf)
+	n, err := decodeName(in)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


twitchy-jsonp:
The Decode function is public so external (not in this package) code fetching `PCP_TPM12_IDBINDING` through Windows PCP APIs can decode it.

tpm2/tpm2.go
```diff
@@ -87,7 +87,12 @@ func decodeTPMLPCRSelection(buf *bytes.Buffer) (PCRSelection, error) {
 	if err := tpmutil.UnpackBuf(buf, &count); err != nil {
 		return sel, err
 	}
-	if count != 1 {
+	switch count {
+	case 0:
+		sel.Hash = AlgUnknown
+		return sel, nil
+	case 1:
```
tmroeder: nit: For clarity, please annotate the `case 1` statement with a comment to say why it's a no-op and `case 0` isn't.

tpm2/structures.go
```diff
@@ -507,6 +507,75 @@ func (ci CertifyInfo) encode() ([]byte, error) {
...
+	o, err := tpmutil.Pack(cd.OutsideInfo)
+	if err != nil {
+		return nil, fmt.Errorf("encoding OutsideInfo: %v", err)
+	}
+	return concat(sel, d, pn, pqn, o)
+}
+
+// DecodeCreationData decodes a TPMS_CREATION_DATA message. No error is
+// returned if the input has extra trailing data.
+func DecodeCreationData(in []byte) (*CreationData, error) {
```
awly: Rename `in` to `buf` to keep consistent with the other PR

twitchy-jsonp:
> How is this struct used?

A server can use this to extract key information about an AIK its considering using, and decide whether its sufficiently protected. When used in combination with `CREATION_INFO` from the attestation structure, this can be shown to have high integrity (thanks to the attestation signature).

twitchy-jsonp:
In my case, windows PCP APIs are throwing me these bytes representing `TPM2B_CREATION_DATA`, and I use this on the server to unpack & verify bits.

awly:
I see, so this is the `creationData` result field for `TPM2_Create` and `TPM2_CreatePrimary`.
Can you add decoding of this field to `Create` and `CreatePrimary` (https://github.com/google/go-tpm/blob/8a9242c41e6f96ade01381ba89a84336d6617f83/tpm2/tpm2.go#L438-L449 and https://github.com/google/go-tpm/blob/8a9242c41e6f96ade01381ba89a84336d6617f83/tpm2/tpm2.go#L339-L378).
We shouldn't actually return it to preserve the func signatures, but would make it clearer in the future where this data comes from.

awly:
Yup! Do all tests pass against a real TPM?

twitchy-jsonp:
I dont have a linux TPM 2.0 device to test the full suite, but it works fine with blobs from Windows PCP.

awly:
OK, verified in a linux VM. All tests still pass.

PR #42, opened 2018-10-26
Add PCR Extend support for TPM2.0

lsiudut:
Support for TPM1.2 landed quite some time ago but it was still
missing for TMP2.0. This commit fills the gap.

Format of the command is quite different from TPM1.2. Few remars
regarding this simplified implementation:

  - it requires session
  - because of above I'm going with authentication with empty
    password
  - support for extending a single PCR only

To keep the code consistent I followed coding style used in
PCREvent function.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -1115,6 +1115,33 @@ func concat(chunks ...[]byte) ([]byte, error) {
 	return bytes.Join(chunks, nil), nil
 }
 
+func encodePCRExtend(pcr uint32, hashAlg Algorithm, hash tpmutil.RawBytes) ([]byte, error) {
+	ha, err := tpmutil.Pack(pcr)
+	if err != nil {
+		return nil, err
+	}
+	auth, err := encodeAuthArea(HandlePasswordSession, "")
+	if err != nil {
+		return nil, err
+	}
+	var pcrCount uint32 = 1
+	extend, err := tpmutil.Pack(pcrCount, hashAlg, hash)
+	if err != nil {
+		return nil, err
+	}
+	return concat(ha, auth, extend)
+}
+
+// PcrExtend extends a value into the selected PCR
+func PCRExtend(rw io.ReadWriter, pcr uint32, hashAlg Algorithm, hash []byte) error {
+	cmd, err := encodePCRExtend(pcr, hashAlg, hash)
+	if err != nil {
+		return err
+	}
+	_, err = runCommand(rw, TagSessions, cmdPCRExtend, tpmutil.RawBytes(cmd))
+	return err
+}
+
 // ReadPCR reads the value of the given PCR.
 func ReadPCR(rw io.ReadWriteCloser, pcr int, hashAlg Algorithm) ([]byte, error) {
 	pcrSelection := PCRSelection{
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -1115,16 +1115,16 @@ func concat(chunks ...[]byte) ([]byte, error) {
 	return bytes.Join(chunks, nil), nil
 }
 
-func encodePCRExtend(pcr uint32, hashAlg Algorithm, hash tpmutil.RawBytes) ([]byte, error) {
+func encodePCRExtend(pcr tpmutil.Handle, hashAlg Algorithm, hash tpmutil.RawBytes, password string) ([]byte, error) {
 	ha, err := tpmutil.Pack(pcr)
 	if err != nil {
 		return nil, err
 	}
-	auth, err := encodeAuthArea(HandlePasswordSession, "")
+	auth, err := encodeAuthArea(HandlePasswordSession, password)
 	if err != nil {
 		return nil, err
 	}
-	var pcrCount uint32 = 1
+	pcrCount := uint32(1)
 	extend, err := tpmutil.Pack(pcrCount, hashAlg, hash)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


lsiudut:
CLA signed

tpm2/tpm2_test.go
```diff
@@ -589,6 +590,66 @@ func TestPCREvent(t *testing.T) {
 	}
 }
 
+func TestPCRExtendSha1(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	var pcrValue = make([]byte, 20)
+	var value = "FFFFFFFFFFFFFFFFFFFF"
+	copy(pcrValue[:], value)
```
awly: `pcrValue := bytes.Repeat([]byte{0xF}, sha1.Size)`

tpm2/tpm2_test.go
```diff
@@ -589,6 +590,66 @@ func TestPCREvent(t *testing.T) {
...
+
+	oldPcrValue, err := ReadPCR(rw, 16, AlgSHA1)
```
awly: Make PCR index a const

tpm2/tpm2_test.go
```diff
@@ -589,6 +590,66 @@ func TestPCREvent(t *testing.T) {
...
+	if err != nil {
+		t.Fatal("Can't read PCR 16 from the TPM: ", err)
+	}
+
+	err = PCRExtend(rw, 16, AlgSHA1, pcrValue)
```
awly: Merge this with error handling

tpm2/tpm2_test.go
awly: `oldPCRValue`

tpm2/tpm2_test.go
```diff
@@ -589,6 +590,66 @@ func TestPCREvent(t *testing.T) {
...
+
+	newPcrValue, err := ReadPCR(rw, 16, AlgSHA1)
+	if err != nil {
+		t.Fatal("Can't read PCR 16 from the TPM: ", err)
+	}
+
+	finalPcr := sha1.Sum(append(oldPcrValue, pcrValue[:]...))
+
+	if !bytes.Equal(finalPcr[:], newPcrValue) {
+		t.Fatal("PCRs not equal!")
```
awly: Print got/want values

tpm2/tpm2_test.go
```diff
@@ -589,6 +590,66 @@ func TestPCREvent(t *testing.T) {
...
+	}
+}
+
+func TestPCRExtendSha256(t *testing.T) {
```
awly: Combine this with the other test

tpm2/tpm2.go
```diff
@@ -1115,6 +1115,33 @@ func concat(chunks ...[]byte) ([]byte, error) {
...
+	var pcrCount uint32 = 1
+	extend, err := tpmutil.Pack(pcrCount, hashAlg, hash)
+	if err != nil {
+		return nil, err
+	}
+	return concat(ha, auth, extend)
+}
+
+// PcrExtend extends a value into the selected PCR
+func PCRExtend(rw io.ReadWriter, pcr uint32, hashAlg Algorithm, hash []byte) error {
```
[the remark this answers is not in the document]
lsiudut: Thanks for looking into that.

So the reason why I used uint32 it TPM2.0 spec. It defines `TPM_HT_PCR` as `uint8`:

> PCR – consecutive numbers, starting at 0, that reference the PCR registers

Well, in context of the actuall TPM calls it's always 32-bit value but it didn't seem like `int` to me. Also it's not handle, at least to my understanding.

I struggled for a while whether to make it `tpmutil.Handle` or not, couldn't find a clear indication on the type in the spec. Now when I dug it it a bit omre it seems that it is `Handle` after all, so I'll make it such (part 2, structs, ch. 7 - Handles).

tpm2/tpm2_test.go
```diff
@@ -589,6 +590,59 @@ func TestPCREvent(t *testing.T) {
 	}
 }
 
+func TestPCRExtend(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+    const pcr = int(16)
```
awly: Looks like formatting is off. Did you run `go fmt` or `goimports` on this file?

tpm2/tpm2_test.go
```diff
@@ -589,6 +590,59 @@ func TestPCREvent(t *testing.T) {
...
+
+    // testing SHA1
```
awly: Refactor this into table-driven test with subtests. Something like:
```
tests := []struct {
  desc string
  hashAlg Algorithm
  hashSize int
  hashSum func([]byte) []byte
} {
  {
    desc: "SHA1",
    hashAlg: AlgSHA1,
    hashSize: sha1.Size,
    hashSum: func(in []byte) []byte {
      s := sha1.Sum(in)
      return s[:]
    }
  },
  {
    desc: "SHA256",
    ...
  }
}
for _, tt := range tests {
  t.Run(tt.desc, func(t *testing.T) {
    oldPCRValue, err := ...
  })
}
```

tpm2/tpm2_test.go
lsiudut: I blame huge jetlag

tpm2/tpm2_test.go
lsiudut: yea this is cool - my limited go skills prevented me from thinking about something like that (y)

tpm2/tpm2_test.go
```diff
@@ -594,52 +594,55 @@ func TestPCRExtend(t *testing.T) {
...
+				s := sha256.Sum256(in)
+				return s[:]
+			},
+		},
 	}
 
-    // testing SHA256
-    pcrValue = bytes.Repeat([]byte{0xF}, sha256.Size)
+	for _, tt := range tests {
+		const pcr = int(16)
```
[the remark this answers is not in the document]
lsiudut: pcrValue is dependant on hashSize

lsiudut:
Thanks! (y)

PR #43, opened 2018-10-31
Fix re-encoding of TPMS_RSA_PARMS structure

twitchy-jsonp:
Fix re-encoding TPMS_RSA_PARMS structure where bitstream is different if exponent was initially at the default value, breaking some TPMU_NAME computations.

This will affect all name digest computations using the RSA EK template specified by TCG (ie: all the windows 8/10 devices I have encountered IRL so far).


twitchy-jsonp:
zomg this PR includes my other one for some reason and I dont know how to fix it. Maybe we wait for the other one to be merged?

twitchy-jsonp:
Git mess has been fixed (thanks @ericchiang !)

tpm2/structures.go
```diff
@@ -119,12 +119,18 @@ func DecodePublic(buf []byte) (Public, error) {
...
-	KeyBits    uint16
-	Exponent   uint32
-	ModulusRaw []byte
-	Modulus    *big.Int
+	Symmetric *SymScheme
+	Sign      *SigScheme
+	KeyBits   uint16
+	// The default Exponent (65537) has two representations; the
+	// zero value, and the value 65537.
+	// If EncodeDefaultExponent is set, an exponent of 65537 will be
```
awly: This seems counter-intuitive. Without reading the comment, i'd assume that `EncodeDefaultExponent` means "encode the exponent if it's the default value".
How about `EncodeDefaultExponentAsZero`?

Also, see https://github.com/google/go-tpm/pull/45, which does the opposite - decoding doesn't default to `defaultRSAExponent` when it's 0.

twitchy-jsonp: I'm happy to rename to `EncodeDefaultExponentAsZero`.

I'm against keeping Exponent in the packed (possibly zero) form - its technical incorrect and potentially breaks code.

tpm2/structures.go
```diff
@@ -119,12 +119,18 @@ func DecodePublic(buf []byte) (Public, error) {
...
-	Exponent   uint32
-	ModulusRaw []byte
-	Modulus    *big.Int
+	Symmetric *SymScheme
+	Sign      *SigScheme
+	KeyBits   uint16
+	// The default Exponent (65537) has two representations; the
+	// zero value, and the value 65537.
+	// If EncodeDefaultExponentAsZero is set, an exponent of 65537
+	// will be encoded as zero. This is necessary to produce an identical
```
awly: `s/zero/0/`

tpm2/tpm2_test.go
```diff
@@ -675,3 +675,34 @@ func TestEncodeDecodeCreationAttestationData(t *testing.T) {
...
+		NameAlg:    AlgSHA1,
+		Attributes: FlagSign | FlagSensitiveDataOrigin | FlagUserWithAuth,
+		RSAParameters: &RSAParams{
+			Sign: &SigScheme{
+				Alg:  AlgRSASSA,
+				Hash: AlgSHA1,
+			},
+			KeyBits:                     2048,
+			EncodeDefaultExponentAsZero: true,
+			Exponent:                    defaultRSAExponent,
```
awly: Add a test for `EncodeDefaultExponentAsZero==false` (within the same test func).

tpm2/tpm2_test.go
```diff
@@ -675,3 +675,34 @@ func TestEncodeDecodeCreationAttestationData(t *testing.T) {
 		t.Errorf("got decoded value:\n%v\nwant:\n%v", decoded, ad)
 	}
 }
+
+func TestEncodeDecodeIdenticalRSADefaultExponent(t *testing.T) {
```
awly: TestEncodeDecodePublicDefaultRSAExponent

tpm2/structures.go

twitchy-jsonp:
> So, to clarify, encoding RSAParams{EncodeDefaultExponentAsZero: true, Exponent: 0} and then decoding it produces a different RSAParams? This feels asymmetric.

I think the issue there is a user setting `EncodeDefaultExponentAsZero` rather than the decode function. Maybe I should unexport that?

awly:
Oh, if this field is only meant to be set via decoding then yes. It should be unexported.

twitchy-jsonp:
Doneskies xD

tpm2/structures.go
```diff
@@ -119,12 +119,18 @@ func DecodePublic(buf []byte) (Public, error) {
...
-	KeyBits    uint16
-	Exponent   uint32
-	ModulusRaw []byte
-	Modulus    *big.Int
+	Symmetric *SymScheme
+	Sign      *SigScheme
+	KeyBits   uint16
+	// The default Exponent (65537) has two representations; the
+	// 0 value, and the value 65537.
+	// If EncodeDefaultExponentAsZero is set, an exponent of 65537
```
awly: match the field name EncodeDefaultExponentAsZero -> encodeDefaultExponentAsZero

tpm2/tpm2_test.go
```diff
@@ -735,6 +735,52 @@ func TestEncodeDecodeCreationAttestationData(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("DecodePublic(%v) returned error: %v", e, err)
+	}
+	if !reflect.DeepEqual(p, d) {
+		t.Errorf("RSA TPMT_PUBLIC with default exponent changed after being encoded+decoded")
+		t.Logf("\tGot:  %+v", d)
+		t.Logf("\tWant: %+v", p)
+	}
+
+	p.RSAParameters.encodeDefaultExponentAsZero = false
```
awly: Wrap this in a loop to deduplicate
```
for _, encodeAsZero := []bool{true, false} {
t.Run(fmt.Sprintf("encodeDefaultExponentAsZero = %v", encodeAsZero), func (t *testing.T) {
  p.RSAParameters.encodeDefaultExponentAsZero = encodeAsZero
  ...
})
```

twitchy-jsonp: Thats a nice trick, thanks!

tpm2/structures.go

PR #45, opened 2018-11-01
Remove setting default exponent from decode.

xyzzyz:
Decode should decode packed buffer as is, it should set any defaults.

### DIFF
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -176,9 +176,6 @@ func decodeRSAParams(in *bytes.Buffer) (*RSAParams, error) {
 	if err := tpmutil.UnpackBuf(in, &params.KeyBits, &params.Exponent, &modBytes); err != nil {
 		return nil, fmt.Errorf("decoding KeyBits, Exponent, Modulus: %v", err)
 	}
-	if params.Exponent == 0 {
-		params.Exponent = defaultRSAExponent
-	}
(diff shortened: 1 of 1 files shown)


awly:
Can you provide rationale for this change?
Also, it overlaps with https://github.com/google/go-tpm/pull/43

twitchy-jsonp:
This is a breaking change; existing code that assumes Exponent is correct & doesnt understand a value of 0 is 65537 (ie: uses which pack into a `*rsa.PublicKey` will break.

Instead, I'm an advocate for keeping track of the original encoding (whether 65537 == an encoding of 0) and making sure the bitstream (from calling `Encode()`) matches the original packed buffer. See #43

xyzzyz:
@twitchy-jsonp right, I just run tests, and it is indeed breaking change. 

> doesnt understand a value of 0 is 65537

Right, that's the point: the packing and unpacking methods shouldn't perform the interpretation of the value, but rather keep themselves to faithfully decoding and encoding exactly what have arrived on the wire. It should rather be done by the code that actually need to use the actual exponent to do actual exponentiation. Current semantics results in confusing behavior regarding encoded Names of algorithms. 

> See #43

I'm happy that I'm not the only one interested in getting this problem solved. I'll follow that other pull request, consider this one suspended.

josephlr:
I talked to @xyzzyz about this. Now that #43 is merged, this PR can be closed.

PR #46, opened 2018-11-05
ECC capability check

pkern:
I tried go-tpm on Intel PTT and got met with 0x1ca errors (value not appropriate) when the ECC functionality was checked. So I added a possibility to query the algorithm descriptions through GetCapability and used that in the unit test to check for a supported TPM. I'm not entirely happy with it as the latter should actually be a more generic function to check for an algorithm's presence but I guess it's a start. My apologies for potentially adhering too strongly to the internal coding style in this change.

Alternatively I suspect one could also just check for 0x1ca, but that seems sort of hacky. :)

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -68,6 +68,15 @@ const (
 	AlgECB       Algorithm = 0x0044
 )
 
+// AlgorithmAttributes represents a TPMA_ALGORITHM value.
+type AlgorithmAttributes uint32
+
+// AlgorithmDescription represents a TPMS_ALGORITHM_DESCRIPTION structure.
+type AlgorithmDescription struct {
+	ID         Algorithm
+	Attributes AlgorithmAttributes
+}
+
 // SessionType defines the type of session created in StartAuthSession.
 type SessionType uint8
 
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -68,15 +68,6 @@ const (
 	AlgECB       Algorithm = 0x0044
 )
 
-// AlgorithmAttributes represents a TPMA_ALGORITHM value.
-type AlgorithmAttributes uint32
-
-// AlgorithmDescription represents a TPMS_ALGORITHM_DESCRIPTION structure.
-type AlgorithmDescription struct {
-	ID         Algorithm
-	Attributes AlgorithmAttributes
-}
-
 // SessionType defines the type of session created in StartAuthSession.
 type SessionType uint8
 
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -188,62 +188,66 @@ func ReadClock(rw io.ReadWriter) (uint64, uint64, error) {
 	return decodeReadClock(resp)
 }
 
-func decodeGetCapability(in []byte) (Capability, []interface{}, bool, error) {
+func decodeGetCapability(in []byte) ([]interface{}, bool, error) {
 	var moreData byte
 	var capReported Capability
 
 	buf := bytes.NewBuffer(in)
 	if err := tpmutil.UnpackBuf(buf, &moreData, &capReported); err != nil {
-		return capReported, nil, false, err
+		return nil, false, err
 	}
 
 	switch capReported {
 	case CapabilityHandles:
 		var numHandles uint32
 		if err := tpmutil.UnpackBuf(buf, &numHandles); err != nil {
-			return capReported, nil, false, fmt.Errorf("could not unpack handle count: %v", err)
+			return nil, false, fmt.Errorf("could not unpack handle count: %v", err)
 		}
 
 		var handles []interface{}
 		for i := 0; i < int(numHandles); i++ {
 			var handle tpmutil.Handle
 			if err := tpmutil.UnpackBuf(buf, &handle); err != nil {
-				return capReported, nil, false, fmt.Errorf("could not unpack handle: %v", err)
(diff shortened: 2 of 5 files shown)


tpm2/constants.go
```diff
@@ -68,6 +68,15 @@ const (
 	AlgECB       Algorithm = 0x0044
 )
 
+// AlgorithmAttributes represents a TPMA_ALGORITHM value.
+type AlgorithmAttributes uint32
```
awly: Move these types to structures.go

tpm2/tpm2.go
```diff
@@ -188,49 +188,62 @@ func ReadClock(rw io.ReadWriter) (uint64, uint64, error) {
...
+			}
+			algs = append(algs, alg)
+		}
+		return capReported, algs, moreData > 0, nil
+	default:
+		return capReported, nil, false, fmt.Errorf("unsupported capability %v", capReported)
+	}
 }
 
 // GetCapability returns various information about the TPM state.
```
awly: Mention supported capabilities and what types are expected in returned slice.
Also describe what moreData means and how to use it.

tpm2/tpm2.go
```diff
@@ -188,49 +188,62 @@ func ReadClock(rw io.ReadWriter) (uint64, uint64, error) {
 	return decodeReadClock(resp)
 }
 
-func decodeGetCapability(in []byte) (Capability, []tpmutil.Handle, error) {
+func decodeGetCapability(in []byte) (Capability, []interface{}, bool, error) {
```
awly: the `Capability` return can be dropped, it's not used by the caller

tpm2/tpm2.go
```diff
@@ -188,49 +188,62 @@ func ReadClock(rw io.ReadWriter) (uint64, uint64, error) {
...
 }
 
 // GetCapability returns various information about the TPM state.
 //
-// Currently only CapabilityHandles is supported (list active handles).
-func GetCapability(rw io.ReadWriter, cap Capability, count uint32, property uint32) ([]tpmutil.Handle, error) {
-	resp, err := runCommand(rw, TagNoSessions, cmdGetCapability, cap, property, count)
+// Currently only CapabilityHandles (list active handles) and CapabilityAlgs
+// (list supported algorithms) are supported.
+func GetCapability(rw io.ReadWriter, capa Capability, count uint32, property uint32) (vals []interface{}, moreData bool, err error) {
```
awly: `count, property uint32`

tpm2/tpm2_test.go
```diff
@@ -285,6 +294,33 @@ func TestHash(t *testing.T) {
 	}
 }
 
+func supportsECC(rw io.ReadWriter) (bool, error) {
+	var err error
+	moreData := true
```
awly: only `moreData` should be outside the loop
`for i := uint32(0); moreData; i++ {`

tpm2/tpm2_test.go
```diff
@@ -285,6 +294,33 @@ func TestHash(t *testing.T) {
...
+		for _, desc := range descs {
+			algorithms = append(algorithms, desc.(AlgorithmDescription))
+		}
+		if !moreData {
+			break
+		}
+		i++
+	}
+	for _, alg := range algorithms {
+		if alg.ID == AlgECC {
```
awly: Check that in the above loop, without accumulating the list of all algorithms

tpm2/tpm2_test.go
```diff
@@ -125,8 +125,17 @@ func TestGetCapability(t *testing.T) {
...
-		t.Fatalf("GetCapability failed: %s", err)
+	for _, tt := range []struct {
+		capa     Capability
+		count    uint32
+		property uint32
+	}{
+		{CapabilityHandles, 1, 0x80000000},
+		{CapabilityAlgs, 1, 0},
+	} {
+		if _, _, err := GetCapability(rw, tt.capa, tt.count, tt.property); err != nil {
```
awly: Loop over returned slice and make sure the types of values are correct

tpm2/tpm2_test.go
```diff
@@ -325,6 +361,11 @@ func TestLoadExternalPublicKey(t *testing.T) {
 		run(t, rp, private)
 	})
 	t.Run("ECC", func(t *testing.T) {
+		if ok, err := supportsECC(rw); !ok && err == nil {
```
awly: Change `supportsECC` to `skipOnUnsupportedAlg(t, rw, AlgECC)`, to remove duplicated checks.

tpm2/tpm2.go
pkern: Unfortunately this is untrue. I wanted to remove it and then I ended up re-adding it because the test actually checks if the byte slice contains the right parsed capability.

awly: That check can be removed from the test

tpm2/tpm2_test.go
pkern: Much nicer, thanks.

tpm2/tpm2_test.go
```diff
@@ -294,31 +301,25 @@ func TestHash(t *testing.T) {
 	}
 }
 
-func supportsECC(rw io.ReadWriter) (bool, error) {
+func skipOnUnsupportedAlg(t *testing.T, rw io.ReadWriter, alg Algorithm) {
 	var err error
```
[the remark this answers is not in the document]
pkern: Done.

PR #48, opened 2018-11-07
Add a TPM2 benchmark for signing operations.

pkern:


tpm2/tpm2_test.go
```diff
@@ -41,7 +41,12 @@ var (
 	mssimPlatformAddr = flag.String("mssim_platform_addr", "localhost:2322", "Host and port of the simulator's platform listener")
 )
 
-func openTPM(t *testing.T) io.ReadWriteCloser {
+type T interface {
```
awly: use `testing.TB` instead

tpm2/structures.go
```diff
@@ -210,6 +210,20 @@ type ECPoint struct {
 	X, Y *big.Int
 }
 
+func (p *ECPoint) x() *big.Int {
+	if p == nil || p.Y == nil {
```
awly: `p.X`

tpm2/benchmark_test.go
```diff
@@ -0,0 +1,78 @@
...
+		Type:       AlgRSA,
+		NameAlg:    AlgSHA256,
+		Attributes: FlagSign | FlagSensitiveDataOrigin | FlagUserWithAuth,
+		RSAParameters: &RSAParams{
+			Sign: &SigScheme{
+				Alg:  AlgRSASSA,
+				Hash: AlgSHA256,
+			},
+			KeyBits: uint16(2048),
+			Modulus: big.NewInt(0),
```
awly: I like what you did for ECPoint coords. Perhaps add the same helper to RSAParams.Modulus?

tpm2/benchmark_test.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		b.Fatalf("CreatePrimary failed: %v", err)
+	}
+	defer FlushContext(rw, signerHandle)
+
+	digest := sha256.Sum256([]byte("randomString"))
+
+	b.StartTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := Sign(rw, signerHandle, defaultPassword, digest[:], pub.RSAParameters.Sign)
```
awly: combine this with error handling

tpm2/tpm2_test.go
pkern: I knew there had to be a type for this but didn't find it (also because TB does not have its functions listed in godoc's TOC for testing).

tpm2/benchmark_test.go
pkern: The reason I did it here was that it was nested one layer down (through Point, which can also be nil). Modulus is accessed directly by RSAParams' encode() function. We could default the modulus to 0 there if we wanted, but right now there's explicit code forbidding that input (also because you can pass in the raw modulus).

pkern:
Sure. On Intel PTT (NUC D34010WYK):

```
$ go test -bench=. --tpm_path=/dev/tpm0 -run=foo -v
goos: linux
goarch: amd64
BenchmarkRSA2048Signing-4       	       1	1249447747 ns/op
--- SKIP: BenchmarkECCNISTP256Signing
	tpm2_test.go:322: Algorithm 35 is not supported by the TPM
PASS
ok  	_/home/pkern/src/go-tpm/tpm2	5.912s
```

The results are fairly consistent. I'll also provide one other output from a different machine.

tpm2/benchmark_test.go
pkern: Done.

tpm2/structures.go
pkern: Good catch, done.

pkern:
On a Dell Latitude 5520 dTPM:

```
% go test --bench=. --run=foo --tpm_path=/dev/tpm0 -v
goos: linux
goarch: amd64
BenchmarkRSA2048Signing-8       	       2	 565028667 ns/op
BenchmarkECCNISTP256Signing-8   	      10	 201683083 ns/op
PASS
ok  	_/home/pkern/src/go-tpm/tpm2	28.078s
```

tpm2/benchmark_test.go
awly: Ah, good point. nil Modulus actually has meaning when raw modulus is used

PR #52, opened 2018-11-30
Implement NVWrite command

folbricht:
This adds an implementation of the NVWrite command.

Would like to also add a test, but can't get NVRead to work without modifications.

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -273,39 +273,3 @@ var hashConstructors = map[Algorithm]func() hash.Hash{
 	AlgSHA384: sha512.New384,
 	AlgSHA512: sha512.New,
 }
-
-// NV attributes
-const (
-	NVAttrPPWrite = 1 << iota
-	NVAttrOwnerWrite
-	NVAttrAuthWrite
-	NVAttrPolicyWrite
-	_
-	_
-	_
-	_
-	_
-	_
-	NVAttrPolicyDelete
-	NVAttrWriteLocked
-	NVAttrWriteAll
-	NVAttrWriteDefine
-	NVAttrWriteSTClear
-	NVAttrGlobalLock
-	NVAttrPPRead
-	NVAttrOwnerRead
-	NVAttrAuthRead
-	NVAttrPolicyRead
-	_
-	_
-	_
-	_
-	_
-	NVAttrNoDA
-	NVAttrOrderly
-	NVAttrClearSTClear
-	NVAttrReadLocked
-	NVAttrWritten
-	NVAttrPlatformCreate
-	NVAttrReadSTClear
-)
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -273,3 +273,39 @@ var hashConstructors = map[Algorithm]func() hash.Hash{
 	AlgSHA384: sha512.New384,
 	AlgSHA512: sha512.New,
 }
+
+// NV attributes
+const (
+	NVAttrPPWrite = 1 << iota
+	NVAttrOwnerWrite
+	NVAttrAuthWrite
+	NVAttrPolicyWrite
+	_
+	_
+	_
+	_
+	_
+	_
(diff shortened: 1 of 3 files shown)


tpm2/constants.go
```diff
@@ -273,3 +273,39 @@ var hashConstructors = map[Algorithm]func() hash.Hash{
 	AlgSHA384: sha512.New384,
 	AlgSHA512: sha512.New,
 }
+
+// NV attributes
```
awly: Are these used for anything?

folbricht: I'm using those when creating the index. Like so:

```go
	if err := tpm2.NVDefineSpace(rwc,
		tpm2.HandleOwner,
		idx,
		emptyPassword,
		emptyPassword,
		nil,
		tpm2.NVAttrOwnerWrite|tpm2.NVAttrOwnerRead|tpm2.NVAttrAuthRead|tpm2.NVAttrPPRead,
		uint16(len(data)),
	); err != nil {
		panic(err)
	}
```

folbricht:
The issue I have with NVRead is this line. It encodes the index twice

https://github.com/google/go-tpm/blob/03d39f2bdd2583d4a85e83ffa8f35b2d672233b1/tpm2/tpm2.go#L996

Pretty sure the first value needs to be HandleOwner. The implementation below works for me. Not sure if I'm using it wrong and how the old way works for you.

```go
func encodeNVRead(owner, index tpmutil.Handle, authString string, offset, dataSize uint16) ([]byte, error) {
	ha, err := tpmutil.Pack(owner, index)
	if err != nil {
		return nil, err
	}
	auth, err := encodeAuthArea(HandlePasswordSession, authString)
	if err != nil {
		return nil, err
	}
	params, err := tpmutil.Pack(dataSize, offset)
	if err != nil {
		return nil, err
	}
	return concat(ha, auth, params)
}

// NVRead reads a full data blob from an NV index.
func NVRead(rw io.ReadWriter, owner, index tpmutil.Handle) ([]byte, error) {
	// Read public area to determine data size.
	resp, err := runCommand(rw, TagNoSessions, cmdReadPublicNV, index)
	if err != nil {
		return nil, fmt.Errorf("running NV_ReadPublic command: %v", err)
	}
	pub, err := decodeNVReadPublic(resp)
	if err != nil {
		return nil, fmt.Errorf("decoding NV_ReadPublic response: %v", err)
	}

	// Read pub.DataSize of actual data.
	cmd, err := encodeNVRead(owner, index, "", 0, pub.DataSize)
	if err != nil {
		return nil, fmt.Errorf("building NV_Read command: %v", err)
	}
	resp, err = runCommand(rw, TagSessions, cmdReadNV, tpmutil.RawBytes(cmd))
	if err != nil {
		return nil, fmt.Errorf("running NV_Read command: %v", err)
	}
	return decodeNVRead(resp)
}
```

tpm2/constants.go
awly: Then let's make a named type for this and use hex values instead of `iota` (like `KeyProp`).
And we should change NVDefineSpace `attributes` argument to use this new type.

folbricht:
Not a problem. Will add a test to this PR, and then make another one for the NV attributes and the NVRead fix.

folbricht:
@awly Added the test in https://github.com/google/go-tpm/pull/52/commits/b14dcab1777dd15a4a828b2d12d82421809e74e3 and also removed the attributes.

The next PR will update the test with named attribs and also validate the data written.

twitchyliquid64:
>  ha, err := tpmutil.Pack(handle, handle) 
> Pretty sure the first value needs to be HandleOwner.

The first parameter is encoding the handle area, which is a `TPMI_RH_NV_AUTH` structure. If you are authorizing as the Owner or Platform, you provide HandleOwner or HandlePlatform. In all other cases, you provide the NV index (at which point authorization will follow to rules for that NV public).

PR #54, opened 2018-12-05
Fix NVRead and Add NVReadPublic / NVReadAll

folbricht:
As discussed in https://github.com/google/go-tpm/pull/52 this PR:

- Corrects NVRead with owner, offset and size as per spec
- Exposes NVReadPublic
- Adds a covenience function `NVReadAll` to implement the old behavior (read all data from an index)
- Updates the NVWrite test to include NVRead for validation

Since it changes the function calls, this is a breaking change.

If you prefer hex values for the attribute bitmaps, I can change the PR.

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -273,3 +273,42 @@ var hashConstructors = map[Algorithm]func() hash.Hash{
 	AlgSHA384: sha512.New384,
 	AlgSHA512: sha512.New,
 }
+
+// NVAttr is a bitmask used in Attributes field of NV indexes. Individual
+// flags should be OR-ed to form a full mask.
+type NVAttr uint32
+
+const (
+	AttrPPWrite NVAttr = 1 << iota
+	AttrOwnerWrite
+	AttrAuthWrite
+	AttrPolicyWrite
+	_
+	_
+	_
+	_
+	_
+	_
+	AttrPolicyDelete
+	AttrWriteLocked
+	AttrWriteAll
+	AttrWriteDefine
+	AttrWriteSTClear
+	AttrGlobalLock
+	AttrPPRead
+	AttrOwnerRead
+	AttrAuthRead
+	AttrPolicyRead
+	_
+	_
+	_
+	_
+	_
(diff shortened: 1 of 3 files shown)


folbricht:
Looks like there may be a clash with #53

twitchy-jsonp:
We did this at the same time! ^^

One comment: You are changing already available APIs, which will break anyone using go-tpm using those APIs. I believe this goes against Go idioms (though this call is up to the maintainers).

In the past i have made a new method (ie: `NVReadEx`) with the extended functionality to avoid breaking existing users.

folbricht:
Yes, #54 changes the API and would be breaking existing implementations using it. On the other hand, I'm not sure the old `NVRead` was working at all so changing it shouldn't actually impact anyone.

Whatever the maintainers decide to do here works for me. If #53 gets merged first, I can update this PR to align with it.

twitchyliquid64:
The old `NVRead` worked fine if the read was less than the TPM's read buffer size (On ST Microelectronics TPMs, this is 1k).

awly:
Thanks for the fixes! Sorry for the slow responses, busy week.
I'll take a look at these tomorrow or Friday.

awly:
The old `NVRead` worked for some use-cases, I wrote code that depends on it.
I agree with @twitchy-jsonp hereon not breaking the API.

Let's keep the `NVAttr` and `NVDefineSpace` part of this PR however. `NVAttr` is uint32 so it would only break the users in very few cases. If they use consts as `NVDefineSpace` arguments, it should "just work".

@tmroeder maybe at some point we make go-tpm a Go module. Then we can make a breaking changes and clean up legacy APIs.

awly:
And yes, let's use hex values for `NVAttr`

twitchy-jsonp:
See my PR #53 for an implementation of NVRead which doesnt break the API. It seems like we are going with this PR, so feel free to copy-pasta in anything that makes sense.

folbricht:
Superceded by https://github.com/google/go-tpm/pull/63 based on comments above.

PR #55, opened 2018-12-06
Add public command code for PolicySecret

josephlr:
Our example currently includes the default AuthPolicy value for the EK
template as a raw byte slice. This is a little opaque. We plan on including
a method to calculate this value in go-tpm-tools. However, this requires
the command code for PolicySecret to be public.

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -234,28 +234,31 @@ const (
 	cmdCertify            tpmutil.Command = 0x00000148
 	cmdCertifyCreation    tpmutil.Command = 0x0000014A
 	cmdReadNV             tpmutil.Command = 0x0000014E
-	cmdCreate             tpmutil.Command = 0x00000153
-	cmdLoad               tpmutil.Command = 0x00000157
-	cmdQuote              tpmutil.Command = 0x00000158
-	cmdSign               tpmutil.Command = 0x0000015D
-	cmdUnseal             tpmutil.Command = 0x0000015E
-	cmdContextLoad        tpmutil.Command = 0x00000161
-	cmdContextSave        tpmutil.Command = 0x00000162
-	cmdFlushContext       tpmutil.Command = 0x00000165
-	cmdLoadExternal       tpmutil.Command = 0x00000167
-	cmdMakeCredential     tpmutil.Command = 0x00000168
-	cmdReadPublicNV       tpmutil.Command = 0x00000169
-	cmdReadPublic         tpmutil.Command = 0x00000173
-	cmdStartAuthSession   tpmutil.Command = 0x00000176
-	cmdGetCapability      tpmutil.Command = 0x0000017A
-	cmdGetRandom          tpmutil.Command = 0x0000017B
-	cmdHash               tpmutil.Command = 0x0000017D
-	cmdPCRRead            tpmutil.Command = 0x0000017E
-	cmdPolicyPCR          tpmutil.Command = 0x0000017F
-	cmdReadClock          tpmutil.Command = 0x00000181
-	cmdPCRExtend          tpmutil.Command = 0x00000182
-	cmdPolicyGetDigest    tpmutil.Command = 0x00000189
-	cmdPolicyPassword     tpmutil.Command = 0x0000018C
+	// CmdPolicySecret is a command code for TPM2_PolicySecret.
+	// It's exported for computing of default AuthPolicy value.
+	CmdPolicySecret     tpmutil.Command = 0x00000151
+	cmdCreate           tpmutil.Command = 0x00000153
+	cmdLoad             tpmutil.Command = 0x00000157
+	cmdQuote            tpmutil.Command = 0x00000158
+	cmdSign             tpmutil.Command = 0x0000015D
+	cmdUnseal           tpmutil.Command = 0x0000015E
+	cmdContextLoad      tpmutil.Command = 0x00000161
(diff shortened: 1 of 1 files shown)


examples/tpm2-ekcert/main.go
```diff
@@ -51,6 +47,22 @@ var (
 	}
 )
 
+// Auth Policy calculations in Credential_Profile_EK_V2.0, section 2.1.5.3, "authPolicy"
+func defaultAuthPolicy() []byte {
+	previousDigest := make([]byte, 32)            // All zero buffer
+	policySecretCommandCode := uint32(0x00000151) // TPM__CC_PolicySecret
```
awly: Add this to `tpm2/constants.go`

examples/tpm2-ekcert/main.go
```diff
@@ -51,6 +47,22 @@ var (
...
+	name := uint32(0x4000000B)                    // TPM_RH_ENDORSEMENT
```
awly: Also this to `constants.go`

examples/tpm2-ekcert/main.go
```diff
@@ -51,6 +47,22 @@ var (
...
+	policyRef := make([]byte, 0)
+
+	buf := bytes.NewBuffer(previousDigest)
+	binary.Write(buf, binary.BigEndian, policySecretCommandCode)
+	binary.Write(buf, binary.BigEndian, name)
+	digest1 := sha256.Sum256(buf.Bytes())
+
+	digest2 := sha256.Sum256(append(digest1[:], policyRef...))
```
awly: policyRef is empty. This is identical to `digest2 := sha256.Sum256(digest1[:])`

examples/tpm2-ekcert/main.go
```diff
@@ -51,6 +47,22 @@ var (
...
+// Auth Policy calculations in Credential_Profile_EK_V2.0, section 2.1.5.3, "authPolicy"
+func defaultAuthPolicy() []byte {
+	previousDigest := make([]byte, 32)            // All zero buffer
+	policySecretCommandCode := uint32(0x00000151) // TPM__CC_PolicySecret
+	name := uint32(0x4000000B)                    // TPM_RH_ENDORSEMENT
+	policyRef := make([]byte, 0)
+
+	buf := bytes.NewBuffer(previousDigest)
+	binary.Write(buf, binary.BigEndian, policySecretCommandCode)
+	binary.Write(buf, binary.BigEndian, name)
```
awly: The above 3 lines are essentially `tpmutil.Pack(make([]byte, 32), policySecretCommandCode, name)`

examples/tpm2-ekcert/main.go
josephlr: Fixed, in fact this already exists as `tpm2.HandleEndorsement`

examples/tpm2-ekcert/main.go
josephlr: Fixed, added as tpm2.CmdPolicySecret

examples/tpm2-ekcert/main.go
josephlr: Using `tpmutil.Pack` with a slice automatically prepends the length, so we actually need to Pack the two uint32s then prepend the byte slice.

examples/tpm2-ekcert/main.go
josephlr: Fixed, this was intended to show how a non-trivial auth policy would be computed. However, a comment is better suited for this purpose.

tpm2/constants.go
```diff
@@ -223,6 +223,7 @@ const (
 	cmdCertify            tpmutil.Command = 0x00000148
 	cmdCertifyCreation    tpmutil.Command = 0x0000014A
 	cmdReadNV             tpmutil.Command = 0x0000014E
+	CmdPolicySecret       tpmutil.Command = 0x00000151
```
awly: This looks a bit awkward as the only exported cmd const. But I can't think of a much better way.
Could you add a comment explaining what this is and why it's exported?

josephlr: Added, note that this comment changes how gofmt formats the code, so many more lines are included in the diff than strcitly necessary

josephlr:
@awly I think it's better if this sort of calculation goes in go-tpm-tools. I've removed most of the code from this CL, the only remaining bit is adding `CmdPolicySecret` (and a comment about why it is public).

tpm2/constants.go
```diff
@@ -223,28 +223,30 @@ const (
...
-	cmdGetCapability      tpmutil.Command = 0x0000017A
-	cmdGetRandom          tpmutil.Command = 0x0000017B
-	cmdHash               tpmutil.Command = 0x0000017D
-	cmdPCRRead            tpmutil.Command = 0x0000017E
-	cmdPolicyPCR          tpmutil.Command = 0x0000017F
-	cmdReadClock          tpmutil.Command = 0x00000181
-	cmdPCRExtend          tpmutil.Command = 0x00000182
-	cmdPolicyGetDigest    tpmutil.Command = 0x00000189
-	cmdPolicyPassword     tpmutil.Command = 0x0000018C
+	// CmdPolicySecret must be public to compute the default AuthPolicy value.
```
awly: This appears in public godoc. Let's change the comment to:
`CmdPolicySecret is a command code for TPM2_PolicySecret. It's exported for computing of default AuthPolicy value.`

josephlr: Done

PR #56, opened 2018-12-10
Refactor encodeAuthArea to reduce chronic pain

twitchy-jsonp:
`encodeAuthArea`, its issues, and TCG handle authorization in general have been the source of ~3 days of frustration. This PR fixes this.

Issues I'm trying to address:

 * `encodeAuthArea` actually only encodes password auths or one auth
 * `encodeAuthArea` only lets you specify a single session (auth) handle, when parameters can use different ones
 * `encodeAuthArea` input parameters diverge from how auth is represented in TCG, leading to a lot of incorrect assumptions as a newcomer & a lot of confusion.

This PR:

 * Creates `AuthCommand` structure, which 1:1 represents how auth is encoded for TCG purposes.
 * Refactors existing use of `encodeAuthArea` to pass an `AuthCommand` structure.

Having each `encode****` method pass `AuthCommand` structures to encodeAuthArea means auth encoding is immediately obvious, next to all the code encoding the rest of the command.

### DIFF
--- a/tpm2/encoding_test.go
+++ b/tpm2/encoding_test.go
@@ -303,7 +303,7 @@ func TestEncodeEvictControl(t *testing.T) {
 }
 
 func TestEncodeAuthArea(t *testing.T) {
-	pwAuth, err := encodeAuthArea(HandlePasswordSession, defaultPassword)
+	pwAuth, err := encodeAuthArea(AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession, Auth: []byte(defaultPassword)})
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


tpm2/encoding_test.go
```diff
@@ -312,7 +312,7 @@ func TestEncodeAuthArea(t *testing.T) {
 		t.Fatalf("got: %v, want: %v", pwAuth, want)
 	}
 
-	pwAuth, err = encodeAuthArea(HandlePasswordSession, "")
+	pwAuth, err = encodeAuthArea(AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession, Auth: []byte("")})
```
awly: `[]byte("")` is `EmptyAuth` in the spec, right? Should we make a var for it in constants.go?

PR #64, opened 2019-01-15
Initialize go.mod file

### LINKED ISSUE #58: Make go-tpm a Go module
Now that go-tpm gets more dependents, it would be good to start versioning to enable them to vendor reliably.
As a bonus, we can start cleaning up the API from legacy cruft by bumping the major/minor versions.

awly:
This makes go-tpm a Go module.
After this PR is submitted, I'll tag v0.0.1 on master.
After that, we can start submitting incompatible changes and tagging
them as v0.(N+1).0.

Fixes #58

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,1 @@
+module github.com/google/go-tpm


awly:
cc @folbricht who has a PR depending on this.

@tmroeder please review and sanity-check the plan in the PR description

awly:
The file is generated. If we had non-stdlib dependencies, hey would end up there too.
Thanks for the quick review!

PR #66, opened 2019-01-23
Add Handle Type constants and some max capabilities constants.

samdamana:
Added all constants for the Handle Type (TPM_HT, Section 7.2 of [part 2 of the spec](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf)).

Added some constants yielding MaxCapabilityHandles (Section 5.5 of [part 4 of the spec](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-4-Supporting-Routines-01.38-code.pdf)).

Needed for WIP tooling around handles within google/go-tpm-tools.

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -80,6 +80,22 @@ const (
 	AlgECB       Algorithm = 0x0044
 )
 
+// HandleType defines a type of handle.
+type HandleType uint8
+
+// Supported handle types
+const (
+	HandleTypePCR           HandleType = 0x00
+	HandleTypeNVIndex       HandleType = 0x01
+	HandleTypeHMACSession   HandleType = 0x02
+	HandleTypeLoadedSession HandleType = 0x02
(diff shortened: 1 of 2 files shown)


tpm2/constants.go
```diff
@@ -171,6 +187,13 @@ const (
 	CapabilityAuthPolicies
 )
 
+// Maximum valous for capabilities
+const (
+	MaxCapabilityBufferLength uint32 = 1024
```
josephlr: It looks like this "constant" isn't actually constant (which is my mistake). We don't need these constants after all, as the user can just pass `math.MaxUint32` for the number of capabilities (and `moreData` will then always be false as `MAX_CAP_BUFFER` has to be small enough to fit inside a normal TPM2 response).

samdamana: Removed these consts.

samdamana:
Ready to be merged.

PR #69, opened 2019-01-30
Make flags consistent

josephlr:
In go-tpm, some of our flags use dashes (in `examples`) and some use
underscores (in `tpm2` testing code). This change makes it so all flags
will use dashes. This ensures that all example binaries still take the
same arguments.

The windows flag name is also changed to make it clear what it is actually doing.


awly:
Thanks! Could you also update https://github.com/google/go-tpm/blob/2e72b2e894315db5f8b08e182733a71ae236ad2c/examples/tpm-sign/verify.go#L27 (it's using camel case)

josephlr:
@awly I fixed the `publicKey` and `mssim_*` flags.

Issue #59, opened 2018-12-22
Retry command on TPM_RC_RETRY


josephlr:
Certain TPM2 commands are allowed to return TPM_RC_RETRY when the client should attempt the command again. This has caused some problems in higher level applications (see https://github.com/google/go-tpm-tools/pull/1) and in other TPM libraries (see https://github.com/tpm2-software/tpm2-tools/issues/469).

It would be nice if runCommand simply checked for TPM_RC_RETRY and reissued the command automatically, this would also help avoid having to add code higher up the stack like [this C library had to do](https://github.com/tpm2-software/tpm2-tools/pull/585/files#diff-5c58d23d240af6c75eebbdea9fef670b).

PR #71, opened 2019-02-11
Decode errors from TPM2 responses

folbricht:
This is a first attempt at decoding TPM response codes and turning them into error objects that are easier to read and deal with.

Any feedback on naming, types, structure is welcome.

### DIFF
--- a/tpm2/error.go
+++ b/tpm2/error.go
@@ -7,10 +7,17 @@ import (
 )
 
 type (
-	RCFmt0  uint8 // Format 0 error codes
-	RCFmt1  uint8 // Format 1 error codes
-	RCWarn  uint8 // Warning codes
-	RcIndex uint8 // Indexes for arguments, handles and sessions in errors
+	// RCFmt0 holds Format 0 error codes
+	RCFmt0 uint8
+
+	// RCFmt1 holds Format 1 error codes
+	RCFmt1 uint8
+
+	// RCWarn holds error codes used in warnings
+	RCWarn uint8
+
+	// RCIndex is used to reference arguments, handles and sessions in errors
+	RCIndex uint8
 )
 
 // Format 0 error codes.
@@ -226,7 +233,7 @@ var warnMsg = map[RCWarn]string{
 
 // Indexes for arguments, handles and sessions.
 const (
-	RC1 RcIndex = 0x01
+	RC1 RCIndex = 0x01
 	RC2         = 0x02
 	RC3         = 0x03
 	RC4         = 0x04
@@ -285,7 +292,7 @@ func (w Warning) Error() string {
 // ParameterError describes an error related to a parameter, and the parameter number.
 type ParameterError struct {
 	Code      RCFmt1
-	Parameter RcIndex
+	Parameter RCIndex
 }
 
 func (e ParameterError) Error() string {
@@ -299,7 +306,7 @@ func (e ParameterError) Error() string {
 // HandleError describes an error related to a handle, and the handle number.
 type HandleError struct {
 	Code   RCFmt1
-	Handle RcIndex
+	Handle RCIndex
 }
 
 func (e HandleError) Error() string {
@@ -313,7 +320,7 @@ func (e HandleError) Error() string {
 // SessionError describes an error related to a session, and the session number.
 type SessionError struct {
 	Code    RCFmt1
-	Session RcIndex
+	Session RCIndex
 }
 
 func (e SessionError) Error() string {
@@ -345,11 +352,11 @@ func decodeResponse(code tpmutil.ResponseCode) error {
 		return Error{RCFmt0(code & 0x7f)}
 	}
 	if code&0x40 > 0 { // Bit 6 set, code in 0:5, parameter number in 8:11
-		return ParameterError{RCFmt1(code & 0x3f), RcIndex((code & 0xf00) >> 8)}
+		return ParameterError{RCFmt1(code & 0x3f), RCIndex((code & 0xf00) >> 8)}
 	}
 	if code&0x800 == 0 { // Bit 11 unset, code in 0:5, handle in 8:10
-		return HandleError{RCFmt1(code & 0x3f), RcIndex((code & 0x700) >> 8)}
+		return HandleError{RCFmt1(code & 0x3f), RCIndex((code & 0x700) >> 8)}
 	}
 	// Code in 0:5, Session in 8:10
-	return SessionError{RCFmt1(code & 0x3f), RcIndex((code & 0x700) >> 8)}
+	return SessionError{RCFmt1(code & 0x3f), RCIndex((code & 0x700) >> 8)}
 }
--- a/tpm2/error.go
+++ b/tpm2/error.go
@@ -7,250 +7,260 @@ import (
 )
 
 type (
-	RcFmt0  uint8 // Format 0 error codes
-	RcFmt1  uint8 // Format 1 error codes
-	RcWarn  uint8 // Warning codes
+	RCFmt0  uint8 // Format 0 error codes
+	RCFmt1  uint8 // Format 1 error codes
+	RCWarn  uint8 // Warning codes
 	RcIndex uint8 // Indexes for arguments, handles and sessions in errors
 )
 
 // Format 0 error codes.
 const (
-	RcInitialize      RcFmt0 = 0x00
-	RcFailure                = 0x01
-	RcSequence               = 0x03
-	RcPrivate                = 0x0B
-	RcHMAC                   = 0x19
-	RcDisabled               = 0x20
-	RcExclusive              = 0x21
-	RcAuthType               = 0x24
-	RcAuthMissing            = 0x25
-	RcPolicy                 = 0x26
-	RcPCR                    = 0x27
-	RcPCRChanged             = 0x28
-	RcUpgrade                = 0x2D
-	RcTooManyContexts        = 0x2E
-	RcAuthUnavailable        = 0x2F
-	RcReboot                 = 0x30
-	RcUnbalanced             = 0x31
-	RcCommandSize            = 0x42
-	RcCommandCode            = 0x43
-	RcAuthSize               = 0x44
-	RcAuthContext            = 0x45
-	RcNVRange                = 0x46
-	RcNVSize                 = 0x47
-	RcNVLocked               = 0x48
-	RcNVAuthorization        = 0x49
-	RcNVUninitialized        = 0x4A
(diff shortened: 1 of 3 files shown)


samdamana:
I love this idea. Look forward to seeing it submitted. :)

tpm2/error.go
```diff
@@ -0,0 +1,325 @@
...
+	RcAuthType:        "authorization handle is not correct for command",
+	RcAuthMissing:     "5 command requires an authorization session for handle and it is not present.",
+	RcPolicy:          "policy failure in math operation or an invalid authPolicy value",
+	RcPCR:             "PCR check fail",
+	RcPCRChanged:      "PCR have changed since checked",
+	RcUpgrade:         "TPM is in field upgrade mode unless called via TPM2_FieldUpgradeData(), then it is not in field upgrade mode",
+	RcTooManyContexts: "context ID counter is at maximum",
+	RcAuthUnavailable: "authValue or authPolicy is not available for selected entity",
+	RcReboot:          "a _TPM_Init and Startup(CLEAR) is required before the TPM can resume operation",
+	RcUnbalanced:      "the protection algorithms (hash and symmetric) are not reasonably balanced. The digest size of the hash must be larger than the key size of the symmetric algorithm.",
```
samdamana: NIT: Inconsistent use of periods. All the other error strings don't use full sentences with periods.

tpm2/error.go
```diff
@@ -0,0 +1,325 @@
...
+	RcNVLocked:        "NV access locked",
+	RcNVAuthorization: "NV access authorization fails in command actions",
+	RcNVUninitialized: "an NV Index is used before being initialized or the state saved by TPM2_Shutdown(STATE) could not be restored",
+	RcNVSpace:         "insufficient space for NV allocation",
+	RcNVDefined:       "NV Index or persistent object already defined",
+	RcBadContext:      "context in TPM2_ContextLoad() is not valid",
+	RcCPHash:          "cpHash value already set or not correct for use",
+	RcParent:          "handle for parent is not a valid parent",
+	RcNeedsTest:       "some function needs testing",
+	RcNoResult:        "returned when an internal function cannot process a request due to an unspecified problem. This code is usually related to invalid parameters that are not properly filtered by the input unmarshaling code",
```
samdamana: NIT: This has multiple sentences like where my last comment is, but ends with no period as it probably should?

folbricht: Normally I'd agree, but these are straight out of the TPM2 spec, Part 2, Table 16. I've seen other implementation who also use the exact strings for error messages. I can change these for consistent punctuation if you think it's better that way.

folbricht:
One thing I was wondering about is do we want to keep the complete, un-parsed response code in the new error types? Right now the the Code element only contains the applicable error code for each type, so 0x22 instead of the original 0x922 for example. Do we want to store both?

tpm2/error.go
```diff
@@ -0,0 +1,325 @@
+package tpm2
+
+import (
+	"fmt"
+
+	"github.com/google/go-tpm/tpmutil"
+)
+
+type (
+	RcFmt0  uint8 // Format 0 error codes
```
awly: `s/Rc/RC/g` https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tpm2/error.go
awly: don't use periods in error messages, in case they get wrapped with fmt.Errorf https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tpm2/error.go
```diff
@@ -0,0 +1,325 @@
...
+	Rc9         = 0x09
+	RcA         = 0x0A
+	RcB         = 0x0B
+	RcC         = 0x0C
+	RcD         = 0x0D
+	RcE         = 0x0E
+	RcF         = 0x0F
+)
+
+type Error struct {
```
awly: Add doc comment https://github.com/golang/go/wiki/CodeReviewComments#initialisms
Here and for all other exported types.

You can leave `Error` methods without a comment

tpm2/error.go
```diff
@@ -0,0 +1,325 @@
...
+	Code RcFmt0
+}
+
+func (e Error) Error() string {
+	return fmt.Sprintf("error code 0x%x : %s", e.Code, fmt0Msg[e.Code])
```
awly: handle the case where `e.Code` is not in the message map

tpm2/error.go
```diff
@@ -0,0 +1,325 @@
...
+	RcReferenceS4           = 0x1C
+	RcReferenceS5           = 0x1D
+	RcReferenceS6           = 0x1E
+	RcNVRate                = 0x20
+	RcLockout               = 0x21
+	RcRetry                 = 0x22
+	RcNVUnavailable         = 0x23
+)
+
+var warnMgs = map[RcWarn]string{
```
awly: warnMgs -> warnMsg

tpm2/error.go
```diff
@@ -0,0 +1,325 @@
...
+		if code&0x800 > 0 { // Bit 11 set, warning with code in bit 0:6
+			return Warning{RcWarn(code & 0x7f)}
+		}
+		// error with code in bit 0:6
+		return Error{RcFmt0(code & 0x7f)}
+	}
+	if code&0x40 > 0 { // Bit 6 set, parameter number in 8:11, code in 0:5
+		return ParameterError{RcFmt1(code & 0x3f), RcIndex((code & 0xf00) >> 8)}
+	}
+	if code&0x800 == 0 { // Bit 11 unset, handle in 8:10, code in 0:5
```
awly: "handle in 8:11" ?

tpm2/error.go
```diff
@@ -0,0 +1,325 @@
...
+		if code&0x400 > 0 { // Bit 10 set, vendor specific code
+			return VendorError{uint32(code)}
+		}
+		if code&0x800 > 0 { // Bit 11 set, warning with code in bit 0:6
+			return Warning{RcWarn(code & 0x7f)}
+		}
+		// error with code in bit 0:6
+		return Error{RcFmt0(code & 0x7f)}
+	}
+	if code&0x40 > 0 { // Bit 6 set, parameter number in 8:11, code in 0:5
```
awly: Swap the order in comment to match order of fields in struct literal below
"// Bit 6 set, code in 0:5, parameter number in 8:11".

same for all errors below

tpm2/error_test.go
```diff
@@ -0,0 +1,79 @@
+package tpm2
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestError(t *testing.T) {
+	var err error
```
awly: remove this and use `:=` in first error

tpm2/error_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Vendor
+	err = decodeResponse(0x501)
+	ve, ok := err.(VendorError)
+	if !ok {
```
awly: Use reflect.DeepEqual here and for all other comparisons

tpm2/error.go
folbricht: Is the question about the number of bits for the handle? The parameter# really is 4 bits, but handle# and session# in the other error types are only 3 bits as per TPM spec, part 1, figure 27

tpm2/error.go
folbricht: done

tpm2/error.go
folbricht: done

tpm2/error.go
folbricht: done

tpm2/error.go
folbricht: done

tpm2/error.go
folbricht: done

tpm2/error_test.go
folbricht: Great idea. Went a little further with that to reduce duplication. Thanks

folbricht:
Thanks for the review. Apart from debugging, my main motivation is actually handling things like RC_RETRY (as described in https://github.com/google/go-tpm/issues/59) and other error cases more gracefully.

tpm2/error_test.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"reflect"
+	"testing"
+
+	"github.com/google/go-tpm/tpmutil"
+)
+
+func TestError(t *testing.T) {
+	tests := []struct {
+		Response tpmutil.ResponseCode
```
awly: make these fields unexported

tpm2/error_test.go
```diff
@@ -0,0 +1,29 @@
...
+		Expected error
+	}{
+		{0x501, VendorError{Code: 0x501}},                           // Vendor
```
awly: Remove comments on each test case, they are redundant

tpm2/error_test.go
```diff
@@ -0,0 +1,29 @@
...
+		{0x100, Error{Code: RCInitialize}},                          // Error
+		{0xfc1, ParameterError{Code: RCAsymmetric, Parameter: RCF}}, // ParameterError
+		{0x7a3, HandleError{Code: RCExpired, Handle: RC7}},          // HandleError
+		{0xfa2, SessionError{Code: RCBadAuth, Session: RC7}},        // SessionError
+	}
+
+	for _, test := range tests {
+		err := decodeResponse(test.Response)
+		if !reflect.DeepEqual(err, test.Expected) {
+			t.Fatalf("expected error %#v for response 0x%x, got %#v", test.Expected, test.Response, err)
```
awly: Put actual error before expected one in the message.
Common format is: `t.Fatalf("decodeResponse(0x%x) = %#v, want %#v", test.response, err, test.expected)`

tpm2/error.go
```diff
@@ -0,0 +1,362 @@
...
+	RCReferenceS6:    "the 7th authorization session handle references a session that is not loaded",
+	RCNVRate:         "the TPM is rate-limiting accesses to prevent wearout of NV",
+	RCLockout:        "authorizations for objects subject to DA protection are not allowed at this time because the TPM is in DA lockout mode",
+	RCRetry:          "the TPM was not able to start the command",
+	RCNVUnavailable:  "the command may require writing of NV and NV is not current accessible",
+}
+
+// Indexes for arguments, handles and sessions.
+const (
+	RC1 RCIndex = 0x01
```
awly: Use `iota` for these consts

awly:
Thanks for fixing all the comments!
Please explicitly comment when PR is ready for another look, since github email notifications don't make it clear when all comments get addressed.

PR #72, opened 2019-02-17
Add method to get a crypto.PublicKey key from tpm2.Public

folbricht:
This change moves the code that extract the public key from tpm2.Public so it can be used to more generally. The reason is that `CreatePrimary()` already returns a crypto.PublicKey, while `ReadPublic()` does not.

### DIFF
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -16,6 +16,9 @@ package tpm2
 
 import (
 	"bytes"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rsa"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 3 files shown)


tpm2/structures.go
```diff
@@ -90,6 +93,30 @@ func (p Public) Encode() ([]byte, error) {
...
+		if !ok {
+			return nil, fmt.Errorf("can't map TPM EC curve ID 0x%x to Go elliptic.Curve value", p.ECCParameters.CurveID)
+		}
+		pubKey = &ecdsa.PublicKey{
+			X:     p.ECCParameters.Point.X,
+			Y:     p.ECCParameters.Point.Y,
+			Curve: curve,
+		}
+	default:
+		return nil, fmt.Errorf("unsupported primary key type 0x%x", p.Type)
```
awly: `s/primary/public/`

folbricht:
Test added. Thanks

tpm2/tpm2_test.go
```diff
@@ -1084,3 +1084,28 @@ func TestQuote(t *testing.T) {
 		t.Errorf("VerifyPKCS1v15 failed: %v", err)
 	}
 }
+
+func TestReadPublicKey(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	keyHandle, pubKey1, err := CreatePrimary(rw, HandleOwner, pcrSelection, emptyPassword, defaultPassword, defaultKeyParams)
```
awly: This effectively compares pub.Key() output against itself, since CreatePrimary calls it.
The only thing this test proves is that CreatePrimary and ReadPublic commands return the same Public.

Could you rewrite this test using LoadExternal instead?

folbricht:
You're right. Re-done the test. If you prefer, I could add that to the existing `TestLoadExternalPublicKey` test instead to reduce some duplication.

awly:
This seems good, I'd prefer to keep 1:1 mapping between public funcs and tests. Thanks!

PR #73, opened 2019-02-18
Add Handle Value Constants

folbricht:
These handle value constants are used in call to GetCapability() when iterating over handles of specific types like in this example for NV handles:

```go
tpm2.GetCapability(rwc, tpm2.CapabilityHandles, 1, uint32(tpm2.NVIndexFirst))
```

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -151,7 +151,20 @@ type TPMProp uint32
 
 // TPM Capability Properties.
 const (
-	NVMaxBufferSize TPMProp = 0x100 + 44
+	NVMaxBufferSize    TPMProp = 0x100 + 44
+	PCRFirst           TPMProp = 0x00000000
+	HMACSessionFirst   TPMProp = 0x02000000
+	LoadedSessionFirst TPMProp = 0x02000000
+	PolicySessionFirst TPMProp = 0x03000000
+	ActiveSessionFirst TPMProp = 0x03000000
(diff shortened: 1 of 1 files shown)


tpm2/constants.go
```diff
@@ -151,7 +151,20 @@ type TPMProp uint32
...
 // TPM Capability Properties.
 const (
-	NVMaxBufferSize TPMProp = 0x100 + 44
+	NVMaxBufferSize    TPMProp = 0x100 + 44
+	PCRFirst           TPMProp = 0x00000000
+	HMACSessionFirst   TPMProp = 0x02000000
+	LoadedSessionFirst TPMProp = 0x02000000
+	PolicySessionFirst TPMProp = 0x03000000
+	ActiveSessionFirst TPMProp = 0x03000000
+	TransientFirst     TPMProp = 0x80000000
```
awly: There are no `*Last` properties for the above consts?

folbricht:
The *Last values do exist in the spec, but are defined with a value that is implementation-dependent, so not really a constant that applies to all of them. These are defined like this:

TRANSIENT_LAST = (TRANSIENT_FIRST+MAX_LOADED_OBJECTS-1)

With the spec saying:

`A similar approach is used for transient objects with the first assigned handle at the bottom of the range
defined by TPM_HT_TRANSIENT and the top of the range determined by the implementation-dependent value of MAX_LOADED_OBJECTS.`

PR #27, opened 2018-07-19
Added ability to seal/unseal data to a PCR.

nathaniellehrer:
Included an example demonstrating the seal/unseal flow.

PR #74, opened 2019-02-22
Implement TPM2_EncryptDecrypt2

samdamana:
(Recreation of closed PR #37)

Symmetric encryption and decryption. There is some convenience chopping logic in case processed data can't fit into one command.

#37 was closed because of some concerns about exposing an API that is too easily misused and the PR lost priority. A strong need for this functionality within the Go interface has become apparent again and I think this is still an acceptable implementation for `TPM2_EncryptDecrypt2`. I believe tpm2.go should be the home for 1:1 mappings of all TPM2 commands even if the command doesn't effectively convey is security guarantees and could be easily misused. Interested to hear feedback! :smiley:


awly:
cc @tmroeder who raised concerns about this last time.
I agree with Sam, the fact that TPM exposes functionality that could be mis-used is a problem with the spec, and should be addressed there.
Omitting a useful piece of functionality because of this isn't the right solution IMO.

For context, our use-case is: storing a sensitive credential (TLS client private key, provisioned per-VM) on disk encrypted instead of plaintext. Using a TPM key for encryption binds it to the host and prevents exfiltration via filesystem access (although it could still be dumped from memory).

tpm2/tpm2.go
```diff
@@ -1368,3 +1368,116 @@ func ReadPCR(rw io.ReadWriter, pcr int, hashAlg Algorithm) ([]byte, error) {
 	}
 	return pcrVal, nil
 }
+
+// EncryptSymmetric encrypts data using a symmetric key. iv is the
```
tmroeder: nit: Please reword to avoid having a lower-case code identifier as the first word in a sentence. E.g., this could be "The iv argument may be ..."

awly:
> As I understand it, the additional chunking is just passing more bytes into the same encryption operation (that is, using the same IV and key), right?

Right, this is for encryption/decryption of data larger than the TPM command buffer size (~1KB).
TPM2_EncryptDecrypt also returns a new IV for the next block on each call.

> I would _really_ like to have a strong warning on this and any other potentially dangerous operations, though.

I agree. @samdamana could you add a big scary warning to both funcs?

> I hope that your encryption is also using integrity protection in addition to encryption, because there are otherwise attacks that might be possible, depending on the exact details of your threat model.

Can you clarify what the possible attacks are? And is there a TPM primitive that will give us encryption + integrity?

I (naively) assume that if encrypted blob gets modified, decryption fails or the resulting data is corrupt. The latter would just make the upstream user of the ecnrypted credential (TLS key) fail. But our main goal is exfiltration protection here.

tmroeder:
> > As I understand it, the additional chunking is just passing more bytes into the same encryption operation (that is, using the same IV and key), right?
> 
> Right, this is for encryption/decryption of data larger than the TPM command buffer size (~1KB).
> TPM2_EncryptDecrypt also returns a new IV for the next block on each call.
> 
> > I would _really_ like to have a strong warning on this and any other potentially dangerous operations, though.
> 
> I agree. @samdamana could you add a big scary warning to both funcs?
> 
> > I hope that your encryption is also using integrity protection in addition to encryption, because there are otherwise attacks that might be possible, depending on the exact details of your threat model.
> 
> Can you clarify what the possible attacks are? And is there a TPM primitive that will give us encryption + integrity?
> 
> I (naively) assume that if encrypted blob gets modified, decryption fails or the resulting data is corrupt. The latter would just make the upstream user of the ecnrypted credential (TLS key) fail. But our main goal is exfiltration protection here.

Basic encryption modes like CTR, CFB or CBC don't detect ciphertext modification. Instead, the data returned is modified, sometimes in ways that the attacker can control. If you're just encrypting the bytes of the key itself, then modifications will corrupt the key, and you might be OK, if you're guaranteed that using a corrupted key won't cause some other problem, and will just fail. But if the plaintext has any structure to it (like a serialized protobuf), it might be possible to make controlled changes to the decrypted plaintext by XOR'ing bytes against the ciphertext.

The fundamental problem is that modes that provide encryption without integrity are not proven secure in a model where adversaries can modify ciphertext and pass it to a decryption oracle. They are only proven secure against adversaries that can cause encryption to happen and observe the ciphertexts.

TPM2_Create is the TPM 2.0 equivalent of the Seal operation, and that should provide both encryption and integrity. You use TPM2_Unseal to get the data back. Then you don't have to use 
low-level commands.

Support for this use case of TPM2_Create was added in PR #27.

awly:
Attacker making controlled changes to the decrypted data _should_ be ok in our case. At worst, they cause minor DOS (they key is rotated eventually anyway) or exploit a vuln in key parsing code.
But that would require disk write access as root, which they can probably do worse things with.

TPM2_Create (Seal) is where we started. https://github.com/google/go-tpm/blob/master/examples/tpm2-seal-unseal/main.go following this example we can get only ~128 bytes of secret data in. That's less than the private key size we need to encrypt.

We were advised to: create a separate block cipher key, encrypt our private key with that and then seal the block cipher key with TPM2_Create.
@immutableT suggested that just using TPM2_EncryptDecrypt achieves the same without the need to store an additional key.

Either solution would improve the status quo (storing plaintext key on disk), we'd like to avoid extra complexity as much as possible.

tpm2/tpm2.go
samdamana: Fixed.

tmroeder:
> Attacker making controlled changes to the decrypted data _should_ be ok in our case. At worst, they cause minor DOS (they key is rotated eventually anyway) or exploit a vuln in key parsing code.
> But that would require disk write access as root, which they can probably do worse things with.
> 
> TPM2_Create (Seal) is where we started. https://github.com/google/go-tpm/blob/master/examples/tpm2-seal-unseal/main.go following this example we can get only ~128 bytes of secret data in. That's less than the private key size we need to encrypt.
> 
> We were advised to: create a separate block cipher key, encrypt our private key with that and then seal the block cipher key with TPM2_Create.
> @immutableT suggested that just using TPM2_EncryptDecrypt achieves the same without the need to store an additional key.
> 
> Either solution would improve the status quo (storing plaintext key on disk), we'd like to avoid extra complexity as much as possible.

Ah, OK.

In that case, the easiest solution that gives you encryption with integrity is probably to use an AEAD using a key that you seal with TPM2_Create. AFAIK, none of the modes in TPM2_EncryptDecrypt2 give you encryption with integrity, so that's not the same as TPM2_Create to wrap a key, then using that key with a mode that gives encryption and integrity.

Either ChaCha20-Poly1305 (https://godoc.org/golang.org/x/crypto/chacha20poly1305) or AES GCM (https://golang.org/pkg/crypto/cipher/#NewGCM) would be fine in this case.

I strongly recommend always using encryption with integrity, since even if you might be able to argue encryption without integrity works for you now, it's fragile and risks problems in the future. Given that TPM2_Create + a call to a common library is fairly simple, that would be my recommendation. However, if you haven't had our internal Google crypto folks review it, you might want to get their take, too.

immutableT:
When discussing the scenario that @awly and @samdamana are working on, it seemed to me that sealing was not the appropriate primitive, since (at least now) we are not leveraging PCR measurements. Therefore, I suggested using symmetric encryption to protect tokens. 

I agree with @tmroeder that integrity validation is important and we should implement unless there is some technical blocker.
Assuming that we can't use TPM2_Create that gives confidentiality and integrity, we could combine symmetric encryption with HMAC to achieve both.

awly:
@tmroeder PTAL
We decided to go with HMAC in addition to TPM2_EncryptDecrypt as @immutableT described.
Integrity would be nice-to-have in our case, but not critical, so we'll implement it incrementally.

tmroeder:
> @tmroeder PTAL
> We decided to go with HMAC in addition to TPM2_EncryptDecrypt as @immutableT described.
> Integrity would be nice-to-have in our case, but not critical, so we'll implement it incrementally.

FWIW, using encryption + HMAC can in fact give you an AEAD (i.e., encryption with integrity) just as well as the integrated AEAD modes can, as long as you do it right: if M is your message, along with IV, and keys K1 (for encryption) and K2 for (HMAC):

C = Enc(K1, M; IV)
T = HMAC(K2, IV || C)

and store IV || C || T  ( == initialization vector, ciphertext, and tag), then you have just as strong properties as with GCM or ChaCha20-Poly1305. The HMAC protects the ciphertext and IV from modification, and you always check it before trying to decrypt: Given IV || C || T, and K1, K2 as before:

if (HMAC(K2, IV ||  C) != T) // must check in constant time
  break

M = Dec(K1, C; IV)

tpm2/tpm2.go
```diff
@@ -1368,3 +1368,126 @@ func ReadPCR(rw io.ReadWriter, pcr int, hashAlg Algorithm) ([]byte, error) {
...
+	return encryptDecryptSymmetric(rw, keyAuth, key, iv, data, false)
+}
+
+// DecryptSymmetric decrypts data using a symmetric key.
+//
+// WARNING: This command performs low-level cryptographic operations.
+// Secure use of this command is subtle and requires careful analysis.
+// Please consult with experts in cryptography for how to use it securely.
+//
+// The iv  is the initialization vector. The iv must not be empty and its size
```
tmroeder: nit: please remove the extra space between "iv" and "is"

samdamana: Good catch. Fixed.

immutableT:
@tmroeder Yes this is what we discussed - use K2 for integrity validation via HMAC.

samdamana:
Thanks @tmroeder that is a great description of the steps and will be useful to refer to when it is implemented.

Before I can do that the TPM2_HMAC primitive needs to be exposed in tpm2.go. I will probably end up making that PR in a few weeks or so. :+1:

PR #75, opened 2019-02-23
Poll the fd of a TPM device before reading

folbricht:
## Problem

When using a real TPM2.0 chip, I regularly (about 50% of all attempts) get EOF when reading the response to a command in https://github.com/google/go-tpm/blob/46f85b10c39a433eaaf03f55e3836663bf1ccb4a/tpmutil/run.go#L52-L56 This appears to be due to the device not being ready to be read from immediately after sending the command, returning EOF. Even adding the slightest additional delay between Write() and Read() works around the issue. I'm not sure this has always been happening, but it's been consistently happening with kernel 4.20 at least, impacting /dev/tpm0 as well as /dev/tpmrm0.

## Proposed solution

Poll the FD until it's ready for reading. Since this code is synchronous, the timeout is set to -1 (block). It's similar to what tpm2-tss is doing in https://github.com/tpm2-software/tpm2-tss/blob/fa423ad53c17d76284181d1715aeadecd9f438d6/src/tss2-tcti/tcti-device.c#L164-L179

### DIFF
--- a/tpmutil/poll_linux.go
+++ b/tpmutil/poll_linux.go
@@ -1,0 +1,20 @@
+package tpmutil
+
+import (
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// Poll blocks until the file descriptior is ready for reading or an error occurs.
+func poll(f *os.File) error {
+	const (
+		events  = 0x001 // POLLIN
+		timeout = -1    // TSS2_TCTI_TIMEOUT_BLOCK
+	)
+	pollfds := []unix.PollFd{
+		{Fd: int32(f.Fd()), Events: events},
+	}
+	_, err := unix.Poll(pollfds, timeout)
+	return err
+}
--- a/tpmutil/poll_other.go
+++ b/tpmutil/poll_other.go
@@ -1,6 +1,10 @@
+// +build windows
+
 package tpmutil
 
-import "os"
+import (
+	"os"
+)
 
 // Not implemented on Windows.
 func poll(f *os.File) error { return nil }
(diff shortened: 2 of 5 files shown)


tpmutil/poll_linux.go
```diff
@@ -0,0 +1,20 @@
...
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// Poll blocks until the file descriptior is ready for reading or an error occurs.
+func poll(f *os.File) error {
+	const (
+		events  = 0x001 // POLLIN
+		timeout = -1    // TSS2_TCTI_TIMEOUT_BLOCK
```
awly: Clarify that this value means infinite timeout

tpmutil/poll_linux.go
```diff
@@ -0,0 +1,20 @@
...
+	)
+	pollfds := []unix.PollFd{
```
awly: pollfds->pollFds

folbricht:
I don't think there actually is a default build tag `unix` yet. So I defined it as `!windows` for now.

Btw, there are already `run_linux.go` and `run_windows.go` in the same dir, preventing build on darwin for example. I can handle that in a separate PR later.

awly:
`run_other.go` instead of `run_windows.go` should fix the build on all platforms.

As for `!windows`, I'm not confident that syscall will work on all supported GOOS https://github.com/golang/go/blob/master/src/go/build/syslist.go#L7

Let's do `// +build linux darwin` for `run_unix.go` and `// +build !linux,!darwin` for `run_other.go`

folbricht:
Done. Also opened a separate https://github.com/google/go-tpm/pull/76 (unrelated to the change in this PR) to allow non-Windows/Linux builds.

PR #76, opened 2019-02-26
Support building on darwin and other OS

folbricht:
This PR enables building on darwin and other OS. There are no syscalls or other low-level operations in here and only Windows appears be needing a different file.


awly:
Thanks!

PR #78, opened 2019-03-02
Implement RSAEncrypt() and RSADecrypt()

folbricht:
This PR implements a couple of the asymmetric primitives, namely RSAEncrypt() and RSADecrypt().

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -1505,6 +1505,9 @@ func encodeRSAEncrypt(key tpmutil.Handle, message []byte, scheme *AsymScheme, la
 	if err != nil {
 		return nil, err
 	}
+	if label != "" {
+		label += "\x00"
+	}
 	l, err := tpmutil.Pack([]byte(label))
 	if err != nil {
 		return nil, err
@@ -1520,7 +1523,7 @@ func decodeRSAEncrypt(resp []byte) ([]byte, error) {
 
 // RSAEncrypt performs RSA encryption in the TPM according to RFC 3447. The key must be
 // a (public) key loaded into the TPM beforehand. Note that when using OAEP with a label,
-// the label needs to be null-terminated and the null byte is included in the padding
+// a null byte is appended to the label and the null byte is included in the padding
 // scheme.
 func RSAEncrypt(rw io.ReadWriter, key tpmutil.Handle, message []byte, scheme *AsymScheme, label string) ([]byte, error) {
 	cmd, err := encodeRSAEncrypt(key, message, scheme, label)
@@ -1551,6 +1554,9 @@ func encodeRSADecrypt(key tpmutil.Handle, password string, message []byte, schem
 	if err != nil {
 		return nil, err
 	}
+	if label != "" {
+		label += "\x00"
+	}
 	l, err := tpmutil.Pack([]byte(label))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


tpm2/tpm2.go
```diff
@@ -1491,3 +1491,90 @@ func encryptDecryptSymmetric(rw io.ReadWriteCloser, keyAuth string, key tpmutil.
...
+}
+
+func decodeRSADecrypt(resp []byte) ([]byte, error) {
+	var out []byte
+	var paramSize uint32
+	_, err := tpmutil.Unpack(resp, &paramSize, &out)
+	return out, err
+}
+
+// RSADecrypt performs RSA decryption with a padding scheme according to RFC 3447. When using OAEP
```
tmroeder: The code itself looks fine, but the comment text made me nervous. It sounded to me like this code was implementing RSA encryption and decryption, using some primitives from the TPM, which is fraught with danger and difficulty.

However, what it appears to be doing instead of exposing the TPM RSA API. Please consider changing the comments to make it clear that this isn't an implementation of RSA, just an interface for performing RSA operations in the TPM.

folbricht:
Reworded the comments a little. Hope it's more clear now. Any suggestions are welcome.

I left the note about the label when using OAEP since I ran into that. Can remove that if you think it's out of place. It's just a little gotcha when working with https://golang.org/pkg/crypto/rsa/#EncryptOAEP which doesn't require the null byte.

tpm2/tpm2.go
```diff
@@ -1491,3 +1491,92 @@ func encryptDecryptSymmetric(rw io.ReadWriteCloser, keyAuth string, key tpmutil.
...
+
+func decodeRSAEncrypt(resp []byte) ([]byte, error) {
+	var out []byte
+	_, err := tpmutil.Unpack(resp, &out)
+	return out, err
+}
+
+// RSAEncrypt performs RSA encryption in the TPM according to RFC 3447. The key must be
+// a (public) key loaded into the TPM beforehand. Note that when using OAEP with a label,
+// the label needs to be null-terminated and the null byte is included in the padding
```
awly: We should null-terminate the string internally

tpm2/tpm2.go
```diff
@@ -1491,3 +1491,92 @@ func encryptDecryptSymmetric(rw io.ReadWriteCloser, keyAuth string, key tpmutil.
...
+// scheme.
+func RSAEncrypt(rw io.ReadWriter, key tpmutil.Handle, message []byte, scheme *AsymScheme, label string) ([]byte, error) {
```
awly: What's the limit on message size? Is there a way to chain RSAEncrypt calls to encrypt a large blob?

tpm2/tpm2.go
folbricht: Done, only if a label is provided, since using just a null as label changes the results. And adding an extra null to a string that's already null-terminated doesn't have any impact (the TPM uses the first null it finds as per sped).

tpm2/tpm2.go
folbricht: The size of the message that can be en-/decrypted depends on the size of the modulus, the algorithm used as well as the hash size. The random-looking numbers I used in the test represent the max for each combination, using a 2048 bit key. The formulas to calculate the max message size can be found in the spec, part 3, table 44. I'm happy to add those to a comment if you prefer.

As for use case, I imagine this can be used to encrypt a large blob with a symmetric key, then encrypt the symmetric key with the public RSA key (in software or a TPM). This key then can only be decrypted in the target TPM, with the blob then presumably decrypted in software using the symmetric key.

PR #81, opened 2019-03-27
Remote attestation

DenisKarch:
Added functionality that is commonly used for remote attestation.

Added:
  - GetCapability
  - GetManufacturer
  - NVReadValue
  - ReadEKCert
  - ActivateIdentity

### DIFF
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -18,6 +18,8 @@ package tpm
 import (
 	"bytes"
 	"crypto"
+	"crypto/aes"
+	"crypto/cipher"
 	"crypto/hmac"
 	"crypto/rand"
 	"crypto/rsa"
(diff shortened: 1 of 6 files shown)


tpmutil/structures.go
```diff
@@ -48,3 +48,6 @@ type responseHeader struct {
 
 // A Handle is a reference to a TPM object.
 type Handle uint32
+
+// A U16Bytes is a byte slice with 16-bit header.
+type U16Bytes []byte
```
ericchiang: Why is this necessary? The encoding logic looks identical to []byte.

tpm/tpm.go
```diff
@@ -713,6 +851,103 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
 	return tpmutil.Pack(pk)
 }
 
+// ReadEKCert reads the EKCert from the NVRAM
+func ReadEKCert(rw io.ReadWriter, ownAuth digest) ([]byte, error) {
+	var (
```
ericchiang: nit: remove this var block. you don't need it and can just use `:=` to allocate new variables.

tpm/tpm.go
```diff
@@ -713,6 +851,103 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+		ekbuf    []byte
+		buff     io.Reader
+		offset   uint32
+		bufSize  uint32
+		tag      uint16
+		certType uint8
+		certSize uint16
+	)
+	offset = 0
+	data, err := NVReadValue(rw, 0x1000f000, offset, 5, ownAuth)
```
ericchiang: if you ever have a raw value like this do the following instead:

```
const descriptiveName = 0x1000f000
data, err := NVReadValue(rw, descriptiveName, offset, 5, ownAuth)
```

tpmutil/structures.go
DenisKarch: The encoding is identical for []byte on TPM2.0. However tpmutil assumes that TPM1.2 uses a 32-bit header which--although true in most cases--is not the case for the TPM_SYMMETRIC_KEY that is returned from ActivateIdentity. I mentioned this in the commit comment but I'll add a comment in the file as well.

DenisKarch:
PR split into 3
getCapability
https://github.com/google/go-tpm/pull/82
ReadEKCert
https://github.com/google/go-tpm/pull/83
ActivateIdentity
https://github.com/google/go-tpm/pull/84

PR #82, opened 2019-03-28
Added ActivateIdentity

DenisKarch:
Adding ActivateIdentity required modifying tpmutil\encoding.go in order
to properly parse TPM_SYMMETRIC_KEY which has a slice whose length is a
uint16 (as opposed to uint32 as with most other TPM1.2 structs)

### DIFF
--- a/tpm/structures.go
+++ b/tpm/structures.go
@@ -17,6 +17,7 @@ package tpm
 import (
 	"crypto"
 	"crypto/rsa"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
@@ -239,6 +240,57 @@ type key12 struct {
 type pubKey struct {
 	AlgorithmParams keyParams
 	Key             []byte
+}
+
+// A symKey is a TPM representation of a symmetric key.
+type symKey struct {
+	AlgID     uint32
+	EncScheme uint16
+	Key       []byte
+}
+
+// TPMMarshal will marshal the symKey into a byte Buffer as specified the TPM spec.
+func (k *symKey) TPMMarshal(out io.Writer) error {
+	if err := binary.Write(out, binary.BigEndian, &k.AlgID); err != nil {
+		return err
+	}
+	if err := binary.Write(out, binary.BigEndian, &k.EncScheme); err != nil {
+		return err
+	}
+	if err := binary.Write(out, binary.BigEndian, uint16(len(k.Key))); err != nil {
+		return err
+	}
+	if err := binary.Write(out, binary.BigEndian, k.Key); err != nil {
+		return err
+	}
+	return nil
+}
+
+// TPMUnmarshal will parse the byte Buffer and populate the symKey.
+func (k *symKey) TPMUnmarshal(in io.Reader) error {
+	if err := binary.Read(in, binary.BigEndian, &k.AlgID); err != nil {
+		return err
+	}
+	if err := binary.Read(in, binary.BigEndian, &k.EncScheme); err != nil {
+		return err
+	}
+	var size uint16
+	if err := binary.Read(in, binary.BigEndian, &size); err != nil {
+		return err
+	}
+	key := make([]byte, size)
+	if err := binary.Read(in, binary.BigEndian, &key); err != nil {
+		return err
+	}
+	k.Key = key
+	return nil
+}
+
+// TPMPackedSize will return the expected size in bytes of the marshaled symKey.
+func (k *symKey) TPMPackedSize() int {
+	// AlgID  + EncScheme + len(Key) + Key
+	// uint32 + uint16    + uint16   + BYTE[]
+	return binary.Size(k.AlgID) + binary.Size(k.EncScheme) + binary.Size(uint16(len(k.Key))) + binary.Size(k.Key)
 }
 
 // A tpmStoredData holds sealed data from the TPM.
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -18,12 +18,15 @@ package tpm
 import (
 	"bytes"
 	"crypto"
+	"crypto/aes"
+	"crypto/cipher"
 	"crypto/hmac"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/sha1"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 
 	"github.com/google/go-tpm/tpmutil"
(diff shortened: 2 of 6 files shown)


DenisKarch:
cc @tmroeder

tpm/structures.go
```diff
@@ -241,6 +242,57 @@ type pubKey struct {
...
 
+// A symKey is a TPM representation of a symmetric key.
+type symKey struct {
+	AlgID     uint32
+	EncScheme uint16
+	Key       []byte
+}
+
+// TPMMarshal will marshal the symKey into a byte Buffer as specified the TPM spec.
+func (k *symKey) TPMMarshal(out io.Writer) error {
```
awly: Maybe only add a custom type and marshaler for the inner key?
The other fields can be handled normally in tpmutil

tpmutil/structures.go
```diff
@@ -48,3 +50,11 @@ type responseHeader struct {
 
 // A Handle is a reference to a TPM object.
 type Handle uint32
+
+// TPMSelfMarshaler is used to support parsing new TPM structs without adding special
+// cases to Pack and UnpackBuf
+type TPMSelfMarshaler interface {
```
awly: Just SelfMarshaler, since it'll be used as `tpmutil.SelfMarshaler`

tpmutil/encoding.go
```diff
@@ -54,6 +54,12 @@ func packedSize(elts ...interface{}) (int, error) {
 		v := reflect.ValueOf(e)
 		switch v.Kind() {
 		case reflect.Ptr:
+			// Check if it's a TPMSelfMarshaler
+			marshaler, ok := e.(TPMSelfMarshaler)
```
awly: Check this outside of the switch, regardless of e's reflect kind.
Same for all other places where it's used

tpmutil/structures.go
```diff
@@ -48,3 +50,11 @@ type responseHeader struct {
...
+	TPMMarshal(out io.Writer) error
+	TPMUnmarshal(in io.Reader) error
+	TPMpackedSize() int
```
awly: `TPMPackedSize` (capital 'P')

tpmutil/structures.go
```diff
@@ -48,3 +50,11 @@ type responseHeader struct {
 
 // A Handle is a reference to a TPM object.
 type Handle uint32
+
+// TPMSelfMarshaler is used to support parsing new TPM structs without adding special
```
awly: Describe primarily what the type represents, not what the motivation behind it is.
`TPMSelfMarshaler allows custom types to override default encoding/decoding behavior in Pack, Unpack and UnpackBuf.`

tpm/structures.go
```diff
@@ -241,6 +242,57 @@ type pubKey struct {
...
+	}
+	k.Key = key
+	return nil
+}
+
+// TPMpackedSize will return the expected size in bytes of the marshaled symKey.
+func (k *symKey) TPMpackedSize() int {
+	// AlgID  + EncScheme + len(Key) + Key
+	// uint32 + uint16    + uint16   + BYTE[]
+	return 8 + len(k.Key)
```
awly: Use `binary.Size` to compute values, not hardcoded const.

tpm/structures.go
DenisKarch: I felt like adding another type that would just be an alias for []byte was a bit ugly so I decided to let the symKey implement the interface instead.

awly: But this ends up doing more work for no reason.
Nothing ugly about an alias type for `[]byte`.

DenisKarch: Just tried making an alias to []byte satisfy the interface. Ran into a problem with pack and unpack parsing structs differently. For unpack we recursively unpack **pointers** to the fields, whereas with pack it packs the fields directly. Thus when unpacking the interface is satisfied but when packing it is not. I only recently picked up go so I'm not too familiar with the internals, but this does not seem to be the case when having a struct satisfy the interface (both struct and *struct will satisfy and interface). That said, in principle I think having entire tpm structs be self marshaled makes more sense than making "new" primitives selfmarshaling.

tpm/tpm.go
```diff
@@ -628,6 +631,125 @@ func MakeIdentity(rw io.ReadWriter, srkAuth []byte, ownerAuth []byte, aikAuth []
 	return blob, nil
 }
 
+func unloadTrspiCred(blob []byte) ([]byte, error) {
+	/*
+	 * Trousers expects the asym blob to have an additional data in the header.
```
tmroeder: This comment doesn't give me enough context about why this function is here. Naively, it seems like a layering violation for this code to have to take into account anything about the Trousers code above it.

Please expand the explanation. Does this mean, for example, that this library is now specific to Trousers?

tpm/tpm.go
```diff
@@ -628,6 +631,125 @@ func MakeIdentity(rw io.ReadWriter, srkAuth []byte, ownerAuth []byte, aikAuth []
...
+	 * The relevant data is duplicated in the TPM_SYMMETRIC_KEY struct so we parse
+	 * and throw the header away.
+	 * Trspi_UnloadBlob_UINT32(offset, &credSize, blob);
```
tmroeder: I don't think this C code belongs in the comment. Instead, please explain this at a higher level

tpm/structures.go
awly: That all sounds like `tpmutil` doesn't do the right thing when marshalling.
There's no reason why we couldn't handle a field implementing an interface when marsahlling.

Can you leave a TODO (for me) to fix this?
I think it's too far into the weeds to block this PR.

tpm/tpm.go
```diff
@@ -628,6 +631,125 @@ func MakeIdentity(rw io.ReadWriter, srkAuth []byte, ownerAuth []byte, aikAuth []
...
+	}
+	cred := make([]byte, header.Credsize)
+	if err := binary.Read(symbuf, binary.BigEndian, cred); err != nil {
+		return nil, err
+	}
+	return cred, nil
+}
+
+// ActivateIdentity sks the TPM to decrypt an EKPub encrypted symmetric session key
+// which it uses to decrpyt the symmetrically encrypted secret.
```
tmroeder: typo: "decpyt" -> "decrypt"

tpm/tpm.go
```diff
@@ -628,6 +631,125 @@ func MakeIdentity(rw io.ReadWriter, srkAuth []byte, ownerAuth []byte, aikAuth []
...
+func ActivateIdentity(rw io.ReadWriter, aikAuth []byte, ownerAuth []byte, aik tpmutil.Handle, asym, sym []byte) ([]byte, error) {
+	// Run OIAP for the AIK
+	oiaprAIK, err := oiap(rw)
+	if err != nil {
+		return nil, fmt.Errorf("failed to start OIAP session: %v", err)
+	}
+
+	// Run OSAP for the Owner, reading a random OddOSAP for our initial command
```
tmroeder: nit: no need to capitalize "Owner" here

tpm/tpm.go
```diff
@@ -628,6 +631,125 @@ func MakeIdentity(rw io.ReadWriter, srkAuth []byte, ownerAuth []byte, aikAuth []
...
+	case esSymCTR:
+		stream := cipher.NewCTR(block, iv)
+		stream.XORKeyStream(secret, ciphertxt)
+	case esSymOFB:
+		stream := cipher.NewOFB(block, iv)
+		stream.XORKeyStream(secret, ciphertxt)
+	case esSymCBCPKCS5:
+		mode := cipher.NewCBCDecrypter(block, iv)
+		mode.CryptBlocks(secret, ciphertxt)
+		// remove PKCS5 padding
```
tmroeder: nit: please make comments into full sentences, with an initial capital and final punctuation.

tpmutil/encoding.go
```diff
@@ -51,6 +51,12 @@ func UseTPM20LengthPrefixSize() {
 func packedSize(elts ...interface{}) (int, error) {
 	var size int
 	for _, e := range elts {
+		// Check if it's a SelfMarshaler
```
tmroeder: No need for this comment, since it repeats exactly what the code is already doing.

Please also remove this comment from the other places this happens below.

DenisKarch:
Addressed everything except for Trousers comment (discussing in messages).

PR #83, opened 2019-03-28
Add NVReadValue and ReadEKCert

DenisKarch:
NVReadValue consumes a location and a length and tries to read length
bytes from the requested location in NVRAM.

ReadEKCert reads the EKCert at the default location in NVRAM.

### DIFF
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -24,6 +24,7 @@ import (
 	"crypto/sha1"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 
 	"github.com/google/go-tpm/tpmutil"
@@ -711,6 +712,112 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
 	}
 
 	return tpmutil.Pack(pk)
+}
+
+// ReadEKCert reads the EKCert from the NVRAM.
+// The TCG PC Client specifies additional headers that are to be stored with the EKCert, we parse them
+// here and return only the DER encoded certificate.
+// TCG PC Client Specific Implementation Specification for Conventional BIOS 7.4.4
+// https://www.trustedcomputinggroup.org/wp-content/uploads/TCG_PCClientImplementation_1-21_1_00.pdf
+func ReadEKCert(rw io.ReadWriter, ownAuth digest) ([]byte, error) {
+	const (
+		certIndex                 = 0x1000f000 // TPM_NV_INDEX_EKCert (TPM Main Part 2 TPM Structures 19.1.2)
+		certTagPCClientStoredCert = 0x1001     // TCG_TAG_PCCLIENT_STORED_CERT
+		certTagPCClientFullCert   = 0x1002     // TCG_TAG_PCCLIENT_FULL_CERT
+		tcgFullCert               = 0          // TCG_FULL_CERT
+		tcgPartialSmallCert       = 1          // TCG_PARTIAL_SMALL_CERT
+	)
+	offset := uint32(0)
+	var header struct {
+		Tag      uint16
+		CertType uint8
+		CertSize uint16
+	}
+
+	data, err := NVReadValue(rw, certIndex, offset, uint32(binary.Size(header)), ownAuth)
+	if err != nil {
+		return nil, err
+	}
+	offset = offset + uint32(binary.Size(header))
+	buff := bytes.NewReader(data)
+
+	if err := binary.Read(buff, binary.BigEndian, &header); err != nil {
+		return nil, err
+	}
+
+	if header.Tag != certTagPCClientStoredCert {
+		return nil, fmt.Errorf("invalid certificate")
+	}
+
+	var bufSize uint32
+	switch header.CertType {
+	case tcgFullCert:
+		var tag uint16
+		data, err := NVReadValue(rw, certIndex, offset, uint32(binary.Size(tag)), ownAuth)
+		if err != nil {
+			return nil, err
+		}
+		bufSize = uint32(header.CertSize) + offset
+		offset = offset + uint32(binary.Size(tag))
+		buff = bytes.NewReader(data)
+
+		if err := binary.Read(buff, binary.BigEndian, &tag); err != nil {
+			return nil, err
+		}
+
+		if tag != certTagPCClientFullCert {
+			return nil, fmt.Errorf("certificate type and tag do not match")
+		}
+	case tcgPartialSmallCert:
+		return nil, fmt.Errorf("certType is not TCG_FULL_CERT: currently do not support partial certs")
+	default:
+		return nil, fmt.Errorf("invalid certType: 0x%x", header.CertType)
+	}
+
+	var ekbuf []byte
+	for offset < bufSize {
+		length := bufSize - offset
+		// TPMs can only read so much memory per command so we read in 128byte chunks.
+		// 128 was taken from go-tspi. The actual max read seems to be platform dependent
+		// but cannot be queried on TPM1.2 (and does not seem to appear in any documentation).
+		if length > 128 {
+			length = 128
+		}
+		data, err = NVReadValue(rw, certIndex, offset, length, ownAuth)
+		if err != nil {
+			return nil, err
+		}
+
+		ekbuf = append(ekbuf, data...)
+		offset += length
+	}
+
+	return ekbuf, nil
+}
+
+// NVReadValue returns the value from a given index, offset, and length in NVRAM.
+// See TPM-Main-Part-2-TPM-Structures 19.1.
+func NVReadValue(rw io.ReadWriter, index, offset, len uint32, ownAuth digest) ([]byte, error) {
+	sharedSecretOwn, osaprOwn, err := newOSAPSession(rw, etOwner, khOwner, ownAuth[:])
+	if err != nil {
+		return nil, fmt.Errorf("failed to start new auth session: %v", err)
+	}
+	defer osaprOwn.Close(rw)
+	defer zeroBytes(sharedSecretOwn[:])
+	authIn := []interface{}{ordNVReadValue, index, offset, len}
+	ca, err := newCommandAuth(osaprOwn.AuthHandle, osaprOwn.NonceEven, sharedSecretOwn[:], authIn)
+	if err != nil {
+		return nil, fmt.Errorf("failed to construct owner auth fields: %v", err)
+	}
+	data, ra, ret, err := nvReadValue(rw, index, offset, len, ca)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read from NVRAM: %v", err)
+	}
+	raIn := []interface{}{ret, ordNVReadValue, data}
+	if err := ra.verify(ca.NonceOdd, sharedSecretOwn[:], raIn); err != nil {
+		return nil, fmt.Errorf("failed to verify authenticity of response: %v", err)
+	}
+	return data, nil
 }
 
 // OwnerReadPubEK uses owner auth to get a blob representing the public part of the
--- a/tpm/commands.go
+++ b/tpm/commands.go
@@ -143,6 +143,18 @@ func getPubKey(rw io.ReadWriter, keyHandle tpmutil.Handle, ca *commandAuth) (*pu
 	return &pk, &ra, ret, nil
 }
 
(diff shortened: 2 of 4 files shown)


tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
 	return tpmutil.Pack(pk)
 }
 
+// ReadEKCert reads the EKCert from the NVRAM
+func ReadEKCert(rw io.ReadWriter, ownAuth digest) ([]byte, error) {
+	var (
```
ericchiang: please remove this block

tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+		}
+
+		ekbuf = append(ekbuf, data...)
+		offset += length
+	}
+
+	return ekbuf, nil
+}
+
+// NVReadValue consumes an index, offset, and length, and returns the value from that location in NVRAM
```
ericchiang: Can you expand on what NVRAM is and what it holds?

tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+func NVReadValue(rw io.ReadWriter, index, offset, len uint32, ownAuth digest) ([]byte, error) {
+	sharedSecretOwn, osaprOwn, err := newOSAPSession(rw, etOwner, khOwner, ownAuth[:])
+	if err != nil {
+		return nil, fmt.Errorf("newOSAPSession failed: %v", err)
```
ericchiang: nit: Errors should describe what went wrong and not refer to internal methods.

fmt.Errorf("creating osap session: %v", err)

tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+	}
+	defer osaprOwn.Close(rw)
+	defer zeroBytes(sharedSecretOwn[:])
+	authIn := []interface{}{ordNVReadValue, index, offset, len}
+	ca, err := newCommandAuth(osaprOwn.AuthHandle, osaprOwn.NonceEven, sharedSecretOwn[:], authIn)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new CommandAuth: %v", err)
```
ericchiang: fmt.Errorf("constructing authenticated tpm command: %v", err)

tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+	}
+	data, ra, ret, err := nvReadValue(rw, index, offset, len, ca)
+	if err != nil {
+		return nil, fmt.Errorf("nvReadValue failed: %v", err)
```
ericchiang: ("read NV value: %v", err)

tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+	offset = 0
+	data, err := NVReadValue(rw, certIndex, offset, 5, ownAuth)
+	if err != nil {
+		return nil, err
+	}
+	offset = offset + 5
+	buff = bytes.NewReader(data)
+
+	if err := binary.Read(buff, binary.BigEndian, &tag); err != nil {
+		return nil, fmt.Errorf("binary.Read failed: %v", err)
```
ericchiang: nit: don't wrap the binary.Read errors, the rest of this file doesn't do it.

tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+	}
+	if err := binary.Read(buff, binary.BigEndian, &certSize); err != nil {
+		return nil, fmt.Errorf("binary.Read failed: %v", err)
+	}
+
+	if tag != 0x1001 { // TCG_TAG_PCCLIENT_STORED_CERT
+		return nil, fmt.Errorf("invalid certificate")
+	}
+
+	switch certType {
```
ericchiang: need a default case

tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+			return nil, fmt.Errorf("certificate type and tag do not match")
+		}
+
+	case 1: // TCG_PARTIAL_SMALL_CERT
+		return nil, fmt.Errorf("certType is not TCG_FULL_CERT: currently do not support partial certs")
+	}
+
+	for offset < bufSize {
+		length := bufSize - offset
+		if length > 128 { // read the cert in 128byte chunks
```
ericchiang: Please explain why you're reading 128 bytes at a time and not some other amount

tpm/tpm.go
```diff
@@ -713,6 +714,104 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+	const certIndex = 0x1000f000
+	offset = 0
+	data, err := NVReadValue(rw, certIndex, offset, 5, ownAuth)
+	if err != nil {
+		return nil, err
+	}
+	offset = offset + 5
+	buff = bytes.NewReader(data)
+
+	if err := binary.Read(buff, binary.BigEndian, &tag); err != nil {
```
ericchiang: this can be

```
var header struct {
    tag uint16
    certType uint8
    certSize uint16
}
if err := binary.Read(buff, binary.BigEndian, &header); err != nil {
    // handle error
}
```

tpm/tpm.go
DenisKarch: I'm actually not 100% certain it should be 128. I took that value from go-tspi. TPMs are limited by how much they can read at a time but I can't find any documentation about what the limit is. I tested it an Infineon TPM on linux and was able to read 1024bytes at a time.

DenisKarch:
Updated code to address review comments.

tpm/tpm.go
```diff
@@ -713,6 +714,103 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+		bufSize = uint32(header.certSize) + offset
+		offset = offset + 2
+		buff = bytes.NewReader(data)
+
+		var tag uint16
+		if err := binary.Read(buff, binary.BigEndian, &tag); err != nil {
+			return nil, err
+		}
+
+		if tag != 0x1002 { // TCG_TAG_PCCLIENT_FULL_CERT
```
ericchiang: These would be better as constants

```
const (
    certTagPCClientFullCert = 0x1002
    certTagPCClientStoredCert = 0x1001
)
```

tpm/tpm.go
```diff
@@ -713,6 +714,103 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+		tag      uint16
+		certType uint8
+		certSize uint16
+	}
+	var bufSize uint32
+	if err := binary.Read(buff, binary.BigEndian, &header); err != nil {
+		return nil, err
+	}
+
+	if header.tag != 0x1001 { // TCG_TAG_PCCLIENT_STORED_CERT
```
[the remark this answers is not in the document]
DenisKarch: Yeah I couldn't find it either at first (I originally copied the code from go-tspi but didn't understand what this additional parsing was doing). Matthew pointed me to this https://www.trustedcomputinggroup.org/wp-content/uploads/TCG_PCClientImplementation_1-21_1_00.pdf section 7 has the structs.

DenisKarch:
Nits addressed. Let me know if there should be more information about the spec in the header.

awly:
@DenisKarch sorry for the delay, was on vacation last week.
I'll try to get to your PRs in the next few days.

DenisKarch:
cc @tmroeder

tpm/tpm.go
```diff
@@ -713,6 +714,106 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+	if err := binary.Read(buff, binary.BigEndian, &header); err != nil {
+		return nil, err
+	}
+
+	if header.Tag != certTagPCClientStoredCert { // TCG_TAG_PCCLIENT_STORED_CERT
+		return nil, fmt.Errorf("invalid certificate")
+	}
+
+	switch header.CertType {
+	case 0: // TCG_FULL_CERT
```
tmroeder: Instead of making the reference a comment, consider making it a constant, like

```go
case tcgFullCert:
```

tpm/tpm.go
```diff
@@ -713,6 +714,106 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+	}
+	offset = offset + 5
+	buff := bytes.NewReader(data)
+
+	var header struct {
+		Tag      uint16
+		CertType uint8
+		CertSize uint16
+	}
+	var bufSize uint32
```
tmroeder: nit: From the placement of this declaration, I assumed it was needed in the next line, but it's not needed until the switch. Consider moving it down to just before the switch to make that clear.

tpm/tpm.go
```diff
@@ -713,6 +714,106 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+	if err := binary.Read(buff, binary.BigEndian, &header); err != nil {
+		return nil, err
+	}
+
+	if header.Tag != certTagPCClientStoredCert { // TCG_TAG_PCCLIENT_STORED_CERT
```
tmroeder: I don't think that having this comment here is as useful as having it on the constant declaration itself, even though that's in the same function as this line.

tpm/tpm.go
```diff
@@ -713,6 +714,106 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+		return nil, fmt.Errorf("invalid certificate")
+	}
+
+	switch header.CertType {
+	case 0: // TCG_FULL_CERT
+		data, err := NVReadValue(rw, certIndex, offset, 2, ownAuth)
```
tmroeder: I'd appreciate understanding why "2" here, just like why "5" in the previous call to NVReadValue. A constant declaration could help make it clear.

tpm/tpm.go
```diff
@@ -713,6 +714,106 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+		}
+
+		ekbuf = append(ekbuf, data...)
+		offset += length
+	}
+
+	return ekbuf, nil
+}
+
+// NVReadValue consumes an index, offset, and length, and returns the value from that location in NVRAM.
```
tmroeder: nit: I wouldn't say that this function "consumes" these values any more than any function consumes its arguments. I don't think that's the terminology this library is using for documenting function parameters. How about something like

"NVReadValue returns the value from a given index, offset, and length in NVRAM."

tpm/tpm.go
```diff
@@ -713,6 +714,106 @@ func OwnerReadSRK(rw io.ReadWriter, ownerAuth digest) ([]byte, error) {
...
+// See TPM-Main-Part-2-TPM-Structures 19.1.
+func NVReadValue(rw io.ReadWriter, index, offset, len uint32, ownAuth digest) ([]byte, error) {
```
tmroeder: Please also add a direct test for NVReadValue. It's useful to be able to quickly distinguish failures of ReadEKCert from failures of NVReadValue.

DenisKarch: I'm not sure adding a test just for NVReadValue is all that useful without NVDefineSpace and NVWriteValue. There are some predefined NV Indices but the only one that can easily be checked/verified is the EKCert. As such maybe it would be better to add more verbose errors and leave the unit test for when DefineSpace and WriteValue are added?

DenisKarch:
Addressed most of the comments. Left a reply about the NVReadValue test.

tpm/tpm.go
tmroeder: OK, it can wait

PR #84, opened 2019-03-28
Add getCapability and GetManufacturer

DenisKarch:
NVReadValue consumes a location and a length and tries to read length
bytes from the requested location in NVRAM.

ReadEKCert reads the EKCert at the default location in NVRAM.

### DIFF
--- a/tpm/commands.go
+++ b/tpm/commands.go
@@ -143,6 +143,21 @@ func getPubKey(rw io.ReadWriter, keyHandle tpmutil.Handle, ca *commandAuth) (*pu
 	return &pk, &ra, ret, nil
 }
 
+// getCapability reads the requested capability and sub-capability from NVRAM
+func getCapability(rw io.ReadWriter, cap, subcap uint32) ([]byte, error) {
+	subCapBytes, err := tpmutil.Pack(subcap)
+	if err != nil {
+		return nil, err
+	}
+	var b []byte
+	in := []interface{}{cap, subCapBytes}
+	out := []interface{}{&b}
+	if _, err := submitTPMRequest(rw, tagRQUCommand, ordGetCapability, in, out); err != nil {
+		return nil, err
+	}
+	return b, nil
+}
+
 func nvReadValue(rw io.ReadWriter, index, offset, len uint32, ca *commandAuth) ([]byte, *responseAuth, uint32, error) {
 	var b []byte
 	var ra responseAuth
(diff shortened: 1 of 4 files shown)


tpm/commands.go
```diff
@@ -143,6 +143,17 @@ func getPubKey(rw io.ReadWriter, keyHandle tpmutil.Handle, ca *commandAuth) (*pu
 	return &pk, &ra, ret, nil
 }
 
+// getCapability reads the requested capability and sub-capability from NVRAM
+func getCapability(rw io.ReadWriter, cap, subcap uint32) ([]byte, error) {
+	var b []byte
+	in := []interface{}{cap, uint32(4), subcap} // tcsd requires capabilities to be of size 4, cannot find any mention of this in the spec
```
ericchiang: This seems to contradict the existing use of ordGetCapability 

https://github.com/google/go-tpm/blob/7a7fe86fbbf203a562400a8a0dbe63a93ae02059/tpm/tpm.go#L46

Any thoughts why this is different?

tpm/tpm.go
```diff
@@ -754,6 +754,11 @@ func ReadPubEK(rw io.ReadWriter) ([]byte, error) {
 	return tpmutil.Pack(pk)
 }
 
+// GetManufacturer returns the manufacturer ID
+func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
+	return getCapability(rw, capProperty, tpmCapPropManufacturer)
```
ericchiang: nit: you probably don't need getCapability as a different method if its only used here.

tpm/commands.go
DenisKarch: It actually doesn't contradict it. The existing use packs the subcapability into a byte slice whereas I have the length and the actual subcap seperate (length+rawBytes). They are both encoded the same way though technically the existing call matches the spec more closely.

tpm/tpm.go
DenisKarch: Although it is only used here it can be used to get a variety of other fields so I figured it would be best to abstract now.

ericchiang: Can you re-use this for GetKeys then?

DenisKarch: Done

DenisKarch:
cc @tmroeder

tmroeder:
LGTM.

My only comment is that the git commit should use the imperative form of the verb rather than the past:

"Add getCapability and GetManufacturer"

awly:
Looks like there's a merge conflict

DenisKarch:
Did a rebase, should work now.

PR #85, opened 2019-04-08
Modify CreateKey

DenisKarch:
CreateKeyEx is identical to CreateKey but it returns all the returns
values instead of just the public and private blobs.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -504,13 +504,10 @@ func create(rw io.ReadWriter, parentHandle tpmutil.Handle, parentPassword, objec
 }
 
 // CreateKey creates a new key pair under the owner handle.
-// Returns private key and public key blobs.
-func CreateKey(rw io.ReadWriter, owner tpmutil.Handle, sel PCRSelection, parentPassword, ownerPassword string, pub Public) ([]byte, []byte, error) {
-	_, private, public, _, _, _, err := create(rw, owner, parentPassword, ownerPassword, nil /*inSensitive*/, pub, sel)
-	if err != nil {
-		return nil, nil, err
-	}
-	return private, public, nil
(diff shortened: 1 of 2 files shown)


tpm2/tpm2.go
```diff
@@ -513,6 +513,14 @@ func CreateKey(rw io.ReadWriter, owner tpmutil.Handle, sel PCRSelection, parentP
 	return private, public, nil
 }
 
+// CreateKeyEx creates a new key pair under the owner handle.
+// Returns private key and public key blobs.
+// This function differs from CreateKey in that all response elements
+// are returned.
+func CreateKeyEx(rw io.ReadWriter, owner tpmutil.Handle, sel PCRSelection, parentPassword, ownerPassword string, pub Public) ([]byte, []byte, error) {
```
tmroeder: The "Ex" suffix is a Windows idiom. That doesn't make it bad or good, but I don't think this API is currently using that naming scheme.

Does this compile? It looks to me like the return type of CreateKeyEx is []byte, []byte, error, but the return type of the create function is much longer.

DenisKarch: Yes, that was my mistake, forgot to call git add before doing commit --amend

DenisKarch: as for Ex, tpm2 seems to use it already for CreatePrimaryEx

tmroeder:
LGTM, but please wait for approval from the reviewers who are more involved in the tpm2 side.

tpm2/tpm2.go
awly: We were using `*Ex` pattern to preserve backwards-compatibility.
Now that this repo is a Go module, we can break compatibility and tag new versions.

Please change `CreatePrimary` instead and name all the return params. All those byte slices are a bit obscure.

PR #86, opened 2019-04-09
Make DecodeSignature public

flanfly:
DecodeAttestationData is already public. If a user wants to verify it
outside of a TPM they need to decode the TPMT_SIGNATURE structure too.

### DIFF
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -454,7 +454,8 @@ type Signature struct {
 	ECC *SignatureECC
 }
 
-func decodeSignature(in *bytes.Buffer) (*Signature, error) {
+// DecodeSignature decodes a serialized TPMT_SIGNATURE structure.
+func DecodeSignature(in *bytes.Buffer) (*Signature, error) {
 	var sig Signature
 	if err := tpmutil.UnpackBuf(in, &sig.Alg); err != nil {
 		return nil, fmt.Errorf("decoding Alg: %v", err)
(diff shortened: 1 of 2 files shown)


awly:
@flanfly signature returned by `Certify` http://go/gh/google/go-tpm/blob/master/tpm2/tpm2.go#L1267 is already parsed.
It can be used directly with e.g. `rsa.VerifyPKCS1v15`

flanfly:
Hey @awly, 
I want to verify the TPMS_ATTEST w/o the help of a TPM. My scenario is that I receive a pair of TPM2B_ATTEST/TPMT_SIGNATURE and need to verify the signature.

twitchyliquid64:
+1 - also need to do this.

@flanfly If you are willing to wait a month or so, we will be open-sourcing a verifier package in [go-attestation](https://github.com/google/go-attestation) capable of verifying quotes, AIK key properties, TPM2_Certify / TPM2_CertifyCreation outputs, etc (it works, just some final kinks to work out).

tpm2/structures.go
```diff
@@ -454,7 +454,7 @@ type Signature struct {
 	ECC *SignatureECC
 }
 
-func decodeSignature(in *bytes.Buffer) (*Signature, error) {
+func DecodeSignature(in *bytes.Buffer) (*Signature, error) {
```
awly: Add a doc comment please

flanfly: done

PR #89, opened 2019-04-10
Add golang.org/x/sys dependency to go.mod

awly:
It's used in TPM FD polling and was missed previously.
This should fix kokoro build too.


awly:
@tmroeder FYI, I'm vendoring golang.org/x/sys/unix, only because our kokoro environment doesn't support modules yet. Are you ok with that?

awly:
OK, since @tmroeder is OOO, i'll merge this. But we can revert later if needed.

tmroeder:
I find it unfortunate that we have to pull in code and add an extra vendor directory so that we can poll a file description in exactly one function in tpmutil. Can we use syscall.Epoll* from the standard library instead? The epoll syscall can perform exactly the function of waiting for an fd to be ready.

I glanced at #75, and I didn't see any discussion about this.

awly:
Good point, x/sys/unix doesn't do as much magic as I thought.
Sent https://github.com/google/go-tpm/pull/100

PR #92, opened 2019-04-18
Replace global prefix size with U32Bytes and U16Bytes

DenisKarch:
Replace global prefix size with U32Bytes and U16Bytes
Modify tpm1.2 and tpm2.0 commands to use these types instead of []byte
@twitchy-jsonp did most of the heavy lifting rewriting tpmutil/encoding.go to use reflect.Value instead of {}interface.

### DIFF
--- a/tpm/structures.go
+++ b/tpm/structures.go
@@ -17,7 +17,6 @@ package tpm
 import (
 	"crypto"
 	"crypto/rsa"
-	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
@@ -193,7 +192,7 @@ type keyParams struct {
 	AlgID     uint32
 	EncScheme uint16
 	SigScheme uint16
-	Params    []byte // Serialized rsaKeyParams or symmetricKeyParams.
+	Params    tpmutil.U32Bytes // Serialized rsaKeyParams or symmetricKeyParams.
 }
 
 // An rsaKeyParams encodes the length of the RSA prime in bits, the number of
@@ -202,13 +201,13 @@ type keyParams struct {
 type rsaKeyParams struct {
 	KeyLength uint32
 	NumPrimes uint32
-	Exponent  []byte
+	Exponent  tpmutil.U32Bytes
 }
 
 type symmetricKeyParams struct {
 	KeyLength uint32
 	BlockSize uint32
-	IV        []byte
+	IV        tpmutil.U32Bytes
 }
 
 // A key is a TPM representation of a key.
@@ -218,9 +217,9 @@ type key struct {
 	KeyFlags        uint32
 	AuthDataUsage   byte
 	AlgorithmParams keyParams
-	PCRInfo         []byte
-	PubKey          []byte
-	EncData         []byte
+	PCRInfo         tpmutil.U32Bytes
+	PubKey          tpmutil.U32Bytes
+	EncData         tpmutil.U32Bytes
 }
 
 // A key12 is a newer TPM representation of a key.
@@ -231,73 +230,29 @@ type key12 struct {
 	KeyFlags        uint32
 	AuthDataUsage   byte
 	AlgorithmParams keyParams
-	PCRInfo         []byte // This must be a serialization of a pcrInfoLong.
-	PubKey          []byte
-	EncData         []byte
+	PCRInfo         tpmutil.U32Bytes // This must be a serialization of a pcrInfoLong.
+	PubKey          tpmutil.U32Bytes
+	EncData         tpmutil.U32Bytes
 }
 
 // A pubKey represents a public key known to the TPM.
 type pubKey struct {
 	AlgorithmParams keyParams
-	Key             []byte
+	Key             tpmutil.U32Bytes
 }
 
 // A symKey is a TPM representation of a symmetric key.
 type symKey struct {
 	AlgID     uint32
 	EncScheme uint16
-	Key       []byte
-}
-
-// TPMMarshal will marshal the symKey into a byte Buffer as specified the TPM spec.
-func (k *symKey) TPMMarshal(out io.Writer) error {
-	if err := binary.Write(out, binary.BigEndian, &k.AlgID); err != nil {
-		return err
-	}
-	if err := binary.Write(out, binary.BigEndian, &k.EncScheme); err != nil {
-		return err
-	}
-	if err := binary.Write(out, binary.BigEndian, uint16(len(k.Key))); err != nil {
-		return err
-	}
-	if err := binary.Write(out, binary.BigEndian, k.Key); err != nil {
-		return err
-	}
-	return nil
-}
-
-// TPMUnmarshal will parse the byte Buffer and populate the symKey.
-func (k *symKey) TPMUnmarshal(in io.Reader) error {
-	if err := binary.Read(in, binary.BigEndian, &k.AlgID); err != nil {
-		return err
-	}
-	if err := binary.Read(in, binary.BigEndian, &k.EncScheme); err != nil {
-		return err
-	}
-	var size uint16
-	if err := binary.Read(in, binary.BigEndian, &size); err != nil {
-		return err
-	}
-	key := make([]byte, size)
-	if err := binary.Read(in, binary.BigEndian, &key); err != nil {
-		return err
-	}
-	k.Key = key
-	return nil
-}
-
-// TPMPackedSize will return the expected size in bytes of the marshaled symKey.
-func (k *symKey) TPMPackedSize() int {
-	// AlgID  + EncScheme + len(Key) + Key
-	// uint32 + uint16    + uint16   + BYTE[]
-	return binary.Size(k.AlgID) + binary.Size(k.EncScheme) + binary.Size(uint16(len(k.Key))) + binary.Size(k.Key)
+	Key       tpmutil.U16Bytes
 }
 
 // A tpmStoredData holds sealed data from the TPM.
 type tpmStoredData struct {
 	Version uint32
-	Info    []byte
-	Enc     []byte
+	Info    tpmutil.U32Bytes
+	Enc     tpmutil.U32Bytes
 }
 
 // String returns a string representation of a tpmStoredData.
@@ -323,7 +278,7 @@ type quoteInfo struct {
 // A pcrComposite stores a selection of PCRs with the selected PCR values.
 type pcrComposite struct {
 	Selection pcrSelection
-	Values    []byte
+	Values    tpmutil.U32Bytes
 }
 
 // convertPubKey converts a public key into TPM form. Currently, this function
--- a/tpm/structures.go
+++ b/tpm/structures.go
@@ -245,7 +245,7 @@ type pubKey struct {
 type symKey struct {
 	AlgID     uint32
 	EncScheme uint16
-	Key       tpmutil.U16Bytes
+	Key       tpmutil.U16Bytes // TPM_SYMMETRIC_KEY uses a 16-bit header for Key data
 }
 
 // A tpmStoredData holds sealed data from the TPM.
--- a/tpmutil/encoding.go
+++ b/tpmutil/encoding.go
@@ -78,6 +78,7 @@ func tryMarshal(buf io.Writer, v reflect.Value) (bool, error) {
 	// pointer with which to implement the interface.
 	// If the pointer of the type implements the interface, we should be
 	// able to construct a value to call TPMMarshal() with.
+	// TODO(awly): Try and avoid blowing away private data by using Addr() instead of Set()
 	if reflect.PtrTo(t).Implements(selfMarshalerType) {
 		tmp := reflect.New(t)
 		tmp.Elem().Set(v)
@@ -137,18 +138,11 @@ func tryUnmarshal(buf io.Reader, v reflect.Value) (bool, error) {
 		return true, v.Interface().(SelfMarshaler).TPMUnmarshal(buf)
 	}
 
-	// We might have a non-pointer struct field, which is settable,
-	// but we dont have a pointer with which to implement the interface.
+	// We might have a non-pointer struct field, which is addressable,
 	// If the pointer of the type implements the interface, and the
-	// value is settable, we should be able to construct a value to call
-	// TPMUnmarshal() with before replacing the value.
-	if v.CanSet() && reflect.PtrTo(t).Implements(selfMarshalerType) {
-		tmp := reflect.New(t)
-		if err := tmp.Interface().(SelfMarshaler).TPMUnmarshal(buf); err != nil {
-			return true, err
-		}
-		v.Set(tmp.Elem())
-		return true, nil
+	// value is addressable, we should be able to call TPMUnmarshal().
+	if v.CanAddr() && reflect.PtrTo(t).Implements(selfMarshalerType) {
+		return true, v.Addr().Interface().(SelfMarshaler).TPMUnmarshal(buf)
 	}
 
 	return false, nil
--- a/tpmutil/encoding.go
+++ b/tpmutil/encoding.go
@@ -23,78 +23,10 @@ import (
 	"reflect"
 )
 
-// lengthPrefixSize is the size in bytes of length prefix for byte slices.
-//
-// In TPM 1.2 this is 4 bytes.
-// In TPM 2.0 this is 2 bytes.
-var lengthPrefixSize int
-
-const (
-	tpm12PrefixSize = 4
-	tpm20PrefixSize = 2
+var (
+	selfMarshalerType = reflect.TypeOf((*SelfMarshaler)(nil)).Elem()
+	handlesAreaType   = reflect.TypeOf((*[]Handle)(nil))
 )
-
-// UseTPM12LengthPrefixSize makes Pack/Unpack use TPM 1.2 encoding for byte
-// arrays.
-func UseTPM12LengthPrefixSize() {
-	lengthPrefixSize = tpm12PrefixSize
-}
-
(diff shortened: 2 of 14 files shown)


twitchy-jsonp:
These changes have been tested using a real TPM for `tpm2`, and tested as well as we can with a simulator on `tpm`. For tests that were already broken with `tpm`, we were able to verify that the same bytes were being emitted by go-tpm.

tpm/structures.go
```diff
@@ -231,73 +230,29 @@ type key12 struct {
...
-	k.Key = key
-	return nil
-}
-
-// TPMPackedSize will return the expected size in bytes of the marshaled symKey.
-func (k *symKey) TPMPackedSize() int {
-	// AlgID  + EncScheme + len(Key) + Key
-	// uint32 + uint16    + uint16   + BYTE[]
-	return binary.Size(k.AlgID) + binary.Size(k.EncScheme) + binary.Size(uint16(len(k.Key))) + binary.Size(k.Key)
+	Key       tpmutil.U16Bytes
```
awly: Add a comment to clarify why this field is special

tpmutil/encoding.go
```diff
@@ -23,93 +23,28 @@ import (
...
 // length.
 func packWithHeader(ch commandHeader, cmd ...interface{}) ([]byte, error) {
 	hdrSize := binary.Size(ch)
-	bodySize, err := packedSize(cmd...)
+	body, err := Pack(cmd...)
 	if err != nil {
-		return nil, fmt.Errorf("couldn't compute packed size for message body: %v", err)
+		return nil, fmt.Errorf("couldn't pack message body: %v", err)
 	}
+	bodySize := binary.Size(body)
```
awly: isn't this just len(body)?

tpmutil/encoding.go
```diff
@@ -120,81 +55,96 @@ func packWithHeader(ch commandHeader, cmd ...interface{}) ([]byte, error) {
...
-		marshaler, ok := e.(SelfMarshaler)
-		if ok {
-			if err := marshaler.TPMMarshal(buf); err != nil {
-				return err
-			}
-			continue
+// tryMarshal attempts to use a Marshal() method defined on the type
+// to pack v into buf. True is returned if the method exists and the
+// marshal was attempted.
+func tryMarshal(buf io.Writer, v reflect.Value) (bool, error) {
```
awly: Also check that v isn't a nil pointer that implements SelfMarshaler

tpmutil/encoding.go
```diff
@@ -120,81 +55,96 @@ func packWithHeader(ch commandHeader, cmd ...interface{}) ([]byte, error) {
...
+	if t.Implements(selfMarshalerType) {
+		return true, v.Interface().(SelfMarshaler).TPMUnmarshal(buf)
+	}
+
+	// We might have a non-pointer struct field, which is settable,
+	// but we dont have a pointer with which to implement the interface.
+	// If the pointer of the type implements the interface, and the
+	// value is settable, we should be able to construct a value to call
+	// TPMUnmarshal() with before replacing the value.
+	if v.CanSet() && reflect.PtrTo(t).Implements(selfMarshalerType) {
```
awly: Use v.Addr instead, in case unmarshaler depends on some state in v.
Something like:
```
t := v.Type()
if reflect.PtrTo(t).Implements(selfMarshalerType) && v.CanAddr() {
  v = v.Addr()
  t = v.Type()
}

if t.Implements(selfMarshalerType) {
  return true, v.Interface().(SelfMarshaler).TPMUnmarshal(buf)
}
return false, nil
```

tpmutil/encoding.go
```diff
@@ -120,81 +55,96 @@ func packWithHeader(ch commandHeader, cmd ...interface{}) ([]byte, error) {
...
 	}
 
 	return nil
 }
 
+// tryUnmarshal attempts to use TPMUnmarshal() to perform the
+// unpack, if the given value implements SelfMarshaler.
+// True is returned if v implements SelfMarshaler & TPMUnmarshal
+// was called, along with an error returned from TPMUnmarshal.
+func tryUnmarshal(buf io.Reader, v reflect.Value) (bool, error) {
```
awly: Check for v being a nil pointer here too

tpmutil/encoding.go
```diff
@@ -120,81 +55,96 @@ func packWithHeader(ch commandHeader, cmd ...interface{}) ([]byte, error) {
...
+	t := v.Type()
+	if t.Implements(selfMarshalerType) {
+		return true, v.Interface().(SelfMarshaler).TPMMarshal(buf)
+	}
+
+	// We might have a non-pointer struct field, but we dont have a
+	// pointer with which to implement the interface.
+	// If the pointer of the type implements the interface, we should be
+	// able to construct a value to call TPMUnmarshal() with.
+	if reflect.PtrTo(t).Implements(selfMarshalerType) {
```
awly: Use `v.CanAddr()` and `v.Addr()` to call the pointer method

tpmutil/encoding.go
```diff
@@ -204,6 +154,58 @@ func Unpack(b []byte, elts ...interface{}) (int, error) {
 	return read, err
 }
 
+func unpackValue(buf io.Reader, v reflect.Value) error {
+	if didUnmarshal, err := tryUnmarshal(buf, v); didUnmarshal {
+		return err
+	}
+
+	if v.Type() == handlesAreaType {
```
awly: Pull this out into an unexported type that implements SelfMarshaler

tpmutil/encoding.go
```diff
@@ -120,81 +55,96 @@ func packWithHeader(ch commandHeader, cmd ...interface{}) ([]byte, error) {
...
-// using encoding/binary for everything else.
-func packType(buf io.Writer, elts ...interface{}) error {
-	for _, e := range elts {
-		marshaler, ok := e.(SelfMarshaler)
-		if ok {
-			if err := marshaler.TPMMarshal(buf); err != nil {
-				return err
-			}
-			continue
+// tryMarshal attempts to use a Marshal() method defined on the type
```
awly: s/Marshal/TPMMarshal/

tpmutil/encoding.go
```diff
@@ -120,81 +55,96 @@ func packWithHeader(ch commandHeader, cmd ...interface{}) ([]byte, error) {
...
+func tryMarshal(buf io.Writer, v reflect.Value) (bool, error) {
+	t := v.Type()
+	if t.Implements(selfMarshalerType) {
+		return true, v.Interface().(SelfMarshaler).TPMMarshal(buf)
+	}
+
+	// We might have a non-pointer struct field, but we dont have a
+	// pointer with which to implement the interface.
+	// If the pointer of the type implements the interface, we should be
+	// able to construct a value to call TPMUnmarshal() with.
```
awly: s/TPMUnmarshal/TPMMarshal/

tpmutil/encoding.go
```diff
@@ -204,6 +154,58 @@ func Unpack(b []byte, elts ...interface{}) (int, error) {
...
+		for i := 0; i < v.NumField(); i++ {
+			f := v.Field(i)
+			if err := unpackValue(buf, f); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+
+	// binary.Read can only set pointer values, so we need to take the address.
```
awly: move this into a default switch case above

tpmutil/encoding.go
```diff
@@ -204,6 +154,58 @@ func Unpack(b []byte, elts ...interface{}) (int, error) {
...
+			*handlesArea = append(*handlesArea, make([]Handle, int(numHandles)-areaLen)...)
+		}
+
+		return binary.Read(buf, binary.BigEndian, handlesArea)
+	}
+
+	switch v.Kind() {
+	case reflect.Ptr:
+		if v.IsNil() {
+			return fmt.Errorf("cannot pack nil %s", v.Type().String())
```
awly: s/pack/unpack/

tpmutil/encoding_test.go
```diff
@@ -247,9 +216,69 @@ func TestEncodingInvalidUnpack(t *testing.T) {
 
 }
 
+func TestSelfMarshaler(t *testing.T) {
+	var b U32Bytes
+	// The slice ui represents uint32(0), which is the length of an empty byte array.
+	ui := []byte{0, 0, 0, 0}
+	uiBuf := bytes.NewBuffer(ui)
+	if err := UnpackBuf(uiBuf, &b); err != nil {
+		t.Fatal("UnpackBuf failed to unpack the empty byte array")
```
awly: add err to the message, here and everywhere below

tpmutil/encoding_test.go
```diff
@@ -247,9 +216,69 @@ func TestEncodingInvalidUnpack(t *testing.T) {
...
+	}
+	pb, err := Pack(&b)
+	if err != nil {
+		t.Fatalf("Pack failed on U32Bytes")
+	}
+	if !bytes.Equal(pb, ui) {
+		t.Fatal("Pack failed to pack U32Bytes properly:", pb)
```
awly: Add expected `ui` value to the message

tpmutil/encoding_test.go
```diff
@@ -247,9 +216,69 @@ func TestEncodingInvalidUnpack(t *testing.T) {
 
 }
 
+func TestSelfMarshaler(t *testing.T) {
```
awly: Make this a table-driven test with subtests

tpmutil/structures.go
```diff
@@ -14,13 +14,84 @@
...
+// U16Bytes is a byte slice with a 16-bit header
+type U16Bytes []byte
+
+// TPMMarshal packs U16Bytes
+func (b *U16Bytes) TPMMarshal(out io.Writer) error {
+	size := uint16(len([]byte(*b)))
+	if err := binary.Write(out, binary.BigEndian, size); err != nil {
+		return err
+	}
+	if err := binary.Write(out, binary.BigEndian, []byte(*b)); err != nil {
```
awly: `return binary.Write(out, binary.BigEndian, []byte(*b))`
similar for all methods below

Also, is `[]byte` conversion here necessary?

tpmutil/structures.go
```diff
@@ -14,13 +14,84 @@
...
+		return err
+	}
+	size := int(tmpSize)
+	if len([]byte(*b)) >= size {
+		*b = (*b)[:size]
+	} else {
+		*b = append(*b, make([]byte, size-len(*b))...)
+	}
+
+	if err := binary.Read(in, binary.BigEndian, b); err != nil {
```
awly: can this be `in.Read([]byte(*b))` ?

tpm/structures.go
DenisKarch: Not sure what you mean by special, the TPM1.2 spec happens to use a 16-bit length for that byte array (I don't think there's any particular reason why).

tpmutil/structures.go
DenisKarch: it could be `if _, err := in.Read([]byte(*b)); err != nil' yeah

awly: Right, just so that we don't incur the type switching overhead of binary.Read

tpmutil/encoding.go
DenisKarch: just to be consistent with the `binary.Size(ch) above` I can change it to len if you'd like.

tpmutil/encoding.go
DenisKarch: Addressable is a subset of Settable, Using .Addr() causes a lot of the tests to break.

tpmutil/structures.go
DenisKarch: Could also switch to using out.Write above.

tpmutil/encoding.go
twitchyliquid64: No, that wont work if v is not addressable. This is checking if a pointer to the type of `v` would implement the interface, so we can construct an equivalent addressable type and then overwrite the reflect.Value.

This is needed in particular for literals with non-pointer fields that implement the interface.

awly: I know, but if v is not addressable, can it be settable?
Here's where this fails:

```
type Foo struct {
  BeFancy bool
  Val string
}

func (f *Foo) TPMUnmarshal(in bytes.Buffer) error {
  // read size header
  // read bytes based on size header into val
  if f.BeFancy {
    val = fmt.Sprintf("fancy(%s)", val)
  }
  return nil
}
```

With your current implementation, `BeFancy` will always be false.
But what if caller set it to true, expecting special unmarshaling?

tpm/structures.go
awly: That's exactly what the comment should say.
Otherwise the next person reading this code will think "hmm, all the other byte slices here use U32Bytes, this one is U16Bytes. must be a bug, let me fix that..."

tpmutil/encoding.go
awly: Yeah, no need to jump through the type check in binary.Size, use len

tpmutil/encoding.go
```diff
@@ -120,81 +55,105 @@ func packWithHeader(ch commandHeader, cmd ...interface{}) ([]byte, error) {
...
-			}
-			continue
+// tryMarshal attempts to use a TPMMarshal() method defined on the type
+// to pack v into buf. True is returned if the method exists and the
+// marshal was attempted.
+func tryMarshal(buf io.Writer, v reflect.Value) (bool, error) {
+	t := v.Type()
+	if t.Implements(selfMarshalerType) {
+		if v.Kind() == reflect.Ptr && v.IsNil() {
+			return true, fmt.Errorf("cannot call TPMMarshal on a nil pointer")
```
awly: add type name for t at least, so that it's easier to debug

tpmutil/encoding.go
twitchyliquid64: Yes I think you can have settable but not addressable. I originally tried that and many different permutations of what you are describing - they did not work in all cases. I created the logic to generate a surrogate type and assign it as a last resort.

I think the example you gave should work fine because of `tmp.Elem().Set(v)`.

If you passed that by pointer, it would work without hitting the logic of generating a new type. If it was passed by value, it would not work regardless.

tpmutil/encoding.go
awly: See my example in the other comment thread: `TPMUnmarshal` may depend on v's current state.
Why are the tests failing? I think all the values passed in should be addressable?

tpmutil/encoding.go
```diff
@@ -212,95 +202,16 @@ func Unpack(b []byte, elts ...interface{}) (int, error) {
...
-		k := v.Kind()
-		if k != reflect.Ptr {
-			return fmt.Errorf("all values passed to Unpack must be pointers, got %v", k)
+		if v.Kind() != reflect.Ptr {
+			return fmt.Errorf("non-pointer value %q passed to UnpackBuf", v.Type().String())
 		}
-
 		if v.IsNil() {
-			return errors.New("can't fill a nil pointer")
+			return errors.New("nill pointer passed to UnpackBuf")
```
awly: s/nill/nil/

tpmutil/encoding_test.go
```diff
@@ -247,9 +216,36 @@ func TestEncodingInvalidUnpack(t *testing.T) {
...
+	}{
+		{[]byte{0, 0}, &empty16},
+		{[]byte{0, 1, 137}, &empty16},
+		{[]byte{0, 0, 0, 0}, &empty32},
+		{[]byte{0, 0, 0, 1, 137}, &empty32},
+	}
+	for _, st := range subTests {
+		buffer := bytes.NewBuffer(st.encoded)
+		if err := UnpackBuf(buffer, st.decoded); err != nil {
+			t.Fatal("UnpackBuf failed:", err)
```
awly: Use t.Fatalf instead

tpmutil/encoding_test.go
```diff
@@ -247,9 +216,36 @@ func TestEncodingInvalidUnpack(t *testing.T) {
...
+		}
+		packed, err := Pack(st.decoded)
+		if err != nil {
+			t.Fatal("Pack failed:", err)
+		}
+		if !bytes.Equal(packed, st.encoded) {
+			t.Fatalf("Pack failed: expected %#v, got %#v", st.encoded, packed)
```
awly: always put the `got` value before `want` in test messages

tpmutil/encoding_test.go
```diff
@@ -247,9 +216,36 @@ func TestEncodingInvalidUnpack(t *testing.T) {
...
+	subTests := []struct {
+		encoded []byte
+		decoded interface{}
+	}{
+		{[]byte{0, 0}, &empty16},
+		{[]byte{0, 1, 137}, &empty16},
+		{[]byte{0, 0, 0, 0}, &empty32},
+		{[]byte{0, 0, 0, 1, 137}, &empty32},
+	}
+	for _, st := range subTests {
```
awly: either use subtests (`t.Run`) or print input for each call that fails below.
otherwise it's really hard to figure out which input failed

tpmutil/structures.go
```diff
@@ -51,10 +141,31 @@ type responseHeader struct {
 // A Handle is a reference to a TPM object.
 type Handle uint32
 
+type handleList []Handle
+
+func (l *handleList) TPMMarshal(out io.Writer) error {
+	return fmt.Errorf("TPM should never marshal []Handle")
```
awly: make this "TPMMarhsal on []Handle is not supported yet".
no reason we couldn't implement this later

DenisKarch: Fair, it's just afaik none of the TPM commands require []Handle, hence why I had that comment.
Fixed.

tpmutil/encoding_test.go
DenisKarch: Added Logf

PR #94, opened 2019-04-22
Implement example to generate a fake EK certificate.

twitchy-jsonp:
This is part 1/2 of implementing the tooling necessary for TPM 1.2 continuous testing.

The second part will create two more examples, which will define the NV space & write the certificate.

### DIFF
--- a/examples/tpm-fakeekcert/tpm-fakeekcert.go
+++ b/examples/tpm-fakeekcert/tpm-fakeekcert.go
@@ -1,0 +1,132 @@
+// Copyright (c) 2014, Google LLC All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"crypto/rand"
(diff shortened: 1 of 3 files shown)


examples/tpm-fakeekcert/tpm-fakeekcert.go
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) 2014, Google LLC All rights reserved.
```
tmroeder: Copyright 2019

examples/tpm-fakeekcert/tpm-fakeekcert.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/google/go-tpm/tpm"
+)
+
+var (
+	ownerAuthEnvVar = "TPM_OWNER_AUTH"
+
+	tpmPath  = flag.String("tpm", "/dev/tpm0", "The path to the TPM device to use")
+	certPath = flag.String("cert", "ek.der", "The path to write the EK out to")
+	certOrg  = flag.String("cert_org", "Acme Co", "The organization string to use in the EKCert")
```
tmroeder: Do we have a standard name that we can use here? I don't know if there's a Google standard for this.

Despite "Acme" being the name of a classic fictional company, it is also the name of real companies, which makes me a little worried about using it here.

twitchyliquid64: The example Go code that generates this uses `Acme Co`. I can change this to `Fake EK` ?

tmroeder: That sounds good, sure.

twitchy-jsonp: The example Go code that generates this uses `Acme Co`. I can change this to `Fake EK` ?

twitchy-jsonp:
(still working on this - i'll let you know when its ready to review)

PR #95, opened 2019-04-24
Add Public function for constructing TPM_QUOTE_INFO

DenisKarch:


tpm/verify.go
```diff
@@ -105,6 +105,16 @@ func (pk *pubKey) unmarshalRSAPublicKey() (*rsa.PublicKey, error) {
...
+	qi, err := newQuoteInfo(data, pcrNums, pcrs)
+	if err != nil {
+		return nil, err
+	}
+	return tpmutil.Pack(qi)
+}
+
 // newQuoteInfo computes a quoteInfo structure for a given pair of data and PCR
 // values.
 func newQuoteInfo(data []byte, pcrNums []int, pcrs []byte) (*quoteInfo, error) {
```
awly: Make newQuoteInfo call tpmutil.Pack and expose that instead (maybe called `EncodeQuoteInfo`)
No need for a separate func.

PR #98, opened 2019-04-30
tpm2: make NVRead/Write tests more reliable

awly:
Remove standalone NVRead tests, because some TPMs may not have the EK
certificate at the NV index it's reading from.
NVReadWrite test already covers this func anyway.

In NVReadWrite, undefine the index this test uses first, so that past
failures don't interfere.

### DIFF
--- a/tpm2/tpm2_test.go
+++ b/tpm2/tpm2_test.go
@@ -999,20 +999,6 @@ func TestCreateAndCertifyCreation(t *testing.T) {
 	}
 }
 
-func TestNVRead(t *testing.T) {
-	rw := openTPM(t)
-	defer rw.Close()
-
-	// Read the NVCert, which should be present on any real TPM.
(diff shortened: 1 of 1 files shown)


tpm2/tpm2_test.go
```diff
@@ -1023,6 +1009,11 @@ func TestNVReadWrite(t *testing.T) {
 		attr                = AttrOwnerWrite | AttrOwnerRead
 	)
 
+	// Undefine the space, just in case previous test failed to.
```
tmroeder: I'm not sure I understand the comment. After removing the TestNVRead case, the previous test is TestCreateAndCertifyCreation, and I don't see an explicit NVUndefineSpace there. Please clarify.

awly: I meant "the previous run of this same test". Clarified in the comment.

PR #100, opened 2019-04-30
Call poll syscall directly and drop dependency on golang.org/x/sys

awly:
This version of poll also checks the syscall result and supports non-infinite timeouts.


tpmutil/poll_unix.go
```diff
@@ -3,20 +3,54 @@
...
 	)
-	pollFds := []unix.PollFd{
-		{Fd: int32(f.Fd()), Events: events},
+	// Convert timeout into milliseconds int (or keep -1 for no timeout).
+	if timeout == pollNoTimeout {
+		timeoutMS = pollNoTimeout
+	} else {
+		timeoutMS = int(timeout / time.Millisecond)
+	}
+	_, _, errno := syscall.Syscall(syscall.SYS_POLL, uintptr(unsafe.Pointer(fd)), uintptr(numFD), uintptr(timeoutMS))
```
tmroeder: Thanks for doing this!

I'd like us to use syscall.Epoll instead, since that will allow us to avoid using the unsafe library, at least in our code (I know the syscall library uses it to implement these epoll methods).

It looks like the current implementation needs unsafe because it's passing a pollFD structure directly to the syscall. Instead, how about using the epoll interface provided by syscall:

```go
fd := int32(f.Fd())

epfd, err := sycall.EpollCreate(numFD)
if err != nil {
  return err
}

event := &syscall.EpollEvent { events: syscall.EPOLLIN, Fd: fd }
if err := syscall.EpollCtl(epfd, syscall.EPOLL_CTL_ADD, fd, &event); err != nil {
  return err
}

out := make([]syscall.EpollEvent, numFD)
n, err := syscall.EpollWait(epfd, out, timeout)
if err != nil {
  return err
}
if n != numFD || out[0].Fd != fd {
  return fmt.Errorf("epoll succeeded without returning the correct file descriptor")
}

return nil
```

awly: That's what I tried first.
However, I'm getting `operation not permitted` error from `syscall.EpollCtl` with vTPM in GCE and can't figure out why.
Quick search suggests it might be because epoll doesn't expect a regular file or character device.
Do you have any ideas?

tmroeder: epoll and its ilk are not valid for regular files because regular files are always ready for reading and writing etc. But a /dev/tpm0 is not.

Does it work on a real TPM device? I would check strace to see if Go is doing something different than epoll, though the Go syscall source seems to show that it's just directly passing it to a syscall.

I'm OK with the version you have, since it's a very simple use of unsafe to pass a struct to a syscall.

awly: strace looks fine:

```
newfstatat(AT_FDCWD, "/dev/tpm0", {st_mode=S_IFCHR|0600, st_rdev=makedev(10, 224), ...}, 0) = 0
openat(AT_FDCWD, "/dev/tpm0", O_RDWR|O_CLOEXEC) = 3
epoll_ctl(4, EPOLL_CTL_ADD, 3, {EPOLLIN|EPOLLOUT|EPOLLRDHUP|EPOLLET, {u32=550653824, u64=140549060448128}}) = -1 EPERM (Operation not permitted)
epoll_ctl(4, EPOLL_CTL_DEL, 3, 0xc0000459b4) = -1 EPERM (Operation not permitted)
write(3, "\200\2\0\0\0M\0\0\0011@\0\0\1\0\0\0\t@\0\0\t\0\0\1\0\0\0\10\0\4\1"..., 77) = 77
futex(0x944a78, FUTEX_WAKE_PRIVATE, 1)  = 1
futex(0x944978, FUTEX_WAKE_PRIVATE, 1)  = 1
epoll_create(1)                         = 28
epoll_ctl(28, EPOLL_CTL_ADD, 3, {EPOLLIN, {u32=3, u64=3}}) = -1 EPERM (Operation not permitted)
close(3)                                = 0
write(1, "--- FAIL: TestEncryptDecrypt (0."..., 37--- FAIL: TestEncryptDecrypt (0.01s)
) = 37
futex(0x944a78, FUTEX_WAKE_PRIVATE, 1)  = 1
futex(0x944978, FUTEX_WAKE_PRIVATE, 1)  = 1
write(1, "    tpm2_test.go:1235: CreatePri"..., 79    tpm2_test.go:1235: CreatePrimary failed: EpollCtl: operation not permitted
) = 79
futex(0x944a78, FUTEX_WAKE_PRIVATE, 1)  = 1
futex(0x944978, FUTEX_WAKE_PRIVATE, 1)  = 1
```

It looks like `os.OpenFile` also calls `epoll_ctl` and fails.
This is run under sudo.

I'll ask some people that have a better understanding of linux syscalls, but seems like we're locked into `poll`.

To clarify - the only reason you prefer `epoll` is to avoid importing `unsafe`? Or is there some behavior difference between the two?

tmroeder: Yes, I'd like to avoid importing unsafe if possible.

awly: Consulted with some folks, `epoll` won't work on `/dev/tpm0`.
`epoll` support is up to the specific character device to implement and apparently this one doesn't.

Looks like we're stuck with `poll`.
Are you ok with that?

tmroeder: yes

PR #103, opened 2019-05-07
CapabilityPCRs

ChriMarMe:
Implements GetCapability-Case for CapabilityPCRs. Extends decodeTPMLPCRSelection for multiple PCRSelections.

### DIFF
--- a/tpm2/encoding_test.go
+++ b/tpm2/encoding_test.go
@@ -37,7 +37,7 @@ func TestDecodeReadPCRs(t *testing.T) {
 	}
 }
 
-func TestEncodeDecodeTPMLSelection(t *testing.T) {
+func TestSingleEncodeDecodeTPMLSelection(t *testing.T) {
 	buf, err := encodeTPMLPCRSelection(pcrSelection)
 	if err != nil {
 		t.Fatal(err)
@@ -49,6 +49,23 @@ func TestEncodeDecodeTPMLSelection(t *testing.T) {
 
 	if !reflect.DeepEqual(got, pcrSelection) {
 		t.Errorf("after decoding: %+v, before encoding: %+v", got, pcrSelection)
+	}
+}
+
+func TestMultiArgsEncodeDecodeTPMLSelection(t *testing.T) {
+	var multipcrselection []PCRSelection
+	multipcrselection = append(multipcrselection, pcrSelection)
+	buf, err := encodeTPMLPCRSelection(multipcrselection...)
+	if err != nil {
+		t.Fatal(err)
+	}
+	got, err := decodeTPMLPCRSelection(bytes.NewBuffer(buf))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !reflect.DeepEqual(got, multipcrselection) {
+		t.Errorf("after decoding: %+v, before encoding %+v", got, multipcrselection)
 	}
 }
 
--- a/tpm2/encoding_test.go
+++ b/tpm2/encoding_test.go
@@ -38,7 +38,7 @@ func TestDecodeReadPCRs(t *testing.T) {
 }
 
 func TestSingleEncodeDecodeTPMLSelection(t *testing.T) {
-	buf, err := encodeTPMLPCRSelection(pcrSelection)
+	buf, err := encodeTPMLPCRSelection(pcrSelection7)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -47,14 +47,13 @@ func TestSingleEncodeDecodeTPMLSelection(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if !reflect.DeepEqual(got, pcrSelection) {
-		t.Errorf("after decoding: %+v, before encoding: %+v", got, pcrSelection)
+	if !reflect.DeepEqual(got, pcrSelection7) {
+		t.Errorf("after decoding: %+v, before encoding: %+v", got, pcrSelection7)
 	}
 }
 
 func TestMultiArgsEncodeDecodeTPMLSelection(t *testing.T) {
-	var multipcrselection []PCRSelection
-	multipcrselection = append(multipcrselection, pcrSelection)
+	multipcrselection := []PCRSelection{pcrSelection0, pcrSelection1, pcrSelection7}
 	buf, err := encodeTPMLPCRSelection(multipcrselection...)
 	if err != nil {
 		t.Fatal(err)
@@ -147,7 +146,7 @@ func TestEncodeCreate(t *testing.T) {
 			Modulus:  big.NewInt(0),
 		},
 	}
-	cmdBytes, err := encodeCreate(HandleOwner, pcrSelection, "", defaultPassword, []byte{255} /*sensitiveData*/, params)
+	cmdBytes, err := encodeCreate(HandleOwner, pcrSelection7, "", defaultPassword, []byte{255} /*sensitiveData*/, params)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -181,7 +180,7 @@ func TestEncodePolicyPCR(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	cmdBytes, err := encodePolicyPCR(tpmutil.Handle(0x03000000), []byte(nil), pcrSelection)
+	cmdBytes, err := encodePolicyPCR(tpmutil.Handle(0x03000000), []byte(nil), pcrSelection7)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -256,7 +255,7 @@ func TestEncodeQuote(t *testing.T) {
 		t.Fatal(err)
 	}
 	toQuote := []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 0xa, 0xb, 0xc, 0xd, 0xe, 0xf, 0x10}
-	cmdBytes, err := encodeQuote(tpmutil.Handle(0x80000001), defaultPassword, "", toQuote, pcrSelection, 0x0010)
+	cmdBytes, err := encodeQuote(tpmutil.Handle(0x80000001), defaultPassword, "", toQuote, pcrSelection7, 0x0010)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -351,7 +350,7 @@ func TestEncodeSensitiveArea(t *testing.T) {
 }
 
 func TestEncodeTPMLPCRSelection(t *testing.T) {
-	s, err := encodeTPMLPCRSelection(pcrSelection)
+	s, err := encodeTPMLPCRSelection(pcrSelection7)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/tpm2/encoding_test.go
+++ b/tpm2/encoding_test.go
@@ -42,7 +42,7 @@ func TestEncodeDecodeTPMLSelection(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	got, err := decodeTPMLPCRSelection(bytes.NewBuffer(buf))
+	got, err := decodeOneTPMLPCRSelection(bytes.NewBuffer(buf))
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -688,7 +688,7 @@ func decodeQuoteInfo(in *bytes.Buffer) (*QuoteInfo, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding PCRSelection: %v", err)
 	}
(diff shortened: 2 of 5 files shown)


ChriMarMe:
I signed it !

tpm2/tpm2.go
```diff
@@ -79,41 +79,42 @@ func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
...
-				continue
+	for i:=0 ; i < int(count); i++ {
+		if err := tpmutil.UnpackBuf(buf, &ts.Hash, &ts.Size); err != nil {
+			return sel, err
+		}
+		ts.PCRs = make(tpmutil.RawBytes, ts.Size)
+		if _, err := buf.Read(ts.PCRs); err != nil {
+			return sel, err
+		}
+		sel = append(sel,PCRSelection{})
```
awly: Could you re-run gofmt/goimports on this file? I think there's a space missing in here

awly: Make a `var s PCRSelection`, fill it in, and append it to sel at the end of this loop body

tpm2/tpm2.go
```diff
@@ -79,41 +79,42 @@ func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
...
+		sel[i].Hash = ts.Hash
+		for j := 0; j < int(ts.Size); j++ {
+			for k := 0; k < 8; k++ {
+				set := ts.PCRs[j] & byte(1<<byte(j))
```
awly: I think this should be `1<<byte(k)`

tpm2/structures.go
```diff
@@ -688,7 +688,7 @@ func decodeQuoteInfo(in *bytes.Buffer) (*QuoteInfo, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding PCRSelection: %v", err)
 	}
-	out.PCRSelection = sel
+	out.PCRSelection = sel[0]
```
awly: Check that `len(sel) == 1` and error out if it's not.

tpm2/tpm2.go
```diff
@@ -242,6 +245,21 @@ func decodeGetCapability(in []byte) ([]interface{}, bool, error) {
 		}
 		return props, moreData > 0, nil
 
+
+	case CapabilityPCRs:
+		var pcrss []interface{}
+		//for i := 0; i < int(numPCRs); i++ {
```
awly: remove this

tpm2/structures.go
```diff
@@ -688,7 +688,12 @@ func decodeQuoteInfo(in *bytes.Buffer) (*QuoteInfo, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding PCRSelection: %v", err)
 	}
-	out.PCRSelection = sel
+	if len(sel) == 1 {
+		out.PCRSelection = sel[0]
+	} else {
+		return nil, fmt.Errorf("PCRSelection is empty")
```
awly: It could also have >1 PCR selection. How about something like:
```
if len(sel) != 1 {
  return nil, fmt.Errorf("got %d PCRSelections, expected 1", len(sel))
}
out.PCRSelection = sel[0]
```

tpm2/structures.go
```diff
@@ -759,7 +764,7 @@ func DecodeCreationData(buf []byte) (*CreationData, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding PCRSelection: %v", err)
 	}
-	out.PCRSelection = sel
+	out.PCRSelection = sel[0]
```
awly: also check len(sel) here as above

tpm2/tpm2.go
```diff
@@ -79,40 +79,42 @@ func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
...
-		return sel, err
-	}
-
-	sel.Hash = ts.Hash
-	for i := 0; i < int(ts.Size); i++ {
-		for j := 0; j < 8; j++ {
-			set := ts.PCRs[i] & byte(1<<byte(j))
-			if set == 0 {
-				continue
+	var s PCRSelection
```
awly: Move this inside the loop body, otherwise `s.PCRs` accumulates all PCRs from all selections, instead of just one.

tpm2/tpm2.go
```diff
@@ -124,11 +126,13 @@ func decodeReadPCRs(in []byte) (map[int][]byte, error) {
 		return nil, err
 	}
 
-	sel, err := decodeTPMLPCRSelection(buf)
+	sels, err := decodeTPMLPCRSelection(buf)
 	if err != nil {
 		return nil, fmt.Errorf("decoding TPML_PCR_SELECTION: %v", err)
 	}
 
+	sel := sels[0]
```
awly: Check len(sels) here as before.

Since this is needed multiple times, maybe add a helper func:
```
func decodeOneTPMLPCRSelection(buf *bytes.Buffer) (PCRSelection, error) {
  sels, err := decodeTPMLPCRSelection(buf)
  if err != nil {
    return PCRSelction{}, err
  }
  if len(sel) != 1 {
    return nil, fmt.Errorf("got %d TPMS_PCR_SELECTION items in TPML_PCR_SELECTION, expected 1", len(sel))
  }
  return sel[0], nil
}
```

awly:
Also make sure `go test` passes. Currently there's a failing test: https://source.cloud.google.com/results/invocations/f8261f83-189d-4d43-9f00-2f1329561c07/targets/go-tpm%2Fgoogle_presubmit/log
In case you don't have access to that link:

```
?   	github.com/google/go-tpm/examples/tpm-clear	[no test files]
?   	github.com/google/go-tpm/examples/tpm-genaik	[no test files]
?   	github.com/google/go-tpm/examples/tpm-keys	[no test files]
?   	github.com/google/go-tpm/examples/tpm-sign	[no test files]
?   	github.com/google/go-tpm/examples/tpm-takeownership	[no test files]
?   	github.com/google/go-tpm/examples/tpm2-ekcert	[no test files]
?   	github.com/google/go-tpm/examples/tpm2-nvread	[no test files]
?   	github.com/google/go-tpm/examples/tpm2-seal-unseal	[no test files]
ok  	github.com/google/go-tpm/tpm	0.013s
--- FAIL: TestEncodeDecodeTPMLSelection (0.00s)
    encoding_test.go:51: after decoding: [{Hash:4 PCRs:[7]}], before encoding: {Hash:4 PCRs:[7]}
FAIL
FAIL	github.com/google/go-tpm/tpm2	0.014s
ok  	github.com/google/go-tpm/tpm2/credactivation	0.005s
ok  	github.com/google/go-tpm/tpmutil	0.003s
?   	github.com/google/go-tpm/tpmutil/mssim	[no test files]
```

ChriMarMe:
Some problems pop up here. I see two possible approaches, but maybe I miss some.

Variadic encode-function: 
Argument of encode-function and return-value of decode-function differ. Which could be a problem of complexity in later use. User always needs to remember that encode- and decode-functions do not operate the same data-structures. Also test will continue to fail.

Encode-function with slice as argument: 
User needs to create a slice of PCRSelection and pass it to the function. So every time even if only one PCRSelection needs to be encoded, user have to build a slice of PCRSelection around it. This also is some kind of extra complexity for users. Test will pass in this solution (I guess)

I'm thankful for a word of advice.

awly:
Either way should work fine and won't be too hard for readers to understand.
I lean slightly towards variadic argument to encode, but up to you.

Tests must get fixed regardless.

tpm2/structures.go
```diff
@@ -688,7 +688,11 @@ func decodeQuoteInfo(in *bytes.Buffer) (*QuoteInfo, error) {
 	if err != nil {
 		return nil, fmt.Errorf("decoding PCRSelection: %v", err)
 	}
-	out.PCRSelection = sel
+	if len(sel) != 1 {
```
awly: Use `decodeOneTPMLPCRSelection` here instead

tpm2/structures.go
```diff
@@ -755,10 +759,7 @@ func DecodeCreationData(buf []byte) (*CreationData, error) {
 	in := bytes.NewBuffer(buf)
 	var out CreationData
 
-	sel, err := decodeTPMLPCRSelection(in)
-	if err != nil {
-		return nil, fmt.Errorf("decoding PCRSelection: %v", err)
-	}
+	sel, err := decodeOneTPMLPCRSelection(in)
```
awly: check the error

tpm2/tpm2.go
```diff
@@ -51,8 +51,12 @@ func FlushContext(rw io.ReadWriter, handle tpmutil.Handle) error {
 	return err
 }
 
-func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
-	if len(sel.PCRs) == 0 {
+func encodeTPMLPCRSelection(sel ...PCRSelection) ([]byte, error) {
+	var s []PCRSelection
```
awly: No need to make a copy of `sel`, it's already `[]PCRSelection`, use it directly

tpm2/tpm2.go
```diff
@@ -64,70 +68,89 @@ func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
...
-		Size: sizeOfPCRSelect,
-		PCRs: make(tpmutil.RawBytes, sizeOfPCRSelect),
-	}
-	// sel.PCRs parameter is indexes of PCRs, convert that to set bits.
-	for _, n := range sel.PCRs {
-		byteNum := n / 8
-		bytePos := byte(1 << byte(n%8))
-		ts.PCRs[byteNum] |= bytePos
+	var retBytes []byte
+	for i := 0; i < len(s); i++ {
```
awly: `for _, s := range sel {`

awly:
```
if len(sel.PCRs) == 0 {
  return tpmutil.Pack(uint32(0))
}
```

to preserve old behavior

tpm2/tpm2.go
```diff
@@ -64,70 +68,89 @@ func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
...
+			Size: sizeOfPCRSelect,
+			PCRs: make(tpmutil.RawBytes, sizeOfPCRSelect),
+		}
+		// s[i].PCRs parameter is indexes of PCRs, convert that to set bits.
+		for _, n := range s[i].PCRs {
+			byteNum := n / 8
+			bytePos := byte(1 << byte(n%8))
+			ts.PCRs[byteNum] |= bytePos
+		}
+		tmpBytes, err := tpmutil.Pack(uint32(1), ts)
```
awly: `TPML_PCR_Selection` encoding should be:
`[len(sel), sel[0], sel[1], ...]`
here you're producing:
`[1, sel[0], 1, sel[1], ...]`

(the hardcoded `uint32(1)` in the old implementation was `len(sel)` because I was too lazy to handle a list)

tpm2/encoding_test.go
```diff
@@ -52,6 +52,23 @@ func TestEncodeDecodeTPMLSelection(t *testing.T) {
 	}
 }
 
+func TestMultiArgsEncodeDecodeTPMLSelection(t *testing.T) {
+	var multipcrselection []PCRSelection
+	multipcrselection = append(multipcrselection, pcrSelection)
```
awly: Make sure there are multiple selections:
`selections := []PCRSelection{pcrSelection, pcrSelection, pcrSelection}`

tpm2/tpm2.go
```diff
@@ -64,70 +64,102 @@ func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
 	// For example, selecting PCRs 3 and 9 looks like:
 	// size(3)  mask     mask     mask
 	// 00000011 00000000 00000001 00000100
-	ts := tpmsPCRSelection{
-		Hash: sel.Hash,
-		Size: sizeOfPCRSelect,
-		PCRs: make(tpmutil.RawBytes, sizeOfPCRSelect),
+	var retBytes []byte
+	var tSize int
```
awly: `tSize` is just `len(sel)`
You can do:
```
retBytes, err := tpmutil.Pack(uint32(len(sel)))
```

tpm2/tpm2.go
```diff
@@ -64,70 +64,102 @@ func encodeTPMLPCRSelection(sel PCRSelection) ([]byte, error) {
...
 	var updateCounter uint32
 	if err := tpmutil.UnpackBuf(buf, &updateCounter); err != nil {
 		return nil, err
 	}
 
-	sel, err := decodeTPMLPCRSelection(buf)
-	if err != nil {
-		return nil, fmt.Errorf("decoding TPML_PCR_SELECTION: %v", err)
-	}
+	sel, err := decodeOneTPMLPCRSelection(buf)
```
awly: another missing error check

PR #50, opened 2018-11-27
Implement CreatePrimaryEx(), refactor CreatePrimary()

twitchy-jsonp:
This implements primitives necessary for remote key attestation for an AIK.

As discussed with awly@, most parameters returned from `CreatePrimaryEx()` are raw byte slices. This is easier to use for their intended use case (remote key attestation), users who just want a primary key have everything they need from `CreatePrimary()` already.

PR #104, opened 2019-05-16
Fix tpm2.CreatePrimaryRawTemplate return value

awly:
This was broken in https://github.com/google/go-tpm/pull/50 because we
started returning raw encoded Public data for the key instead of parsed
crypto.PublicKey.
This didn't get caught because we had no tests and because []byte
implements crypto.PublicKey (it's an empty interface).


awly:
cc @twitchy-jsonp @sambdavidson

sambdavidson:
Look great.
/lgtm

twitchyliquid64:
LGTM.

PR #105, opened 2019-05-18
Detect version of TPM in OpenTPM()

josephlr:
This PR improves the error messages when using the `tpm2` library with a TPM 1.2 or vice-versa.

Before, the user would get an opaque input/output error. Now they get
```
Error: open /dev/tpm0: device is not a TPM 2.0
```

On Windows, this detection is done using [Tbsi_GetDeviceInfo](https://docs.microsoft.com/en-us/windows/desktop/api/tbs/nf-tbs-tbsi_getdeviceinfo), which required adding a binding to `tpmutil/tbs`.

On Linux/Unix, this detection is done by trying to issue a call to `GetManufacturer()`. Per the TPM spec, this call succeeds even if the TPM is in failure mode. This required adding some `TPM_PT` constants and `GetManufacturer()` wrapper function to `tpm2`.

See the commit history if you want to see the individual parts of this change.

Note: This command removes `PropertyTag` in favor of `TPMProp`. While `PropertyTag` is not used anywhere, this is technically a breaking change.

### DIFF
--- a/tpm/open_other.go
+++ b/tpm/open_other.go
@@ -1,0 +1,42 @@
+// +build !windows
+
+// Copyright (c) 2019, Google LLC All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tpm
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/google/go-tpm/tpmutil"
+)
+
+// OpenTPM opens a channel to the TPM at the given path. If the file is a
+// device, then it treats it like a normal TPM device, and if the file is a
+// Unix domain socket, then it opens a connection to the socket.
+func OpenTPM(path string) (io.ReadWriteCloser, error) {
+	rwc, err := tpmutil.OpenTPM(path)
+	if err != nil {
+		return nil, err
+	}
+
+	// Make sure this is a TPM 1.2
+	_, err = GetManufacturer(rwc)
+	if err != nil {
+		rwc.Close()
+		return nil, fmt.Errorf("open %s: device is not a TPM 1.2", path)
+	}
+	return rwc, nil
+}
--- a/tpm/open_windows.go
+++ b/tpm/open_windows.go
@@ -1,0 +1,37 @@
+// Copyright (c) 2018, Google LLC All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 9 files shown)


tpm/open_other.go
```diff
@@ -0,0 +1,42 @@
...
+// device, then it treats it like a normal TPM device, and if the file is a
+// Unix domain socket, then it opens a connection to the socket.
+func OpenTPM(path string) (io.ReadWriteCloser, error) {
+	rwc, err := tpmutil.OpenTPM(path)
+	if err != nil {
+		return nil, err
+	}
+
+	// Make sure this is a TPM 1.2
+	_, err = GetManufacturer(rwc)
```
twitchy-jsonp: Will this work if the TPM is in startup as well? iirc a specific sequence of commands must be received at startup for initialization to succeed?

tpm/open_windows.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+	"io"
+
+	"github.com/google/go-tpm/tpmutil"
+	"github.com/google/go-tpm/tpmutil/tbs"
+)
+
+// OpenTPM opens a channel to the TPM.
+func OpenTPM() (io.ReadWriteCloser, error) {
+	info, err := tbs.GetDeviceInfo()
```
twitchy-jsonp: NOTE: I'm unsure as to how `TBS_GetDeviceInfo` behaves if Windows considers it not ready (ie: windows did not claim ownership)

tpm2/tpm2.go
```diff
@@ -309,6 +304,17 @@ func GetCapability(rw io.ReadWriter, capa Capability, count, property uint32) (v
...
 }
 
+// GetManufacturer returns the manufacturer ID
+func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
+	caps, _, err := GetCapability(rw, CapabilityTPMProperties, 1, uint32(Manufacturer))
+	if err != nil {
+		return nil, err
+	}
+
+	prop := caps[0].(TaggedProperty)
```
twitchy-jsonp: Should we sanity-check `len(caps) >= 1` here?

josephlr: So if GetCapablity succeeds, it should always return at least the number of properties requested. In theory, the TPM could just do the wrong thing, but in that case, panicking seems fine.

We could also just add a check for this in GetCapability itself, but that seems like it should be in another PR.

tpm/open_windows.go
josephlr: I'll just go test this on an old version of windows server. Windows now [automatically takes ownership of the TPM](https://docs.microsoft.com/en-us/windows/security/information-protection/tpm/initialize-and-configure-ownership-of-the-tpm). But it shouldn't do it on Server 2012

tpm/open_other.go
josephlr: So this will not work if the TPM has not received the TPM2_Startup command (as that has to be the first command the TPM receives).  However, for a physical TPM this should be sent by the firmware ([example](https://github.com/tianocore/edk2/blob/master/SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c#L875-L899)), and for a simulator, it should be sent before go-tpm gets a handle to it ([example](https://github.com/google/go-tpm-tools/blob/5d2fd7f4b3e59d91cba97ab88e8c0777711737a0/simulator/simulator.go#L148-L164)).

Because of this, for instance, the Linux kernel's first command to the TPM is usually `GetCapablity`, not `Startup`. 

Does that seem reasonable?

tpm/open_windows.go
josephlr: I just confirmed that this works out of the box on Windows Server 2012

tpm/open_other.go
twitchy-jsonp: SGTM.

PR #106, opened 2019-05-18
Fix windows build

josephlr:
PR #100 broke the Windows build of go-tpm. This change:
  - Fixes the build when `GOOS=windows` and adds build tags to the examples to make it clear they don't build on windows
  - Simplifies `poll()` to only take a file, since we never use the timeout param
  - Changes file names and build tags to allow for testing on non windows/linux platforms
  - Removes a binary file (`tpm2.test`) that accidentally got committed in #104, and adds a `.gitignore` file to prevent this issue in the future

I don't know how to change the presubmit scripts, but we sould at least make
sure that go-tpm builds on windows (i.e. just making sure `GOOS=windows go build ./...` works)


twitchy-jsonp:
FWICT `go-tpm/tpm` & `go-tpm/tpm2` currently build on Windows (or at least the vendored version under Google's repositories).

LGTM thou.

josephlr:
> FWICT `go-tpm/tpm` & `go-tpm/tpm2` currently build on Windows (or at least the vendored version under Google's repositories).

Ya it looks like the vendored Google repo is at #97, so it didn't pull in the PR that broke things.

PR #108, opened 2019-07-03
Tpm1 2 nv define space / nv write

ChriMarMe:
Add NVDefineSpace, NVWrite for tpm 1.2

### DIFF
--- a/tpm/commands.go
+++ b/tpm/commands.go
@@ -158,10 +158,13 @@ func getCapability(rw io.ReadWriter, cap, subcap uint32) ([]byte, error) {
 	return b, nil
 }
 
-func nvDefineSpace(rw io.ReadWriter, nvData nvDataPublic, ca *commandAuth, encAuth ...authValue) ([]byte, *responseAuth, uint32, error) {
+// nvDefineSpace has to get encAuth of type authValue to be according to spec.
+// This is not implemented
+func nvDefineSpace(rw io.ReadWriter, nvData nvDataPublic, ca *commandAuth) ([]byte, *responseAuth, uint32, error) {
 	var b tpmutil.U32Bytes
 	var ra responseAuth
-	in := []interface{}{nvData, ca, encAuth}
+	// byte(20) is just a place holder for encAuth argument
+	in := []interface{}{nvData, ca, byte(20)}
 	out := []interface{}{&b, &ra}
 	ret, err := submitTPMRequest(rw, tagRQUAuth1Command, ordNVDefineSpace, in, out)
 	if err != nil {
--- a/tpm/commands.go
+++ b/tpm/commands.go
@@ -158,6 +158,18 @@ func getCapability(rw io.ReadWriter, cap, subcap uint32) ([]byte, error) {
 	return b, nil
 }
 
+func nvDefineSpace(rw io.ReadWriter, nvData nvDataPublic, ca *commandAuth, encAuth ...authValue) ([]byte, *responseAuth, uint32, error) {
+	var b tpmutil.U32Bytes
+	var ra responseAuth
+	in := []interface{}{nvData, ca, encAuth}
+	out := []interface{}{&b, &ra}
+	ret, err := submitTPMRequest(rw, tagRQUAuth1Command, ordNVDefineSpace, in, out)
+	if err != nil {
+		return nil, nil, 0, err
+	}
+	return b, &ra, ret, nil
+}
+
 func nvReadValue(rw io.ReadWriter, index, offset, len uint32, ca *commandAuth) ([]byte, *responseAuth, uint32, error) {
 	var b tpmutil.U32Bytes
 	var ra responseAuth
--- a/tpm/commands.go
+++ b/tpm/commands.go
@@ -170,6 +170,18 @@ func nvDefineSpace(rw io.ReadWriter, nvData nvDataPublic, ca *commandAuth, encAu
 	return b, &ra, ret, nil
 }
 
+func nvWriteValue(rw io.ReadWriter, index, offset uint32, data []byte, ca *commandAuth) ([]byte, *responseAuth, uint32, error) {
+	var b tpmutil.U32Bytes
+	var ra responseAuth
+	in := []interface{}{index, offset, binary.Size(data), data, ca}
+	out := []interface{}{&b, &ra}
+	ret, err := submitTPMRequest(rw, tagRQUAuth1Command, ordNVWriteValue, in, out)
+	if err != nil {
+		return nil, nil, 0, err
+	}
+	return b, &ra, ret, nil
+}
+
 func nvReadValue(rw io.ReadWriter, index, offset, len uint32, ca *commandAuth) ([]byte, *responseAuth, uint32, error) {
 	var b tpmutil.U32Bytes
 	var ra responseAuth
--- a/tpm/constants.go
+++ b/tpm/constants.go
@@ -167,23 +167,24 @@ const (
 // Locality Selection
 
 const (
-	locFour  byte = 0x16
-	locThree byte = 0x8
-	locTwo   byte = 0x4
-	locOne   byte = 0x2
-	locZero  byte = 0x1
+	locZero byte = 1 << iota
+	locOne
+	locTwo
+	locThree
+	locFour
 )
 
-// nvPremissions and Operations
+// NV Permissions and Operations
 
 const (
 	nvPerPPWrite    uint32 = 0x00000001
 	nvPerOwnerWrite uint32 = 0x00000002
 	nvPerAuthWrite  uint32 = 0x00000004
 	nvPerWriteAll   uint32 = 0x00000800
-	//REALLY REALLY BAD!!!!!!!!!!!!!!!!
-	//PERMALOCK! Only use if you know what you're doing!!!!
-	//nvPerWriteDefine  uint32 = 0x00001000
+	// Warning: The Value 0x00001000 is defined in the spec as TPM_NV_PER_WRITEDEFINE,
+	// but it is not included directly in this code because it locks the given NV Index permanently
+	// if used uncorrectly. This operation can't be undone in any way.
+	// Do not use this value unless you know what you're doing!
 	nvPerWriteSTClear uint32 = 0x00002000
 	nvPerGlobalLock   uint32 = 0x00004000
 	nvPerPPRead       uint32 = 0x00008000
--- a/tpm/constants.go
+++ b/tpm/constants.go
@@ -53,6 +53,7 @@ const (
 	ordOwnerReadInternalPub uint32 = 0x00000081
 	ordFlushSpecific        uint32 = 0x000000BA
 	ordPcrReset             uint32 = 0x000000C8
+	ordNVDefineSpace        uint32 = 0x000000CC
 	ordNVReadValue          uint32 = 0x000000CF
 )
 
--- a/tpm/constants.go
+++ b/tpm/constants.go
@@ -163,6 +163,32 @@ const (
 	authPrivUseOnly byte = 0x03
 )
 
+// Locality Selection
+
+const (
+	locFour  byte = 0x16
+	locThree byte = 0x8
+	locTwo   byte = 0x4
+	locOne   byte = 0x2
+	locZero  byte = 0x1
+)
+
+// nvPremissions and Operations
+
+const (
+	nvPerPPWrite      uint32 = 0x00000001
+	nvPerOwnerWrite   uint32 = 0x00000002
+	nvPerAuthWrite    uint32 = 0x00000004
+	nvPerWriteAll     uint32 = 0x00000800
+	nvPerWriteDefine  uint32 = 0x00001000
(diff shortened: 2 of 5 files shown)


ChriMarMe:
Hey guys, can I have some comments on that? Thanks in advance :)

josephlr:
Sorry for the delayed response, I don't know anything about TPM1.2, so I'm not the right person to review this PR, someone else on the team might be. 

The main issue is that we have a [simulator](https://github.com/google/go-tpm-tools/tree/master/simulator) for testing the TPM2 related changes (as well as real hardware on our desktops). We don't have any of that for TPM1.2, so reviews are harder.

ChriMarMe:
That's kinda bad news for me :/

awly:
@tmroeder has the most expertise on TPM 1.2.

tpm/constants.go
```diff
@@ -162,6 +164,34 @@ const (
...
+)
+
+// nvPremissions and Operations
+
+const (
+	nvPerPPWrite    uint32 = 0x00000001
+	nvPerOwnerWrite uint32 = 0x00000002
+	nvPerAuthWrite  uint32 = 0x00000004
+	nvPerWriteAll   uint32 = 0x00000800
+	//REALLY REALLY BAD!!!!!!!!!!!!!!!!
```
tmroeder: Please revise this comment to make clear what the danger is for this value.

If the value is so dangerous that it is not provided by the library (since it's commented out), then I'd propose not giving it the form of code at all. For example, these lines could read something like the following (filling in the missing information with the right values):

```go
// WARNING: The value 0x1000 is defined in the spec as <whatever name it has in the spec>,
// but it is not included directly in this code because <reason that it is dangerous>.
// Please do not use that value unless you know what you're doing.
```

Then you have to at least convert this comment into code before using the dangerous value.

tpm/structures.go
```diff
@@ -43,6 +43,11 @@ type pcrSelection struct {
 	Mask pcrMask
 }
 
+type nvAttribute struct {
```
tmroeder: Although these structures are not exported and don't require comments from Go style, we have been trying to keep them commented so that developers understand the details. Please add similar comments to this and other library-private structures.

tpm/structures.go
```diff
@@ -67,6 +72,18 @@ type pcrInfo struct {
 	DigestAtCreation digest
 }
 
+type nvDataPublic struct {
+	tag          tpmutil.Tag
+	index        tpmutil.Handle
+	pcrshort1    pcrInfoShort
```
tmroeder: Go style requires this and the next name to be "pcrShort1". However, I don't want to encode type information in field names, either. Please rename this and the next field to remove the type information ("short").

tpm/structures.go
```diff
@@ -67,6 +72,18 @@ type pcrInfo struct {
...
+	pcrshort2    pcrInfoShort
+	Attrbute     nvAttribute
```
tmroeder: Spelling: "Attrbute" -> "Attribute"

Also, why is this exported when the other fields are not? Should this be unexported to "attribute"?

tpm/tpm.go
```diff
@@ -904,6 +904,61 @@ func ReadEKCert(rw io.ReadWriter, ownAuth digest) ([]byte, error) {
 	return ekbuf, nil
 }
 
+// NVDefineSpace defines space in NVRAM for the indicated index.
+// Definition include the access requirements for writing and reading the area
```
tmroeder: tiny nit: I'd prefer to add the definite article here: "Definition include" -> "The definition includes".

Also, please end the comment with punctuation: in this case, a period.

tpm/tpm.go
```diff
@@ -904,6 +904,61 @@ func ReadEKCert(rw io.ReadWriter, ownAuth digest) ([]byte, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to start new auth session: %v", err)
+	}
+	defer osaprOwn.Close(rw)
+	defer zeroBytes(sharedSecretOwn[:])
+	perm := uint32(0)
+	for _, i := range perms {
+		perm += i
+	}
+	nvdata := nvDataPublic{0x0018, index, pcrInfoShort{}, pcrInfoShort{}, nvAttribute{tpmutil.Tag(0x0017), uint32(perm)}, false, false, false, len}
```
tmroeder: Please add information about the hard-coded constants here. Can they be made into named constants in this file?

tpm/tpm.go
```diff
@@ -904,6 +904,61 @@ func ReadEKCert(rw io.ReadWriter, ownAuth digest) ([]byte, error) {
...
+		return nil, fmt.Errorf("failed to define space in NVRAM: %v", err)
+	}
+	raIn := []interface{}{ret, ordNVDefineSpace, tpmutil.U32Bytes(data)}
+	if err := ra.verify(ca.NonceOdd, sharedSecretOwn[:], raIn); err != nil {
+		return nil, fmt.Errorf("failed to verify authenticity of response: %v", err)
+	}
+	return data, nil
+}
+
+// NVWriteValue writes a given value with its length to a given adress in NVRAM
```
tmroeder: spelling: "adress" -> "address".

Also, consider spelling out more explicitly what "address" means in this context. I think it means that data will be written to len(data) bytes starting at index + offset.

tpm/tpm_test.go
```diff
@@ -232,6 +232,33 @@ func TestResizeableSlice(t *testing.T) {
 	}
 }
 
+func TestNVDefineSpace(t *testing.T) {
```
tmroeder: Have these been run successfully on a simulator?

tpm/tpm_test.go
```diff
@@ -232,6 +232,33 @@ func TestResizeableSlice(t *testing.T) {
...
+	rwc := openTPMOrSkip(t)
+	defer rwc.Close()
+	index := tpmutil.Handle(0x40000001)
+	permissions := []uint32{nvPerWriteSTClear}
+	length := uint32(54)
```
tmroeder: Where does "54" come from in this example? Can it be calculated in the code instead of being hard coded?

tpm/tpm_test.go
```diff
@@ -232,6 +232,33 @@ func TestResizeableSlice(t *testing.T) {
...
+	}
+
+}
+
+func TestNVWriteValue(t *testing.T) {
+	rwc := openTPMOrSkip(t)
+	defer rwc.Close()
+	index := uint32(0x40000001)
+	ownerAuth := getAuth(ownerAuthEnvVar)
+	randomdata := make([]byte, 20)
```
tmroeder: Go style insists on "randomData" over "randomdata" (since it's two words).

However, it would be even better in Go to call this simply `r`.

tpm/tpm_test.go
```diff
@@ -232,6 +232,33 @@ func TestResizeableSlice(t *testing.T) {
...
+	rand.Read(randomdata)
```
tmroeder: Please check the return value of Read and fail if it fails.

tpm/tpm_test.go
```diff
@@ -232,6 +232,33 @@ func TestResizeableSlice(t *testing.T) {
...
 }
 
+func TestNVDefineSpace(t *testing.T) {
+	rwc := openTPMOrSkip(t)
+	defer rwc.Close()
+	index := tpmutil.Handle(0x40000001)
+	permissions := []uint32{nvPerWriteSTClear}
+	length := uint32(54)
+	ownerAuth := getAuth(ownerAuthEnvVar)
+	_, err := NVDefineSpace(rwc, index, length, permissions, ownerAuth)
```
tmroeder: Please fold this into the if statement if the return value isn't going to be checked. Is there something that can be checked here?

tpm/tpm_test.go
```diff
@@ -232,6 +232,33 @@ func TestResizeableSlice(t *testing.T) {
...
+
+func TestNVWriteValue(t *testing.T) {
+	rwc := openTPMOrSkip(t)
+	defer rwc.Close()
+	index := uint32(0x40000001)
+	ownerAuth := getAuth(ownerAuthEnvVar)
+	randomdata := make([]byte, 20)
+	rand.Read(randomdata)
+	_, err := NVWriteValue(rwc, index, 0, randomdata, ownerAuth)
+	if err != nil {
```
tmroeder: Same comment as in the previous test: if it's not possible to test anything to do with the return value, then fold the previous line into this one. However, I'd prefer to check the return value if possible.

tpm/tpm_test.go
ChriMarMe: The check for a successful operation is done in submitTPMRequest-Function.
I could parse the the response from the tpm, but I cant see the benefit here.

tpm/tpm_test.go
ChriMarMe: DefineSpace has a variable length, so i just typed 54. Can't be "calculated" because depends on usecase.

tpm/commands.go
```diff
@@ -158,6 +158,30 @@ func getCapability(rw io.ReadWriter, cap, subcap uint32) ([]byte, error) {
 	return b, nil
 }
 
+func nvDefineSpace(rw io.ReadWriter, nvData nvDataPublic, ca *commandAuth, encAuth ...authValue) ([]byte, *responseAuth, uint32, error) {
+	var b tpmutil.U32Bytes
+	var ra responseAuth
+	in := []interface{}{nvData, ca, encAuth}
```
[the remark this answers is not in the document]
ChriMarMe: I guess my thought was something like: go hasn't default function-arguments, so i used variadic insteat. 

Now that I'm looking at it, i see that i didn't implement the encAuth process at all. Should do it for completion. Could take some time. Haven't touch the code in a month, need to wrap my head around it again :)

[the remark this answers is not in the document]
ChriMarMe: I implemented the variadic argument just so the function head is complete but the argument vor encAuth is optional. For my use case it's sufficient and I don't see the need for implementing more functions just for elaborate implementation of case sensitive encAuth.

tpm/tpm_test.go
tmroeder: I'm not sure I understand this response. Are you saying that 54 is arbitrary and doesn't correspond to any of the rest of the code? If so, then please add a comment to say that.

tpm/commands.go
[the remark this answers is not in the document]
tmroeder: Writing the function signature with a final variadic argument tells callers that it is valid to pass more than one value here: variadic parameters are not the same as optional parameters. I only see one call to this function in the implementation, and it is not using this argument.

If this is not currently needed for the code, then drop the parameter entirely and add a comment about not implementing that part of the spec.

tpm/constants.go
```diff
@@ -162,6 +164,35 @@ const (
...
+// NV Permissions and Operations
+
+const (
+	nvPerPPWrite    uint32 = 0x00000001
+	nvPerOwnerWrite uint32 = 0x00000002
+	nvPerAuthWrite  uint32 = 0x00000004
+	nvPerWriteAll   uint32 = 0x00000800
+	// Warning: The Value 0x00001000 is defined in the spec as TPM_NV_PER_WRITEDEFINE,
+	// but it is not included directly in this code because it locks the given NV Index permanently
+	// if used uncorrectly. This operation can't be undone in any way.
```
tmroeder: minor usage error: "uncorrectly" is not an English word. This should be "incorrectly".

tpm/constants.go
```diff
@@ -162,6 +164,35 @@ const (
...
+
+// NV Permissions and Operations
+
+const (
+	nvPerPPWrite    uint32 = 0x00000001
+	nvPerOwnerWrite uint32 = 0x00000002
+	nvPerAuthWrite  uint32 = 0x00000004
+	nvPerWriteAll   uint32 = 0x00000800
+	// Warning: The Value 0x00001000 is defined in the spec as TPM_NV_PER_WRITEDEFINE,
+	// but it is not included directly in this code because it locks the given NV Index permanently
```
tmroeder: Please reflow the text of this comment to even out the lines to match line length of the rest of the code.

tpm/tpm_test.go
```diff
@@ -232,6 +233,35 @@ func TestResizeableSlice(t *testing.T) {
...
+	defer rwc.Close()
+	index := uint32(0x40000001)
+	ownerAuth := getAuth(ownerAuthEnvVar)
+	r := make([]byte, 20)
+	_, err := rand.Read(r)
+	if err != nil {
+		t.Fatal("Couldn't read random:", err)
+	}
+
+	if _, err = NVWriteValue(rwc, index, 0, r, ownerAuth); err != nil {
```
tmroeder: Please also read back the value and confirm that it matches the value that was written. That will give an end-to-end test of the consistency of NVWriteValue and NVReadValue.

tpm/commands.go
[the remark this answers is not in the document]
ChriMarMe: Alright

tpm/constants.go
ChriMarMe: whopsy. :D

PR #110, opened 2019-07-18
Fix parameter decoding on TPM2

### LINKED ISSUE #102: tpm2: CreateKey should not return a tpmutil.Handle
The TPM2_Create command does not return a handle, but tpm2.CreateKey tries to decode and return one anyway, resulting in a bogus tpmutil.Handle value.

Relatedly, the TPM2_CreateLoaded command doesn't seem available.

josephlr:
Fixes #102

[Part 3 of the Spec](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38-code.pdf) (sections 4.2.1 and 4.3) describe how, when a command uses Authorization Tags, there is an additional UINT32 `paramSize` in the response, between the handles and parameters. We usually properly deal with this parameter, but not always.

This [spreadsheet of TPM2 commands](https://docs.google.com/spreadsheets/d/1WfZCb1G7wK0tZetN51lI7bPWbazSY3m5AlUYIzRjhK0/edit?usp=sharing) lists all the TPM2 commands (including if they should have authorization sessions). We only care about commands where:
  - An authorization session is required
  - We care about the return parameters (otherwise we won't even bother decoding the response).

The only commands we were messing up were TPM2_Create and TPM2_PolicySecret. These are now fixed.

EDIT: Note that the changes to `Create` are backwards incompatible, so incrementing the version will be required.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -534,31 +534,31 @@ func ReadPublic(rw io.ReadWriter, handle tpmutil.Handle) (Public, []byte, []byte
 	return decodeReadPublic(resp)
 }
 
-func decodeCreate(in []byte) (handle tpmutil.Handle, private, public, creationData, creationHash tpmutil.U16Bytes, creationTicket Ticket, err error) {
+func decodeCreate(in []byte) (private, public, creationData, creationHash tpmutil.U16Bytes, creationTicket Ticket, err error) {
 	buf := bytes.NewBuffer(in)
-
-	if err := tpmutil.UnpackBuf(buf, &handle, &private, &public, &creationData, &creationHash); err != nil {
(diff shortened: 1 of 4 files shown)


tpm2/tpm2.go
```diff
@@ -705,11 +721,7 @@ func PolicySecret(rw io.ReadWriter, entityHandle tpmutil.Handle, entityAuth Auth
 
 	// Tickets are only provided if expiry is set.
 	if expiry != 0 {
-		tk, err := decodeTicket(bytes.NewBuffer(resp))
```
twitchy-jsonp: Am I correct in understanding this didn't work in the current implementation, because it was missing decode for the `TPM2B_TIMEOUT` parameter?

PR #114, opened 2019-07-19
Add Import command

josephlr:
Depends on #122 (for proper testing).

This allows TPM2_Import to be called from go-tpm.

The code is ready for review, but this PR still needs tests.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -578,6 +578,50 @@ func Seal(rw io.ReadWriter, parentHandle tpmutil.Handle, parentPassword, objectP
 	return private, public, nil
 }
 
+func encodeImport(parentHandle tpmutil.Handle, auth AuthCommand, publicBlob, privateBlob, symSeed, encryptionKey tpmutil.U16Bytes, sym *SymScheme) ([]byte, error) {
+	ph, err := tpmutil.Pack(parentHandle)
+	if err != nil {
+		return nil, err
+	}
+	encodedAuth, err := encodeAuthArea(auth)
+	if err != nil {
+		return nil, err
+	}
+	data, err := tpmutil.Pack(encryptionKey, publicBlob, privateBlob, symSeed)
+	if err != nil {
+		return nil, err
+	}
+	encodedScheme, err := sym.encode()
+	if err != nil {
+		return nil, err
+	}
+
+	return concat(ph, encodedAuth, data, encodedScheme)
+}
+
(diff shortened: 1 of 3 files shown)


tpm2/tpm2.go
```diff
@@ -578,6 +578,50 @@ func Seal(rw io.ReadWriter, parentHandle tpmutil.Handle, parentPassword, objectP
...
+	var outPrivate tpmutil.U16Bytes
+	_, err := tpmutil.Unpack(resp, &paramSize, &outPrivate)
+	return outPrivate, err
+}
+
+// ImportUsingAuth allows a user to import a key created on a different computer
+// or in a different TPM. The publicBlob and privateBlob must always be
+// provided. symSeed should be non-nil iff an "outer wrapper" is used. Both of
+// encryptionKey and sym should be non-nil iff an "inner wrapper" is used.
+func ImportUsingAuth(rw io.ReadWriter, parentHandle tpmutil.Handle, auth AuthCommand, publicBlob, privateBlob, symSeed, encryptionKey []byte, sym *SymScheme) ([]byte, error) {
```
awly: I'm not a huge fan of this "UsingAuth" naming pattern.
Can this command be used without auth?

If the other alternative is using PasswordSession, maybe we refactor everything to take `AuthCommand` and add a helper `func PasswordAuth(pass string) AuthCommand`?

josephlr:
> Add a test please

Done, the test is a just a "plain" Import that doesn't use an inner or outer wrapper. See https://github.com/google/go-tpm-tools/pull/36 for more complex tests.

tpm2/tpm2.go
josephlr: > I'm not a huge fan of this "UsingAuth" naming pattern.
> Can this command be used without auth?

Nope, it will always need some sort of authorization (usually session authorization). I changed the name to `Import` to reflect this.

> If the other alternative is using PasswordSession, maybe we refactor everything to take `AuthCommand` and add a helper `func PasswordAuth(pass string) AuthCommand`?

I think a better idea might be to have a common `Auth` interface which is implemented by `PasswordAuth` and `SessionAuth`, allowing a user to just pass`PasswordAuth` or `SessionAuth` whereever an auth value is needed, but that's something we should probably address later.

giapnguyen74:
Could you give an example for import a AES 128 bit key? I try to replace tmp2_import external aes key but no luck. Thanks.

PR #117, opened 2019-07-23
Add Public.MatchesTemplate

josephlr:
For example, this makes it easier to check if a persistent handle is the cached version of a key we want to create, allowing us to write something like:

```go
func getKey(template Public, cached Handle) Handle {
    pub, _, _, err := ReadPublic(rw, cached)
    if err == nil && pub.MatchesTemplate(template) {
        return cached
    }
    return CreatePrimary(rw, ... , template)
}
```

### DIFF
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -61,12 +61,12 @@ type Public struct {
 	Attributes KeyProp
 	AuthPolicy tpmutil.U16Bytes
 
-	// Only one of the following fields should be set. When encoding/decoding,
-	// one will be picked based on Type.
+	// Exactly one of the following fields should be set
+	// When encoding/decoding, one will be picked based on Type.
 	RSAParameters       *RSAParams
 	ECCParameters       *ECCParams
 	SymCipherParameters *SymCipherParams
-	KeyedHashParameters *KeyedHashParams // Optional for AlgKeyedHash
+	KeyedHashParameters *KeyedHashParams
 }
 
 // Encode serializes a Public structure in TPM wire format.
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -66,7 +66,7 @@ type Public struct {
 	RSAParameters       *RSAParams
 	ECCParameters       *ECCParams
 	SymCipherParameters *SymCipherParams
-	KeyedHashUnique     tpmutil.U16Bytes // Optional for AlgKeyedHash
+	KeyedHashParameters *KeyedHashParams // Optional for AlgKeyedHash
 }
 
 // Encode serializes a Public structure in TPM wire format.
@@ -80,7 +80,7 @@ func (p Public) Encode() ([]byte, error) {
 	case AlgRSA:
 		params, err = p.RSAParameters.encode()
 	case AlgKeyedHash:
-		params, err = tpmutil.Pack(AlgNull, p.KeyedHashUnique)
+		params, err = p.KeyedHashParameters.encode()
 	case AlgECC:
 		params, err = p.ECCParameters.encode()
 	case AlgSymCipher:
@@ -156,6 +156,8 @@ func (p Public) MatchesTemplate(template Public) bool {
 		return p.ECCParameters.matchesTemplate(template.ECCParameters)
 	case AlgSymCipher:
 		return p.SymCipherParameters.matchesTemplate(template.SymCipherParameters)
+	case AlgKeyedHash:
+		return p.KeyedHashParameters.matchesTemplate(template.KeyedHashParameters)
 	default:
 		return true
 	}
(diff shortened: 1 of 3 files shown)


tpm2/structures.go
```diff
@@ -137,6 +138,49 @@ func (p Public) Name() (*Name, error) {
...
+		pExponent := pParams.Exponent
+		if pExponent == 0 {
+			pExponent = defaultRSAExponent
+		}
+		tExponent := tParams.Exponent
+		if tExponent == 0 {
+			tExponent = defaultRSAExponent
+		}
+		return reflect.DeepEqual(pParams.Symmetric, tParams.Symmetric) &&
+			reflect.DeepEqual(pParams.Sign, tParams.Sign) &&
```
twitchy-jsonp: iirc these are fixed-size structs, can we compare them directly instead?

tpm2/tpm2_test.go
```diff
@@ -1363,3 +1363,47 @@ func TestCreatePrimaryRawTemplate(t *testing.T) {
...
+			Type:       AlgRSA,
+			NameAlg:    AlgSHA256,
+			Attributes: FlagSignerDefault,
+			RSAParameters: &RSAParams{
+				Sign: &SigScheme{
+					Alg:  AlgRSASSA,
+					Hash: AlgSHA256,
+				},
+				KeyBits:  2048,
+				Exponent: 0,
```
twitchy-jsonp: Nit: the `Exponent` field will never contain zero in a `Public` that was wire decoded - it will always be rewritten to `defaultRSAExponent` and have the `encodeDefaultExponentAsZero` bit set. Maybe we should test using the `defaultRSAExponent` value?

tpm2/structures.go
```diff
@@ -137,6 +138,49 @@ func (p Public) Name() (*Name, error) {
...
+// both provided in the template and overriden in the key creation process.
+func (p Public) MatchesTemplate(template Public) bool {
+	if p.Type != template.Type ||
+		p.NameAlg != template.NameAlg ||
+		p.Attributes != template.Attributes ||
+		!bytes.Equal(p.AuthPolicy, template.AuthPolicy) {
+		return false
+	}
+	switch p.Type {
+	case AlgRSA:
```
awly: How about moving type-specific logic into methods on those types?
e.g. code in this `case` would be `return p.RSAParameters.matchesTemplate(template.RSAParameters)`

tpm2/tpm2_test.go
awly: or test both?

tpm2/tpm2_test.go
```diff
@@ -1363,3 +1363,47 @@ func TestCreatePrimaryRawTemplate(t *testing.T) {
 		t.Errorf("got key exponent %v, want %v", pubRSA.E, defaultKeyParams.RSAParameters.Exponent)
 	}
 }
+
+func TestMatchesTemplate(t *testing.T) {
```
awly: Add tests for other types of template (ECC and SymCipher)
Ideally use subtests here (`t.Run`)

tpm2/tpm2_test.go
```diff
@@ -1363,3 +1363,112 @@ func TestCreatePrimaryRawTemplate(t *testing.T) {
...
+					Type:       AlgKeyedHash,
+					NameAlg:    AlgSHA256,
+					Attributes: FlagSignerDefault,
+					KeyedHashParameters: &KeyedHashParams{
+						Alg: AlgNull,
+					},
+				}
+			},
+			func(pub *Public) { pub.KeyedHashParameters.Unique = make([]byte, 256) },
+			func(pub *Public) { pub.Attributes |= FlagNoDA },
```
awly: Use AlgHMAC in KeyedHashParameters and change pub.KeyedHashParameters.Hash here

tpm2/tpm2_test.go
```diff
@@ -1363,3 +1363,112 @@ func TestCreatePrimaryRawTemplate(t *testing.T) {
...
 }
+
+func TestMatchesTemplate(t *testing.T) {
+	var tests = []struct {
+		name       string
+		makePublic func() Public
+		goodChange func(*Public)
+		badChange  func(*Public)
+	}{
+		{"RSA",
```
awly: move test names on new line, aligned with all other fields

tpm2/tpm2_test.go
```diff
@@ -1363,3 +1363,112 @@ func TestCreatePrimaryRawTemplate(t *testing.T) {
 		t.Errorf("got key exponent %v, want %v", pubRSA.E, defaultKeyParams.RSAParameters.Exponent)
 	}
 }
+
+func TestMatchesTemplate(t *testing.T) {
+	var tests = []struct {
```
awly: `tests := []struct {`

only use var at package level or for declaring zero values

awly: Add another case where pub.Type doesn't match

tpm2/structures.go
josephlr: The main reason this doesn't work is when we are dealing with pointer types. We have `var a *Foo` and `var b *Foo`, and we want them to compare equal if they are either both `nil`, or their field values compare equal. Unfortunately, there's not an easy way to do this in go.

Comparing like `a == b` will check if they point to the same object (not what we want), and `*a == *b` risks a segfault on `nil` pointers. Unfortunately, it seems like `reflect.DeepEqual` is the easiest wa

twitchy-jsonp: SGTM

josephlr:
I updated the comments on the new `KeyedHashParameters` to make it clear that the parameter is not optional when `Type == AlgKeyedHash`. This should avoid any nullptr dereference problems.

Issue #115, opened 2019-07-22
tpm2.Public cannot create default ECC Template


josephlr:
In [go-tpm-tools](https://github.com/google/go-tpm-tools), we want to be able to generate the default TPM2 EK/SRK for ECC [like we are currently doing for RSA](https://github.com/google/go-tpm-tools/blob/de523c4444ff9d69578766d9a7f3c598f8276a25/tpm2tools/template.go#L23-L44).

However, the [TCG Credential Profile EK](https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_Credential_Profile_EK_V2.1_R13.pdf) says the the buffers representing X, Y for ECC curves must be 16 bytes of all zeros each. This doesn't work for the current `tpm2.Public` as that's equivalent to the number 0, resulting in an incorrect encoding.

This is arguably a bug by the TCG (and they fixed it for later templates, see the templates in B.4 of the above linked doc). However, this "bug" is standardized so we need to support it. We solved a similar issue for `RSAParams` by using `ModulusRaw`, so we probably also need  `XRaw` and `YRaw` parameters for `ECPoint`.

PR #118, opened 2019-07-23
Update ECPoint type

lulumel0n:
Per https://github.com/google/go-tpm/issues/115
Allow the ECCParams to supply buffers as X and Y for the Point field, in order to comply with the Default ECC EK Public Area Template.

When needed, the buffer will be converted to Int.

This change is not backward compatible, existing tests/code using ECPoint will need to be updated.


tpm2/structures.go
```diff
@@ -256,23 +256,25 @@ type ECCParams struct {
...
 	}
-	return p.X
+	return x.SetBytes(p.XRaw)
 }
 
-func (p *ECPoint) y() *big.Int {
-	if p == nil || p.Y == nil {
-		return big.NewInt(0)
+func (p ECPoint) Y() *big.Int {
+	y := big.NewInt(0)
```
josephlr: you can just use `new(big.Int).SetBytes(x)` and `new(big.Int).SetBytes(y)` here. The nil slice of length zero works fine with `SetBytes`.

tpm2/structures.go
```diff
@@ -295,8 +297,9 @@ func (p *ECCParams) encode() ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("encoding KDF: %v", err)
 	}
-	x, y := p.Point.x().Bytes(), p.Point.y().Bytes()
-	point, err := tpmutil.Pack(tpmutil.U16Bytes(x), tpmutil.U16Bytes(y))
+
```
josephlr: Delete these blank lines.

tpm2/tpm2_test.go
```diff
@@ -727,7 +727,7 @@ func TestSign(t *testing.T) {
 					Hash: AlgSHA256,
 				},
 				CurveID: CurveNISTP256,
-				Point:   ECPoint{X: big.NewInt(0), Y: big.NewInt(0)},
+				Point:   ECPoint{XRaw: big.NewInt(0).Bytes(), YRaw: big.NewInt(0).Bytes()},
```
josephlr: Just delete this line (the default value of `XRaw=nil, YRaw=nil` is fine here)

tpm2/encoding_test.go
```diff
@@ -385,6 +385,14 @@ func TestECCParamsEncodeDecode(t *testing.T) {
 		t.Fatalf("decodeECCParams: %v", err)
 	}
 
+	if params.Point.X().Cmp(pk.PublicKey.X) != 0 || params.Point.Y().Cmp(pk.PublicKey.Y) != 0 {
+		t.Fatalf("Point deserialized: (%d, %d)\nGenerated: (%d, %d)", params.Point.X(), params.Point.Y(), pk.PublicKey.X, pk.PublicKey.Y)
```
josephlr: For these `FatalF` messages, just use the shorter `"got: %+v\nwant: %+v"` format string, like we do below.

josephlr:
Also, please update the PR description to reflect the new changes

tpm2/structures.go
lulumel0n: Done

tpm2/structures.go
lulumel0n: Done

tpm2/tpm2_test.go
lulumel0n: Done

tpm2/encoding_test.go
lulumel0n: Done

tpm2/encoding_test.go
```diff
@@ -385,6 +385,19 @@ func TestECCParamsEncodeDecode(t *testing.T) {
 		t.Fatalf("decodeECCParams: %v", err)
 	}
 
+	if params.Point.X().Cmp(pk.PublicKey.X) != 0 {
+		t.Fatalf("got: %+v\nwant: %+v", params.Point.X(), pk.PublicKey.X)
```
awly: Clarify what got/want in these failure messages are

awly: Sorry, I meant in the actual `t.Fatalf` text, so you don't need to look at test code to understand which check failed.

lulumel0n: np, added

tpm2/structures.go
```diff
@@ -295,8 +289,7 @@ func (p *ECCParams) encode() ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("encoding KDF: %v", err)
 	}
-	x, y := p.Point.x().Bytes(), p.Point.y().Bytes()
-	point, err := tpmutil.Pack(tpmutil.U16Bytes(x), tpmutil.U16Bytes(y))
+	point, err := tpmutil.Pack(tpmutil.U16Bytes(p.Point.XRaw), tpmutil.U16Bytes(p.Point.YRaw))
```
josephlr: `tpmutil.U16Bytes` isn't needed here (as that's already the type of `p.Point.XRaw`.

`tpmutil.Pack(p.Point.XRaw, p.Point.YRaw)` should work.

josephlr:
@awly, this is a backwards incompatible change, but makes dealing with encoding/decoding issues much easier.

Would you want to do the same thing for `RSAParams.Modulus` (in a followup PR)? The main downside is that RSA keys are more common, so it might break a little more code, but I think we should do it.

tpm2/structures.go
lulumel0n: good catch, done

awly:
SGTM, we always assume that importers will lock the version of this library.
Breaking API isn't a problem.

PR #122, opened 2019-09-04
Move TPM2 integrations tests to use simulator

josephlr:
This PR now makes it so integration tests run against the simulator by default, instead of being skipped.
This required:
  
  - moving the integration tests to their own package (to prevent an import loop)
  - exposing `Encode` and `DecodeName` for the `Name` structure
  - moving some test cases between files  
  - a bunch of smaller, minor changes

EDIT: Given the nature of this PR, reviewing commit by commit should be the easiest. As each individual change is not that large.

@twitchy-jsonp @awly, do any changes need to be made to the CI to account for the fact that tests are now in `tpm2` and `tpm2/test`?


twitchy-jsonp:
CI runs the following from the project root:

```shell
go build github.com/google/go-tpm/...
go test github.com/google/go-tpm/...
```

I think thats sufficient?

I have an open action item to finish implementing TPM 1.2 CI integration.

tpm2/structures.go
```diff
@@ -888,7 +888,8 @@ type Name struct {
 	Digest *HashValue
 }
 
-func decodeName(in *bytes.Buffer) (*Name, error) {
+// DecodeName deserializes a Name hash from the TPM wire format.
```
twitchy-jsonp: Maybe mention this is a `TPM2B_NAME` structure explicitly?

josephlr:
> CI runs the following from the project root:
> 
> ```shell
> go build github.com/google/go-tpm/...
> go test github.com/google/go-tpm/...
> ```
> 
> I think thats sufficient?

We will also need to run `go get -d -t -u github.com/google/go-tpm/...` to properly download any necessary dependencies (i.e. the simulator). That seems to be the reason the CI is failing right now.

tpm2/structures.go
josephlr: Done,  added a comment on the struct itself.

PR #125, opened 2019-09-06
Fix broken go.mod (missing all dependencies)

twitchy-jsonp:


twitchy-jsonp:
The sum of indirect dependencies is recorded to provide integrity, without relying on the integrity of any module proxy.

josephlr:
Ahh and It's done on the module level, not the package level. That makes
sense.

PR #126, opened 2019-09-07
Cirrus

josephlr:
Configuration is much easier, and we avoid needing internal CI
configuration. Files based on those in https://github.com/google/go-tpm-tools/pull/39

EDIT: it looks like Cirrus didn't get properly installed, so this isn't working yet. We will have to wait on the admin to enable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
 FROM golang:latest
 # We need OpenSSL headers to build the simulator
-RUN apt-get update && apt-get install -y libssl-dev
+RUN apt-get update && apt-get install -y \
+    libssl-dev \
+ && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM golang:latest
+# We need OpenSSL headers to build the simulator
+RUN apt-get update && apt-get install -y libssl-dev
```
fkorotkov: Would suggest to remove apt cache to minimize the image. No need to have information about all the available packages in the final image. 😅 

```suggestion
RUN apt-get update && apt-get install -y libssl-dev && rm -rf /var/lib/apt/lists/*
```

josephlr: Thanks! Fixed here and in https://github.com/google/go-tpm-tools/pull/42

PR #128, opened 2019-09-12
TPM1.2: Add functions GetAlgs and GetNVList

ChriMarMe:
Two more functions for TPM1.2

### DIFF
--- a/tpm/constants.go
+++ b/tpm/constants.go
@@ -58,8 +58,10 @@ const (
 
 // Capability types.
 const (
+	capAlg      uint32 = 0x00000002
 	capProperty uint32 = 0x00000005
 	capFlag     uint32 = 0x00000004
+	capNVList   uint32 = 0x0000000D
 	capHandle   uint32 = 0x00000014
 )
 
@@ -110,23 +112,46 @@ const (
 	pidTransport
 )
 
+// Algorithm type for more convenient handling.
+// see Algorithm ID for possible values.
+type Algorithm uint32
+
 // Algorithm ID values.
 const (
-	_ uint32 = iota
-	algRSA
+	_ Algorithm = iota
+	AlgRSA
 	_ // was DES
 	_ // was 3DES in EDE mode
-	algSHA
-	algHMAC
-	algAES128
-	algMGF1
-	algAES192
-	algAES256
-	algXOR
+	AlgSHA
+	AlgHMAC
+	AlgAES128
+	AlgMGF1
+	AlgAES192
+	AlgAES256
+	AlgXOR
 )
 
+// AlgMap Map of Algorithms to Strings for nicer output and comparisons, etc.
+var AlgMap = map[Algorithm]string{
+	AlgRSA:    "RSA",
+	AlgSHA:    "SHA1",
+	AlgHMAC:   "HMAC",
+	AlgAES128: "AER128",
+	AlgMGF1:   "MFG1",
+	AlgAES192: "AES192",
+	AlgAES256: "AES256",
+}
+
+func (a Algorithm) String() string {
+	n, ok := AlgMap[a]
+	if !ok {
+		return "unknown_algorithm"
+	}
+	return n
+}
+
 // Encryption schemes. The values esNone and the two that contain the string
-// "RSA" are only valid under algRSA. The other two are symmetric encryption
+// "RSA" are only valid under AlgRSA. The other two are symmetric encryption
 // schemes.
 const (
 	_ uint16 = iota
@@ -138,7 +163,7 @@ const (
 	esSymCBCPKCS5 = 0xff // esSymCBCPKCS5 was taken from go-tspi
 )
 
-// Signature schemes. These are only valid under algRSA.
+// Signature schemes. These are only valid under AlgRSA.
 const (
 	_ uint16 = iota
 	ssNone
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -567,7 +567,7 @@ func MakeIdentity(rw io.ReadWriter, srkAuth []byte, ownerAuth []byte, aikAuth []
 	}
 
 	aikParams := keyParams{
-		AlgID:     algRSA,
+		AlgID:     AlgRSA,
 		EncScheme: esNone,
 		SigScheme: ssRSASaPKCS1v15SHA1,
 		Params:    packedParams,
@@ -707,7 +707,7 @@ func ActivateIdentity(rw io.ReadWriter, aikAuth []byte, ownerAuth []byte, aik tp
 		secret    []byte
 	)
 	switch id := symkey.AlgID; id {
-	case algAES128:
+	case AlgAES128:
 		block, err = aes.NewCipher(symkey.Key)
 		if err != nil {
 			return nil, fmt.Errorf("aes.NewCipher failed: %v", err)
@@ -999,6 +999,35 @@ func GetPermanentFlags(rw io.ReadWriter) (PermanentFlags, error) {
 	return ret, err
 }
 
+// GetAlgs returns a list of algorithms supported by the TPM device.
+func GetAlgs(rw io.ReadWriter) ([]Algorithm, error) {
+	var algs []Algorithm
+	for i := AlgRSA; i <= AlgXOR; i++ {
+		buf, err := getCapability(rw, capAlg, uint32(i))
+		if err != nil {
+			return nil, err
+		}
+		if uint8(buf[0]) > 0 {
+			algs = append(algs, Algorithm(i))
+		}
+
+	}
+	return algs, nil
+}
+
+// GetNVList returns a list of TPM_NV_INDEX values that
+// are currently allocated NV storage through TPM_NV_DefineSpace.
+func GetNVList(rw io.ReadWriter) ([]uint32, error) {
+	var nvList []uint32
+	buf, err := getCapability(rw, capNVList, 0)
+	if err != nil {
+		return nil, err
+	}
+	r := bytes.NewReader(buf)
+	err = binary.Read(r, binary.LittleEndian, &nvList)
+	return nvList, err
+}
+
 // OwnerClear uses owner auth to clear the TPM. After this operation, the TPM
 // can change ownership.
 func OwnerClear(rw io.ReadWriter, ownerAuth digest) error {
@@ -1072,7 +1101,7 @@ func TakeOwnership(rw io.ReadWriter, newOwnerAuth digest, newSRKAuth digest, pub
 		return err
 	}
 	srkParams := keyParams{
-		AlgID:     algRSA,
+		AlgID:     AlgRSA,
 		EncScheme: esRSAEsOAEPSHA1MGF1,
 		SigScheme: ssNone,
 		Params:    srkpb,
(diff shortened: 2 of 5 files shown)


ChriMarMe:
@tmroeder 
Would you be so kind and have a look?

ChriMarMe:
Really need some review here please

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
...
+// A byte is 1 a supported algorithm
+func GetAlgs(rw io.ReadWriter) ([]byte, error) {
+	var algs []byte
+	for i := uint32(1); i <= 10; i++ {
+		buf, err := getCapability(rw, capAlg, i)
+		if err != nil {
+			return nil, err
+		}
+		if buf == nil {
+			return nil, fmt.Errorf("Function GetAlgs() returned no error and no value")
```
awly: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
 	return getCapability(rw, capProperty, tpmCapPropManufacturer)
 }
 
+// GetAlgs returns a slice of bytes.
+// A byte is 1 a supported algorithm
```
awly: To clarify - each byte is a const identifying an algorithm, right?
Can you define type Algorithm and a bunch of consts for it?

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
 	return getCapability(rw, capProperty, tpmCapPropManufacturer)
 }
 
+// GetAlgs returns a slice of bytes.
```
awly: "GetAlgs returns a list of algorithms supported by the TPM device."

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
...
+	for i := uint32(1); i <= 10; i++ {
+		buf, err := getCapability(rw, capAlg, i)
+		if err != nil {
+			return nil, err
+		}
+		if buf == nil {
+			return nil, fmt.Errorf("Function GetAlgs() returned no error and no value")
+		}
+		for _, item := range buf {
+			algs = append(algs, item)
```
awly: I don't know how to read the 1.2 spec yet.
What do you expect to see in buf for each getCapability request?

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
...
+		}
+
+	}
+	return algs, nil
+}
+
+func GetNVList(rw io.ReadWriter) ([]uint32, error) {
```
awly: Add a doc comment

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
...
+	var nvlist []uint32
```
awly: nvList

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
 	return getCapability(rw, capProperty, tpmCapPropManufacturer)
 }
 
+// GetAlgs returns a slice of bytes.
+// A byte is 1 a supported algorithm
+func GetAlgs(rw io.ReadWriter) ([]byte, error) {
+	var algs []byte
+	for i := uint32(1); i <= 10; i++ {
```
awly: use the algorithm consts I mentioned above

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
...
+		buf, err := getCapability(rw, capAlg, i)
```
awly: is there a way to query all of them at once?

tpm/tpm.go
```diff
@@ -975,6 +975,40 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
...
+}
+
+func GetNVList(rw io.ReadWriter) ([]uint32, error) {
+	var nvlist []uint32
+	buf, err := getCapability(rw, capNVList, 0)
+	if err != nil {
+		return nil, err
+	}
+	r := bytes.NewReader(buf)
+	err = binary.Read(r, binary.LittleEndian, &nvlist)
```
awly:
```
err = binary.Read(r, binary.LittleEndian, &nvlist)
return nvList, err
```

tpm/tpm_test.go
```diff
@@ -70,6 +70,28 @@ func TestGetManufacturer(t *testing.T) {
 	t.Logf("TPM VendorID: %v\n", vendorID)
 }
 
+func TestGetAlgs(t *testing.T) {
+	rwc := openTPMOrSkip(t)
+	defer rwc.Close()
+
+	algs, err := GetAlgs(rwc)
+	if err != nil {
+		t.Fatal("Couldn't read Algorithms")
```
awly: add err to the message

tpm/tpm_test.go
```diff
@@ -70,6 +70,28 @@ func TestGetManufacturer(t *testing.T) {
...
+	}
+	t.Logf("Algorithms are: %v\n", algs)
```
awly: can you compare this against a known set of good values?
the spec says "The TPM MUST support the algorithms TPM_ALG_RSA, TPM_ALG_SHA, TPM_ALG_HMAC, and TPM_ALG_MGF1"

tpm/tpm_test.go
```diff
@@ -70,6 +70,28 @@ func TestGetManufacturer(t *testing.T) {
...
+}
+
+func TestGetNVList(t *testing.T) {
+	rwc := openTPMOrSkip(t)
+	defer rwc.Close()
+
+	nvlist, err := GetNVList(rwc)
+	if err != nil {
+		t.Fatal("Couldn't read NVList")
```
awly: add err to the message

tpm/tpm.go
ChriMarMe: The spec doesn't tell me how many bits/bytes are returned to repesent the boolean. I just testes it and i just get one byte.

tpm/constants.go
```diff
@@ -108,9 +110,11 @@ const (
 	pidTransport
 )
 
+type Algorithm uint32
```
awly: Add doc comment

tpm/constants.go
```diff
@@ -123,6 +127,10 @@ const (
 	algXOR
 )
 
+func (a Algorithm) String() string {
+	return [...]string{"", "RSA", "", "", "SHA", "HMAC", "AES128", "MGF1", "AES192", "AES256", "XOR"}[a]
```
awly: This makes me a bit nervous about out-of-bounds panics when a new const is added.
How about using a map[Algorithm]string and handling unknown algs explicitly?

tpm/tpm.go
awly: Iterate over `[]Algorithm` or at least from `algRSA` to `algXOR`

tpm/constants.go
```diff
@@ -108,9 +110,11 @@ const (
 	pidTransport
 )
 
+type Algorithm uint32
+
 // Algorithm ID values.
 const (
-	_ uint32 = iota
+	_ Algorithm = iota
 	algRSA
```
awly: All of these consts should be exported

tpm/tpm_test.go
```diff
@@ -70,6 +70,61 @@ func TestGetManufacturer(t *testing.T) {
 	t.Logf("TPM VendorID: %v\n", vendorID)
 }
 
+func TestGetAlgs(t *testing.T) {
+	rwc := openTPMOrSkip(t)
+	defer rwc.Close()
+
+	algs, err := GetAlgs(rwc)
+	if err != nil {
+		t.Fatalf("Couldn't read Algorithms: %v\n", err)
```
awly: remove `\n`

tpm/tpm_test.go
```diff
@@ -70,6 +70,61 @@ func TestGetManufacturer(t *testing.T) {
...
+	}
+	for iterator, item := range algs {
```
awly: This looks wrong.
`GetAlgs` returns a list of Algorithms that are supported.
Here, the test code assumes that index of algs is the Algorithm identifier (it's not, it's just slice index) and the value is whether it's enabled (the value is Algorithm identifier).

You probably want something like:
```
want := []Algorithm{AlgRSA, AlgSHA, AlgHMAC, AlgMGF1}
outer:
for _, alg := range want {
  for _, got := range algs {
    if got == alg {
      continue outer
    }
    t.Errorf("GetAlgs returned %v, which does not contain a required algorithm %v", algs, alg)
  }
}
```

tpm/constants.go
ChriMarMe: All algs specified are already implemented as const here.

See: TPM-Main-Part-2-TPM-Structures_v1.2_rev116_01032011, Page 18.
 
So noone should get the idea of adding new algs as const, at least as long as he sticks to the specification.

tpm/constants.go
ChriMarMe: Agreed

tpm/tpm_test.go
ChriMarMe: Alright. I see my mistake.
Thank you :)

tpm/constants.go
awly: There's still possibility for user error:
```
id := someFuncThatReturnsUint32()
alg := tpm.Algorithm(id)
fmt.Println(alg) // This panics if id was not one of the consts
```

While that is clearly a bug in their code, there's no reason it should panic.

The solution you have here is clever, but clever is bad and risky.

tpm/tpm.go
```diff
@@ -988,6 +988,35 @@ func GetPermanentFlags(rw io.ReadWriter) (PermanentFlags, error) {
 	return ret, err
 }
 
+// GetAlgs returns a list of algorithms supported by the TPM device.
+func GetAlgs(rw io.ReadWriter) ([]Algorithm, error) {
+	var algs []Algorithm
+	for _, i := range []Algorithm{AlgRSA: AlgXOR} {
```
awly: This doesn't do what you think https://play.golang.org/p/7ilCvaudWxz
You're creating a slice of Algorithm, with all values being 0 except for value at index AlgRSA which is set to AlgXOR.
You want:
```
for i := AlgRSA; i <= AlgXOR; i++ {
```

ChriMarMe: I'm really thankful for your patience:)

tpm/constants.go
```diff
@@ -110,23 +112,38 @@ const (
...
+	AlgHMAC
+	AlgAES128
+	AlgMGF1
+	AlgAES192
+	AlgAES256
+	AlgXOR
 )
 
+// AlgMap Map of Algorithms to Strings for nicer output and comparisons, etc.
+var AlgMap = map[Algorithm]string{
```
awly: This should still be hidden via a String method:
```
var algNames = map[Algorithm]string{
	AlgRSA:    "RSA",
	AlgSHA:    "SHA1",
	AlgHMAC:   "HMAC",
	AlgAES128: "AER128",
	AlgMGF1:   "MFG1",
	AlgAES192: "AES192",
	AlgAES256: "AES256",
}

func (a Algorithm) String() string {
  n, ok := algNames[a]
  if !ok {
    return "unknown_algorithm"
  }
  return n
}

ChriMarMe: Hey, don't worry :) im really thankful for the reviewing and nit-picking is good, because I can improve my coding. :+1:

Issue #124, opened 2019-09-06
tpm2: Problem with expectedDigest in PolicyPCR


karandave-bc:
When a key is sealed against some PCR values without providing **expectedDigest** in the **PolicyPCR** to get the policy to seal with, everything works fine. Below is the example of it. (I am creating a sessionHandle and Policy digest as mentioned in examples of go-tpm.)
**err := tpm2.PolicyPCR(rwc, sessHandle, nil /*expectedDigest*/, pcrSelection)**

But when expected/predicted value of PCR is provided in PolicyPCR to get the new policy to seal against expected PCR value, it gives me error. Below is the example of it. I am using **tpm2.AlgSHA256**.
**err := tpm2.PolicyPCR(rwc, sessHandle, expectedDigest, pcrSelection)**

Here **expectedDigest** is **[ ] byte** type and I am getting it as a value of command line flag/argument. (Also tried defining with direct value in the function).

Is there anything that I am missing here or doing incorrectly?

Below is a little code sample of what I am doing.

`
func seal (c *cli) {
          pcrs_slice := c.IntSlice("pcrs")
          expectedDigest := c.String("digest")
          
           pcrs := make([]int, len(pcrs_slice))
           copy(pcrs[:], pcrs_slice)
     
           digest := []byte(nil)
           if expectedDigest != noDigest {
                   digest, _ = hex.DecodeString(expectedDigest)
           }

           fmt.Printf("Digest: ")
           fmt.Println(hex.EncodeToString(digest[:]))          
          
          rwc, err := tpm2.OpenTPM(device)

          sessHandle, policy, err := policyPCRPasswordSession(rwc, pcrs, digest, objectPassword)
          if err != nil {
                fmt.Fprintf(os.Stderr, "unable to get policy: %v", err)
                os.Exit(1)
          }

          /*** code for seal and load ***/
}
`
`
func policyPCRPasswordSession(rwc io.ReadWriteCloser, pcrs []int, expectedDigest []byte, password string) (sessHandle tpmutil.Handle, policy []byte, retErr error) {
        // this is not a very secure session.
        sessHandle, _, err := tpm2.StartAuthSession(
                rwc,
                tpm2.HandleNull,  /*tpmKey*/
                tpm2.HandleNull,  /*bindKey*/
                make([]byte, 16), /*nonceCaller*/
                nil,              /*secret*/
                tpm2.SessionPolicy,
                tpm2.AlgNull,
                tpm2.AlgSHA256)
        if err != nil {
[...]

PR #132, opened 2019-09-17
tpm2: add a test for PolicyPCR

awly:
Also, clarify what the expectDigest argument to PolicyPCR should be in
docs.

Updates #124

cc @karandave-bc

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -1478,3 +1478,58 @@ func TestPlainImport(t *testing.T) {
 		t.Errorf("Got %X, expected %X", outBuff, inBuff)
 	}
 }
+
+func TestPolicyPCR(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	sessHandle, _, err := StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionPolicy, AlgNull, AlgSHA1)
+	if err != nil {
+		t.Fatalf("StartAuthSession failed: %v", err)
+	}
+	defer FlushContext(rw, sessHandle)
+
+	sel := PCRSelection{Hash: AlgSHA1, PCRs: []int{0, 1, 2, 3}}
+	pcrs, err := ReadPCRs(rw, sel)
+	if err != nil {
+		t.Fatalf("ReadPCRs failed: %s", err)
+	}
+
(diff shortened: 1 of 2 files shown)


tpm2/tpm2.go
```diff
@@ -781,6 +781,12 @@ func encodePolicyPCR(session tpmutil.Handle, expectedDigest tpmutil.U16Bytes, se
 }
 
 // PolicyPCR sets PCR state binding for authorization on a session.
+//
+// Note that expectedDigest must be a *digest* of the expected PCR value. You
```
josephlr: We should note here that expected digest is optional. If you provide it, the digest is checked, if you don't provide it, the call will always succeed (as if you had passed the right digest).

tpm2/test/tpm2_test.go
```diff
@@ -1478,3 +1478,46 @@ func TestPlainImport(t *testing.T) {
 		t.Errorf("Got %X, expected %X", outBuff, inBuff)
 	}
 }
+
+func TestPolicyPCR(t *testing.T) {
```
josephlr: This is a great test. We should also add in a test for passing `nil` for the expected digest.

tpm2/tpm2.go

tpm2/test/tpm2_test.go
awly: Done

awly:
PTAL

tpm2/test/tpm2_test.go
```diff
@@ -1478,3 +1478,58 @@ func TestPlainImport(t *testing.T) {
...
+	t.Logf("expectedDigest=%x", expectedDigest)
+
+	tests := []struct {
+		desc           string
+		expectedDigest []byte
+		wantErr        bool
+	}{
+		// PolicyPCR expects a *digest* of the PCR value, not the value itself.
+		// Make sure PolicyPCR with value actually fails.
+		{desc: "success: digest passed in expectedDigest", expectedDigest: expectedDigest[:]},
```
josephlr: This is Good, but subtest descriptions are usually `CamelCase`, just like test names.

awly: Disagree, subtest descriptions are arbitrary text and `t.Run` normalizes them using few simple substitutions.
We've used camelCase in a few places here to match const/func names.

I'll merge without fixing this, but happy to discuss further offline.

josephlr: Np

PR #133, opened 2019-09-17
Add SignWithSession

BenDHillier:
SignWithSession is needed to support Signing Keys with a PCR policy.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -1289,12 +1289,12 @@ func Shutdown(rw io.ReadWriter, typ StartupType) error {
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


tpm2/tpm2.go
```diff
@@ -1327,10 +1327,10 @@ func decodeSign(buf []byte) (*Signature, error) {
 	return DecodeSignature(in)
 }
 
-// Sign computes a signature for digest using a given loaded key. Signature
+// SignWithSession computes a signature for digest using a given loaded key. Signature
```
josephlr: This comment shouldn't be exactly identical, at least explain the difference between it and `Sign`.

BenDHillier: Added more descriptive comment

PR #137, opened 2019-09-19
Update readme

awly:
Clean up outdated text.
Point out that 1.2 is not maintained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,32 @@
 Go-TPM
 ======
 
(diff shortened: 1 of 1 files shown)


josephlr:
This seems reasonable. @twitchy-jsonp, do you think we should say all 1.2 PRs will be rejected, or should we mention that you will review them?

Also, CC: @ChriMarMe who's been working on 1.2 PRs

PR #138, opened 2019-09-23
Add KDFe and create kdf helper function used by KDFa and KDFe.

BenDHillier:
KDFe implemented based on the specification outlined in section 11.4.9.3 of the TPM revision 2 specification part 1.

Test vectors were tested against the KDFe implementation in https://github.com/microsoft/ms-tpm-20-ref

Also changed the KDFa implementation to mask off excess bits, as described in the spec. 

" If bits is not an even multiple of 8, then the returned value occupies the least significant bits of the returned octet array, and the additional, high-order bits in the 0th octet are CLEAR. The unused bits of the most significant octet (MSO) are masked off and not shifted."


tpm2/test/kdf_test.go
```diff
@@ -70,3 +70,87 @@ func TestKDFa(t *testing.T) {
 		}
 	}
 }
+
+func TestKDFe(t *testing.T) {
+	hashAlg := AlgSHA256
```
josephlr: Would is be possible to get 1 test case for sha1 and one for sha256, instead of 2 for sha256

tpm2/kdf.go
```diff
@@ -68,3 +68,43 @@ func KDFa(hashAlg Algorithm, key []byte, label string, contextU, contextV []byte
...
+// The z parameter is the x coordinate of one parties private ECC key and the other parties public ECC key.
+// The partyUInfo and partyVInfo are the x coordinate of the initiators and the responders ECC points respectively.
+// Only the SHA256 hash algorithm is implemented at this time.
+func KDFe(hashAlg Algorithm, z []byte, use string, partyUInfo, partyVInfo []byte, bits int) ([]byte, error) {
+	var counter uint32
+	remaining := (bits + 7) / 8
+	var out []byte
+
+	var h hash.Hash
+	switch hashAlg {
```
josephlr: Can we use `tpm2.HashConstructor` here, instead of special-casing hash functions?

tpm2/kdf.go
```diff
@@ -68,3 +68,43 @@ func KDFa(hashAlg Algorithm, key []byte, label string, contextU, contextV []byte
...
+		return nil, fmt.Errorf("hash algorithm 0x%x is not supported", hashAlg)
+	}
+	for remaining > 0 {
+		counter++
+		if err := binary.Write(h, binary.BigEndian, counter); err != nil {
+			return nil, fmt.Errorf("pack counter: %v", err)
+		}
+		h.Write(z)
+		h.Write([]byte(use))
+		h.Write([]byte{0})
```
josephlr: Comment that this is null terminator

tpm2/kdf.go
```diff
@@ -68,3 +68,43 @@ func KDFa(hashAlg Algorithm, key []byte, label string, contextU, contextV []byte
...
+		h.Write(partyUInfo)
+		h.Write(partyVInfo)
+		out = h.Sum(out)
+		remaining -= h.Size()
+		h.Reset()
+	}
+
+	if len(out) > bits/8 {
```
josephlr: Why do we round down here as opposed to rounding up above?

tpm2/test/kdf_test.go
BenDHillier: Added, as well as some more edge cases

tpm2/kdf.go
BenDHillier: This strips excess bytes if the expected number of bytes is not a multiple of the hash digest size.

tpm2/kdf.go
```diff
@@ -28,43 +25,66 @@ import (
...
+	})
 	return out, nil
 }
+
+func kdf(h hash.Hash, bits int, update func()) []byte {
+	bytes := (bits + 7) / 8
+	out := []byte{}
+
+	counter := uint32(1)
+	for remaining := 0; remaining < bytes; remaining += h.Size() {
```
josephlr: Idea, lets change this to:

```go
for counter := 1; out(len) < bytes; counter += 1 {
  // Stuff 
  binary.Write(h, binary.BigEndian, uint32(counter))
  // Other stuff
}
```

Then we don't need `remaining`, `counter` stays in the loop, and the code is more readable.

tpm2/kdf.go
```diff
@@ -28,43 +25,66 @@ import (
...
+		h.Reset()
+		binary.Write(h, binary.BigEndian, counter)
+		update()
+		out = h.Sum(out)
+		counter++
+	}
+	// out's length is a multiple of hash size. If bytes isn't a multiple of hash size, strip excess.
```
josephlr: We don't need this `if`, just do `out = out[:bytes]`

tpm2/kdf.go
```diff
@@ -28,43 +25,66 @@ import (
...
+	if len(out) > bytes {
+		out = out[:bytes]
+	}
+	// If bits isn't a multiple of 8, mask off excess most significant bits from zeroth byte.
+	if bits%8 != 0 {
```
josephlr: This is slightly more readable if we do the `% 8` only once.

```go
if maskBits := uint8(bits % 8); maskBits > 0 {
  out[0] &= (1 << maskBits)  - 1
}
```

tpm2/kdf.go
```diff
@@ -28,43 +25,66 @@ import (
...
+}
 
-	if len(out) > bits/8 {
-		out = out[:bits/8]
+// KDFe implements TPM 2.0's ECDH key derivation function, as defined in
+// section 11.4.9.3 of the TPM revision 2 specification part 1.
+// See: https://trustedcomputinggroup.org/resource/tpm-library-specification/
+// The z parameter is the x coordinate of one parties private ECC key and the other parties public ECC key.
+// The partyUInfo and partyVInfo are the x coordinates of the initiators and the responders ECC points respectively.
+// The use parameter is a non-null-terminated string.
```
josephlr: Here and above, be sure to document parameters in their function order: `z`, `use`, etc...

josephlr:
@BenDHillier be sure to also update the description to talk about the `KDFa` changes as well.

CC: @twitchy-jsonp (who authored the original `KDFa` function) Do these changes look good to you? The test vectors all still pass. There is an update to the func to correctly handle masking unused high bits.

tpm2/kdf.go
```diff
@@ -28,43 +25,66 @@ import (
...
+func kdf(h hash.Hash, bits int, update func()) []byte {
+	bytes := (bits + 7) / 8
+	out := []byte{}
+
+	counter := uint32(1)
+	for remaining := 0; remaining < bytes; remaining += h.Size() {
+		h.Reset()
+		binary.Write(h, binary.BigEndian, counter)
+		update()
+		out = h.Sum(out)
```
josephlr: Add a newline before this, it helps space things out visually

tpm2/kdf.go
```diff
@@ -28,43 +25,66 @@ import (
...
+		counter++
+	}
+	// out's length is a multiple of hash size. If bytes isn't a multiple of hash size, strip excess.
+	if len(out) > bytes {
+		out = out[:bytes]
+	}
+	// If bits isn't a multiple of 8, mask off excess most significant bits from zeroth byte.
```
josephlr: Instead of describing _what_ we do, describe _why_ we do it. Something like:

`// Mask off excess bits if they weren't requested by the caller.`

Maybe add a spec reference (if there is one), or otherwise explain why this step is required.

PR #140, opened 2019-10-01
TPM 1.2: GetNVIndex to get information about a given NVIndex

ChriMarMe:
Add GetNVIndex-Function to read attributes of given NVIndex
Add TestGetNVIndex to test it
Add structures, constants and pretty print functions

### DIFF
--- a/tpm/constants.go
+++ b/tpm/constants.go
@@ -14,7 +14,11 @@
 
 package tpm
 
-import "github.com/google/go-tpm/tpmutil"
+import (
+	"fmt"
+	"github.com/google/go-tpm/tpmutil"
+	"strings"
+)
 
 // Supported TPM commands.
 const (
@@ -62,6 +66,7 @@ const (
 	capProperty uint32 = 0x00000005
 	capFlag     uint32 = 0x00000004
 	capNVList   uint32 = 0x0000000D
+	capNVIndex  uint32 = 0x00000011
 	capHandle   uint32 = 0x00000014
 )
 
@@ -70,6 +75,63 @@ const (
 	tpmCapPropManufacturer uint32 = 0x00000103
 	tpmCapFlagPermanent    uint32 = 0x00000108
 )
+
+// permission type
+type permission uint32
+
+// NV Permissions and Operations
+// Note: Permissions are summable
+const (
+	nvPerPPWrite    permission = 0x00000001
+	nvPerOwnerWrite permission = 0x00000002
+	nvPerAuthWrite  permission = 0x00000004
+	nvPerWriteAll   permission = 0x00000800
+	// Warning: The Value 0x00001000 is
+	// defined in the spec as
+	// TPM_NV_PER_WRITEDEFINE, but it is
+	// not included directly in this
+	// code because it locks the given
+	// NV Index permanently if used
+	// incorrectly. This operation can't
+	// be undone in any way. Do not use
+	// this value unless you know what
+	// you're doing!
+	nvPerWriteSTClear permission = 0x00002000
+	nvPerGlobalLock   permission = 0x00004000
+	nvPerPPRead       permission = 0x00008000
+	nvPerOwnerRead    permission = 0x00100000
+	nvPerAuthRead     permission = 0x00200000
+	nvPerReadSTClear  permission = 0x80000000
+)
+
+// permMap : Map of TPM_NV_Permissions to its strings for convenience
+var permMap = map[permission]string{
+	nvPerPPWrite:      "PPWrite",
+	nvPerOwnerWrite:   "OwnerWrite",
+	nvPerAuthWrite:    "AuthWrite",
+	nvPerWriteAll:     "WriteAll",
+	nvPerWriteSTClear: " WriteSTClear",
+	nvPerGlobalLock:   "GlobalLock",
+	nvPerPPRead:       "PPRead",
+	nvPerOwnerRead:    "OwnerRead",
+	nvPerAuthRead:     "AuthRead",
+	nvPerReadSTClear:  "ReadSTClear",
+}
+
+// String returns a textual representation of the set of permissions
+func (p permission) String() string {
+	var retString strings.Builder
+	for iterator, item := range permMap {
+		if (p & iterator) != 0 {
+			retString.WriteString(item + " + ")
+		}
+	}
+	if retString.String() == "" {
+		return "Permission/s not found"
+	}
+	return strings.TrimSuffix(retString.String(), " + ")
+
+}
 
 // Entity types. The LSB gives the entity type, and the MSB (currently fixed to
 // 0x00) gives the ADIP type. ADIP type 0x00 is XOR.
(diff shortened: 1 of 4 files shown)


tpm/constants.go
```diff
@@ -71,6 +72,62 @@ const (
...
+	nvPerWriteSTClear permission = 0x00002000
+	nvPerGlobalLock   permission = 0x00004000
+	nvPerPPRead       permission = 0x00008000
+	nvPerOwnerRead    permission = 0x00100000
+	nvPerAuthRead     permission = 0x00200000
+	nvPerReadSTClear  permission = 0x80000000
+)
+
+// PermMap : Map of TPM_NV_Permissions to its strings for convenience
+var PermMap = map[permission]string{
```
twitchy-jsonp: Make this lower-case so its not available as the package API. Users of our package should be able to get a textual representation of the permission by calling`String()` on the `Permission` object.

tpm/constants.go
```diff
@@ -71,6 +72,62 @@ const (
...
+	nvPerWriteAll:     "WriteAll",
+	nvPerWriteSTClear: " WriteSTClear",
+	nvPerGlobalLock:   "GlobalLock",
+	nvPerPPRead:       "PPRead",
+	nvPerOwnerRead:    "OwnerRead",
+	nvPerAuthRead:     "AuthRead",
+	nvPerReadSTClear:  "ReadSTClear",
+}
+
+// PrettyPrinting for Permissions
```
twitchy-jsonp: Nit: Doc comment strings should generally be full sentences, and in the case of a method or type, generally start with the method name or type. For instance, `String returns a human friendly representation of the set of permissions.`

tpm/constants.go
```diff
@@ -71,6 +72,62 @@ const (
...
+func (p permission) String() string {
+	var retString string
+	for iterator, item := range PermMap {
+		p = p - iterator
+		if p == 0 {
```
twitchy-jsonp: It seems like what you are trying to do here is test the bits to see if the permission is contained in `p`?

If so, the way you do that is like this:

```go
if (p & iterator) != 0 {
  // permission is set
}
```

And then return `retString` at the end.

Also, consider using `strings.Builder` from the standard library for this - its more idomatic and more efficient.

tpm/constants.go
```diff
@@ -92,6 +149,47 @@ const (
...
+	locThree: "Locality 3",
+	locFour:  "Locality 4",
+}
+
+// PrettyPrint for Localities
+func (l Locality) String() string {
+	var retString string
+	n, ok := LocaMap[l]
+	if !ok {
+		//In this case the given locality may a sum of localities
```
twitchy-jsonp: Same as above, consider using bitwise AND tests (`&`) and `strings.Builder` for this.

tpm/constants.go
```diff
@@ -92,6 +149,47 @@ const (
...
+		for iterator, item := range LocaMap {
+			l = l - iterator
+			if l == 0 {
+				retString += item
+				return retString
+			}
+			retString += item + " + "
+		}
+		return "unknown_locality"
```
twitchy-jsonp: In the case of unknown bits, it might be useful to return the raw value, like this: 

```go
return fmt.Sprintf("locality<%d>", int(l))
```

tpm/structures.go
```diff
@@ -109,6 +109,27 @@ type PermanentFlags struct {
 	DisableFullDALogicInfo       bool
 }
 
+// NVDataPublic implements the structure of TPM_NV_DATA_PUBLIC
```
twitchy-jsonp: Nit: Full sentences.

tpm/tpm.go
```diff
@@ -1028,6 +1028,20 @@ func GetNVList(rw io.ReadWriter) ([]uint32, error) {
...
+// information about the requested NV Index.
+// See: TPM-Main-Part-2-TPM-Structures_v1.2_rev116_01032011, P.167
+func GetNVIndex(rw io.ReadWriter, nvIndex uint32) (NVDataPublic, error) {
+	var nvInfo NVDataPublic
+	buf, err := getCapability(rw, capNVIndex, 0)
+	if err != nil {
+		return nvInfo, err
+	}
+	r := bytes.NewReader(buf)
+	err = binary.Read(r, binary.LittleEndian, &nvInfo)
```
twitchy-jsonp: Should this error be returned?

tpm/tpm.go
```diff
@@ -1028,6 +1028,20 @@ func GetNVList(rw io.ReadWriter) ([]uint32, error) {
...
 }
 
+// GetNVIndex returns the structure of nvDataPublic which contains
+// information about the requested NV Index.
+// See: TPM-Main-Part-2-TPM-Structures_v1.2_rev116_01032011, P.167
+func GetNVIndex(rw io.ReadWriter, nvIndex uint32) (NVDataPublic, error) {
+	var nvInfo NVDataPublic
+	buf, err := getCapability(rw, capNVIndex, 0)
+	if err != nil {
+		return nvInfo, err
```
twitchy-jsonp: Consider wrapping this error to give more context to the user:

```go
return nvInfo, fmt.Errorf("failed to get capability NVIndex: %v", err)
```

twitchy-jsonp: Consider wrapping this error to give more context to the user:

```go
return nvInfo, fmt.Errorf("failed to get capability NVIndex: %v", err)
```

ChriMarMe:
Would be nice to get some feedback today. I'm implementing NVDefineSpace and WriteValue so for the tests I need that function.

tpm/constants.go
```diff
@@ -71,6 +76,63 @@ const (
...
+		if (p & iterator) != 0 {
+			retString.WriteString(item + " + ")
+		}
+	}
+	if retString.String() == "" {
+		return "Permission/s not found"
+	}
+	return retString.String()
+
+}
```
twitchy-jsonp: Looks good!

One nit: it might not be ideal to always add the ` + ` to the end, because if you only have two permissions, you end up with this: `OwnerRead + AuthRead + `.

The easiest way to fix this would be:

```go
func (p permission) String() string {
	var retString strings.Builder
	for iterator, item := range permMap {
		if (p & iterator) != 0 {
			retString.WriteString(item + " + ")
		}
	}
	if retString.String() == "" {
		return "Permission/s not found"
	}
	return strings.TrimSuffix(retString.String(), " + ")
}
```

If you prefer, another pattern to build this string is by keeping a slice of the different pieces & joining them at the end using `strings.Join()`:

```go
// String returns a textual representation of the set of permissions
func (p permission) String() string {
	var out []string
	for iterator, item := range permMap {
		if (p & iterator) != 0 {
			out = append(out, item)
		}
	}
	if len(out) == 0 {
		return "Permission/s not found"
	}
	return strings.Join(out, " + ")
}
```

tpm/constants.go
```diff
@@ -92,6 +154,47 @@ const (
...
+
+// PrettyPrint for Localities
+func (l Locality) String() string {
+	var retString strings.Builder
+	n, ok := LocaMap[l]
+	if !ok {
+		//In this case the given locality may a sum of localities
+		for iterator, item := range LocaMap {
+			if l&iterator != 0 {
+				retString.WriteString(item + " + ")
```
twitchy-jsonp: Ditto comment above about avoiding a trailing ` + `.

tpm/constants.go
```diff
@@ -92,6 +154,47 @@ const (
...
+const (
+	locZero Locality = 1 << iota
+	locOne
+	locTwo
+	locThree
+	locFour
+)
+
+// LocaMap maps Locality values to strings for convenience
+var LocaMap = map[Locality]string{
```
twitchy-jsonp: Consider making this private, so the caller has to use `.String()` to get the string representation.

tpm/constants.go
```diff
@@ -92,6 +154,47 @@ const (
...
+				retString.WriteString(item + " + ")
+			}
+		}
+		if retString.String() == "" {
+			return fmt.Sprintf("locality %d", int(l))
+		}
+		return retString.String()
+	}
+	return n
+}
```
twitchy-jsonp: If only a single locality can be set at a time, you can simplify this to a switch clause instead:

```go
func (l Locality) String() string {
  switch l {
    case locZero:
      return "locality 0"
    case locOne:
      return "locality 1"
   // ...
  }
  return ""
}
```

ChriMarMe: More than one locality can be set at a time. 
Thank you for the fast review :)

twitchy-jsonp: In that case, can you get rid of
```go
n, ok := LocaMap[l]
if !ok {
  // but leave the code in here
}
```

And just have the bit within the `if !ok` clause? The code within that clause should handle all cases.

It should then look very similar to the `permission.String()` method.

ChriMarMe: Ok :)

PR #141, opened 2019-10-01
Explicitly define key type when generating pubkey

lulumel0n:
It will be easy to manipulate the public key later, since crypto.PublicKey is just an empty interface.


twitchy-jsonp:
Could you talk a little bit about the intent of this change?

Both before and after the change `Key()` will return a `crypto.PublicKey` value wrapping a pointer to the concrete RSA/ECDSA type. I would also expect the compiler to emit the same assembly after SSA lowering.

lulumel0n:
Never mind, I forget to dereference the pointer when doing the type assertion...

josephlr:
Closing this as it's not needed, see comments above.

PR #142, opened 2019-10-03
Add QuoteRaw() in tpm2

lulumel0n:
Add a new QuoteRawSignature() function in tpm2, which is very similar to Quote() except that it will return the Signature part as a raw byte array. This will allow to save the signature in a protobuf without Encoding the signature again. 

This will be used in google/go-tpm-tools#53

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -906,24 +906,42 @@ func encodeQuote(signingHandle tpmutil.Handle, parentPassword, ownerPassword str
 	return concat(ha, auth, params, pcrs)
 }
 
-func decodeQuote(in []byte) ([]byte, *Signature, error) {
+func decodeQuote(in []byte) ([]byte, []byte, error) {
 	buf := bytes.NewBuffer(in)
 	var paramSize uint32
 	var attest tpmutil.U16Bytes
 	if err := tpmutil.UnpackBuf(buf, &paramSize, &attest); err != nil {
 		return nil, nil, err
 	}
-	sig, err := DecodeSignature(buf)
-	return attest, sig, err
+	// the unread part of the buf is the raw signature
+	return attest, buf.Bytes(), nil
 }
 
 // Quote returns a quote of PCR values. A quote is a signature of the PCR
 // values, created using a signing TPM key.
 //
-// Returns attestation data and the signature.
+// Returns attestation data and the decoded signature.
 //
 // Note: currently only RSA signatures are supported.
 func Quote(rw io.ReadWriter, signingHandle tpmutil.Handle, parentPassword, ownerPassword string, toQuote []byte, sel PCRSelection, sigAlg Algorithm) ([]byte, *Signature, error) {
+	attest, sigRaw, err := QuoteRawSignature(rw, signingHandle, parentPassword, ownerPassword, toQuote, sel, sigAlg)
+	if err != nil {
+		return nil, nil, err
+	}
+	sig, err := DecodeSignature(bytes.NewBuffer(sigRaw))
+	if err != nil {
+		return nil, nil, err
+	}
+	return attest, sig, nil
+}
+
+// QuoteRawSignature is very similar to Quote, except that it will return
+// the raw signature in a byte array without decoding.
+//
+// Returns attestation data and the raw signature.
+//
+// Note: currently only RSA signatures are supported.
+func QuoteRawSignature(rw io.ReadWriter, signingHandle tpmutil.Handle, parentPassword, ownerPassword string, toQuote []byte, sel PCRSelection, sigAlg Algorithm) ([]byte, []byte, error) {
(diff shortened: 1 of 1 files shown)


twitchy-jsonp:
I've been doing

```go
rawSig, err := tpmutil.Pack(sig.Alg, sig.RSA.HashAlg, sig.RSA.Signature)
```

in go-attestation.

lulumel0n:
Currently I only see DecodeSignature(), shouldn't we add an EncodeSignature() and have these in it?

tpm2/tpm2.go
```diff
@@ -906,24 +906,42 @@ func encodeQuote(signingHandle tpmutil.Handle, parentPassword, ownerPassword str
...
+	}
+	return attest, sig, nil
+}
+
+// QuoteRawSignature is very similar to Quote, except that it will return
+// the raw signature in a byte array without decoding.
+//
+// Returns attestation data and the raw signature.
+//
+// Note: currently only RSA signatures are supported.
```
josephlr: The RSA-only support is no longer true, remove here and above.

tpm2/tpm2.go
```diff
@@ -906,24 +906,42 @@ func encodeQuote(signingHandle tpmutil.Handle, parentPassword, ownerPassword str
...
+		return nil, nil, err
+	}
+	return attest, sig, nil
+}
+
+// QuoteRawSignature is very similar to Quote, except that it will return
+// the raw signature in a byte array without decoding.
+//
+// Returns attestation data and the raw signature.
+//
```
josephlr: These lines aren't necessary (as it's just repating the above info).

tpm2/tpm2.go
```diff
@@ -906,24 +906,42 @@ func encodeQuote(signingHandle tpmutil.Handle, parentPassword, ownerPassword str
...
+// Note: currently only RSA signatures are supported.
+func QuoteRawSignature(rw io.ReadWriter, signingHandle tpmutil.Handle, parentPassword, ownerPassword string, toQuote []byte, sel PCRSelection, sigAlg Algorithm) ([]byte, []byte, error) {
```
josephlr: Just call it QuoteRaw

tpm2/tpm2.go
```diff
@@ -906,24 +906,42 @@ func encodeQuote(signingHandle tpmutil.Handle, parentPassword, ownerPassword str
...
+func decodeQuote(in []byte) ([]byte, []byte, error) {
 	buf := bytes.NewBuffer(in)
 	var paramSize uint32
 	var attest tpmutil.U16Bytes
 	if err := tpmutil.UnpackBuf(buf, &paramSize, &attest); err != nil {
 		return nil, nil, err
 	}
-	sig, err := DecodeSignature(buf)
-	return attest, sig, err
+	// the unread part of the buf is the raw signature
```
josephlr: I don't think this line is correct. I think we would have to use `paramSize` to figure out where the remaining buffer ends.

See Part 1, Section 18.1 pf the spec

tpm2/tpm2.go
lulumel0n: done

tpm2/tpm2.go
lulumel0n: done

tpm2/tpm2.go
lulumel0n: done

tpm2/tpm2.go
lulumel0n: Calculated the signature buf size based on paramSize and attestation buf size.

tpm2/tpm2.go
```diff
@@ -906,24 +906,36 @@ func encodeQuote(signingHandle tpmutil.Handle, parentPassword, ownerPassword str
...
+func decodeQuote(in []byte) ([]byte, []byte, error) {
 	buf := bytes.NewBuffer(in)
 	var paramSize uint32
 	var attest tpmutil.U16Bytes
 	if err := tpmutil.UnpackBuf(buf, &paramSize, &attest); err != nil {
 		return nil, nil, err
 	}
-	sig, err := DecodeSignature(buf)
-	return attest, sig, err
+	rawSig := buf.Next(int(paramSize) - len(attest))
```
josephlr: I think this might have an "off by 2 error" as the attest structure is a TPM2B_ATTEST which has a two byte size before the buffer, so this would give the wrong lengths.

The code we want is something like:

```go
buf := bytes.NewBuffer(in)
var paramSize uint32
if err := tpmutil.UnpackBuf(buf, &paramSize); err != nil {
	return nil, nil, err
}
buf.Truncate(paramSize)
var attest tpmutil.U16Bytes
if err := tpmutil.UnpackBuf(buf, &paramSize, &attest); err != nil {
	return nil, nil, err
}
return attest, buf.Bytes(), nil
```

lulumel0n: Make sense to me.
attest doesn't have a paramSize in front of it, so I removed the paramSize

PR #143, opened 2019-10-11
Add Hash() to tpm2.Algorithm which will return crypto.Hash

lulumel0n:
In some cases, I want to have a mapping between the hashalgo in tpm2 and crypto package. So I can easily reference to a crypto hashAlgo from a tpm2 Algo.

Currently using HashConstructor, we can get hash.Hash interface from a tpm2.Algorithm. But we couldn't get crypto.Hash from hash.Hash or tpm2.Algorithm

We need something similar to the go-attestation here in go-tpm so everyone can use it:
https://github.com/google/go-attestation/blob/25ce56400c9ff0a90a3f8b3eb47bf1698fe9319a/attest/tpm.go#L102b

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -17,14 +17,22 @@ package tpm2
 import (
 	"crypto"
 	"crypto/elliptic"
-	"crypto/sha1"
-	"crypto/sha256"
-	"crypto/sha512"
 	"fmt"
-	"hash"
+
+	// register hash algorithm
+	_ "crypto/sha1"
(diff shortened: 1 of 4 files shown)


josephlr:
This isn’t needed as it already exists. See https://godoc.org/github.com/google/go-tpm/tpm2#Algorithm.HashConstructor

josephlr:
@twitchy-jsonp this is a breaking change (but a small one), is this OK with you?

PR #144, opened 2019-10-15
TPM 1.2: NVDefineSpace

ChriMarMe:
- Add NVDefineSpace and TestNVDefineSpace
- Fixed all dependencies of localities

### DIFF
--- a/tpm/commands.go
+++ b/tpm/commands.go
@@ -158,6 +158,22 @@ func getCapability(rw io.ReadWriter, cap, subcap uint32) ([]byte, error) {
 	return b, nil
 }
 
+// nvDefineSpace allocates space in NVRAM
+func nvDefineSpace(rw io.ReadWriter, nvData NVDataPublic, enc digest, ca *commandAuth) (*responseAuth, uint32, error) {
+	var ra responseAuth
+	in := []interface{}{nvData, enc}
+	if ca != nil {
+		in = append(in, ca)
+	}
+	out := []interface{}{&ra}
+	ret, err := submitTPMRequest(rw, tagRQUAuth1Command, ordNVDefineSpace, in, out)
+	if err != nil {
+		return nil, 0, err
+	}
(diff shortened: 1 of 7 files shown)


ChriMarMe:
Running into the following problem when testing against swtpm:

Couldn't define space in NV RAM failed to define space in NVRAM: unable to read all contents in to U32Bytes

I can't figure out what the problem is. I checked all structs, fieldnames, ordinals, etc. 
Maybe someone with more knowledge can help me finding the bug.

tpm/commands.go
```diff
@@ -158,6 +158,23 @@ func getCapability(rw io.ReadWriter, cap, subcap uint32) ([]byte, error) {
...
 
+// nvDefineSpace allocates space in NVRAM
+func nvDefineSpace(rw io.ReadWriter, nvData NVDataPublic, enc digest, ca *commandAuth) ([]byte, *responseAuth, uint32, error) {
+	var b tpmutil.U32Bytes
+	var ra responseAuth
+	in := []interface{}{nvData, enc}
+	if ca != nil {
+		in = append(in, ca)
+	}
+	out := []interface{}{&b, &ra}
```
twitchy-jsonp: I think your issue is here. You are expecting a `U32Bytes` followed by the `responseAuth`, which would correspond to these outgoing operands:

 - uint32 (for the size of the data section)
 - []byte (for the variable data section)
 - nonceEven (part of `responseAuth`)
 - nonceOdd (part of `responseAuth`)
 - bool continueAuthSession (part of `responseAuth`)
 - authdata ownerAuth (part of `responseAuth` I think?)

My read of nvDefineSpace outgoing operands is that is only has the `responseAuth` outgoing parameters. As such, will it work if you do: `out := []interface{}{&ra}` and get rid of the `b` local?

ChriMarMe: Yes indeed good sir. Thank you very much!

PR #145, opened 2019-10-17
Add SealCertifiedWithPCR()

lulumel0n:
Current Seal() function only returns public and private. So it is impossible for user to certify a sealed blob with CertifyCreation().
I created a new function called SealCertifiedWithPCR() which will take in an additional parameter PCRselection, and return the creationData/hash and the ticket in addition to public and private, which can be used by CertifyCreation()

I also updated decodeCertify(), otherwise it would throw exception when sigAlg is AlgNull (it will try to unpack hashAlg and signature which doesn't exist in this case).

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -930,6 +930,66 @@ func TestCreateAndCertifyCreation(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


lulumel0n:
Closing. Will separate the CL, and part of this (CertifiedWithPCR()) will go to google/go-tpm-tools

PR #147, opened 2019-10-18
Fix tpmtool.poll function to build on arm64

### LINKED ISSUE #146: tpmutil does not compile with GOARCH=arm64
The error is `undefined: syscall.SYS_POLL`.

arm64 does not have `syscall.SYS_POLL`. Instead, the `syscall.SYS_PPOLL` syscall should be used. The functionality of the ppoll syscall is a superset of the poll syscall, so this should be relatively straightforward to fix.

rjoleary:
To achieve this, the underlying syscall is changed from poll to ppoll.
When timeout and sigmask are null, the behaviour of poll and ppoll are
equivalent (according to the man page).

Signed-off-by: Ryan O'Leary <ryanoleary@google.com>

Fixes #146

### DIFF
--- a/tpmutil/poll_unix.go
+++ b/tpmutil/poll_unix.go
@@ -5,38 +5,29 @@ package tpmutil
 import (
 	"fmt"
 	"os"
-	"syscall"
-	"unsafe"
+
+	"golang.org/x/sys/unix"
 )
 
-type pollFD struct {
-	fd      int32
-	events  int16
-	revents int16
-}
-
-// poll blocks until the file descriptior is ready for reading or an error occurs.
+// poll blocks until the file descriptor is ready for reading or an error occurs.
 func poll(f *os.File) error {
 	var (
-		fd = &pollFD{
-			fd:     int32(f.Fd()),
-			events: 0x1, // POLLIN
-		}
-		numFD     = 1
-		timeoutMS = -1 // Do not set a timeout
+		fds = []unix.PollFd{{
+			Fd:     int32(f.Fd()),
+			Events: 0x1, // POLLIN
+		}}
+		timeout = 0 // No timeout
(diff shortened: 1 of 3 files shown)


josephlr:
Thanks for the fix @rjoleary, but I think it might be easier to just use [`unix.Poll`](https://godoc.org/golang.org/x/sys/unix#Poll) for `x/sys/unix`. Apparently the stdlib `syscall` package has some known issues (like the one you found) and is "frozen", see https://golang.org/doc/go1.4. 

The `unix.Poll` function also is properly implemented across all Linux arches, it even has [this `arm64` workaround](https://github.com/golang/sys/blob/b09406accb4736d857a32bf9444cd7edae2ffa79/unix/syscall_linux_arm64.go#L204-L214) to make up for the fact that `arm64` does not have the `poll` syscall.

rjoleary:
Thanks josephlr! That's a lot better. I'll make that change.

rjoleary:
Done!

tpmutil/poll_unix.go
```diff
@@ -5,38 +5,29 @@ package tpmutil
...
 	)
-	_, _, errno := syscall.Syscall(syscall.SYS_POLL, uintptr(unsafe.Pointer(fd)), uintptr(numFD), uintptr(timeoutMS))
-	// Convert errno into an error, otherwise err != nil checks up the stack
-	// will hit unexpectedly on 0 errno.
-	var err error
-	if errno != 0 {
-		err = errno
+
+	_, err := unix.Poll(fds, timeout)
+	if err != nil {
```
josephlr: Merge these two lines

josephlr: Nevermind, I can just fix it.

PR #149, opened 2019-10-18
add CreateKeyWithSensitive() in tpm2

lulumel0n:
Currently, all exported Create*() functions don't support creating with "sensitive" data. Added a function to allow user to create with "sensitive" data.

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -876,6 +876,39 @@ func TestEncodeDecodePublicDefaultRSAExponent(t *testing.T) {
 	}
 }
 
+func TestCreateKeyWithSensitive(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+	parentHandle, _, err := CreatePrimary(rw, HandleOwner, pcrSelection7, emptyPassword, defaultPassword, Public{
+		Type:       AlgRSA,
+		NameAlg:    AlgSHA256,
+		Attributes: FlagRestricted | FlagDecrypt | FlagUserWithAuth | FlagFixedParent | FlagFixedTPM | FlagSensitiveDataOrigin,
+		RSAParameters: &RSAParams{
+			Symmetric: &SymScheme{
+				Alg:     AlgAES,
+				KeyBits: 128,
+				Mode:    AlgCFB,
+			},
+			KeyBits: 2048,
+		},
+	})
+	if err != nil {
+		t.Fatalf("CreatePrimary failed: %s", err)
+	}
+	defer FlushContext(rw, parentHandle)
+
+	secret := []byte("test_secret")
+	_, _, _, _, _, err = CreateKeyWithSensitive(rw, parentHandle, pcrSelection7, defaultPassword, defaultPassword, Public{
+		Type:       AlgKeyedHash,
+		NameAlg:    AlgSHA256,
+		Attributes: FlagFixedTPM | FlagFixedParent,
+		AuthPolicy: nil,
+	}, secret)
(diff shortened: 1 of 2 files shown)


josephlr:
This looks fine, but could we also add a test just to make sure invoking this function succeeds, nothing fancy. We should also rename to `CreateKeyWithSensitive` for consistancy with other names`.

Also, this will now leave us with 7 `Create`-like functions in `github.com/google/go-tpm/tpm2`:
  - `CreatePrimary`
  - `CreatePrimaryEx`
  - `CreatePrimaryRawTemplate`
  - `CreateKey`
  - `CreateKeyUsingAuth`
  - `CreateKeyWithSensitive`
  - `Seal`

We might want to consider unifying these before releasing `0.3`

twitchy-jsonp:
What do you envision as the unified API for `0.3` ?

josephlr:
> What do you envision as the unified API for `0.3` ?

A few things:
- Unifying and handle and it's authorization, so you can just pass one of 
   - A bare handle (will assume empty password, i.e. no auth)
   - A "password authenticated" handle, which is a handle+password struct
   - A "session authenticated" handle, which is a handle+session struct
   - Interfaces would make this nice
   - Reduces # of handles to a ton of calls.
- Having a common struct containing all of the creation data, so that we don't need 3 different methods
  - This `CreationData` struct would also allow code to better work with either `Create` or `CreatePrimary`
- Unifying the components of the SensitiveData structure, to better match the Spec

This would result in functions looking like:

```go
func CreatePrimary(
    rw io.ReadWriter,
    handle Handle,
    template Public,
    sensitive Sensitive,
    certify CertifyInput,
) (Handle, CreationData, error)

func Create(
    rw io.ReadWriter,
    handle Handle,
    template Public,
    sensitive Sensitive,
    certify CertifyInput,
) (CreationData, error)

// We would keep this alias for convenience, but remove the password stuff
func Create(
    rw io.ReadWriter,
    handle Handle,
    secret []byte,
    authPolicy []byte,
) ([]byte, []byte, error)
```

We could try to unify `Create` and `CreatePrimary` but that's probably too confusing.

tpm2/test/tpm2_test.go
```diff
@@ -876,6 +876,39 @@ func TestEncodeDecodePublicDefaultRSAExponent(t *testing.T) {
...
+
+	secret := []byte("test_secret")
+	_, _, _, _, _, err = CreateKeyWithSensitive(rw, parentHandle, pcrSelection7, defaultPassword, defaultPassword, Public{
+		Type:       AlgKeyedHash,
+		NameAlg:    AlgSHA256,
+		Attributes: FlagFixedTPM | FlagFixedParent,
+		AuthPolicy: nil,
+	}, secret)
+	if err != nil {
+		t.Fatalf("CreateKeyWithSensitive failed: %s", err)
```
josephlr: Errorf

tpm2/test/tpm2_test.go
```diff
@@ -876,6 +876,39 @@ func TestEncodeDecodePublicDefaultRSAExponent(t *testing.T) {
...
+			},
+			KeyBits: 2048,
+		},
+	})
+	if err != nil {
+		t.Fatalf("CreatePrimary failed: %s", err)
+	}
+	defer FlushContext(rw, parentHandle)
+
+	secret := []byte("test_secret")
```
josephlr: Move this inline, for consistancy

twitchy-jsonp:
LGTM on the proposed API. Let me know if theres any bits of that you want me to pick at.

PR #155, opened 2019-12-05
Add bounds checks for unpacks into U32/U16 bytes

### LINKED ISSUE #154: Should unpack of U16/U32Bytes validate a max size?
Malicious attacker-controlled input could make its way to `tpmutil.Unpack()` of `U32Bytes` or `U16Bytes`. For instance, calling `tpm.UnmarshalPubRSAPublicKey()` with `20ff ffff 2020 20ff ffff 2020` will cause a massive allocation, crashing the machine.

Given that `TPMUnmarshal` takes a reader, theres no upfront knowledge of the size of the buffer being processed.

Anyone have an opinion on these options?

1. Add a max size (maybe 1mb?) which is larger than all expected use-cases
2. Refactor `tpmutil.Unpack` to pass the size of the buffer down the chain so size prefix fields can be properly validated.

WDYT?

twitchy-jsonp:
Fixes #154

### DIFF
--- a/tpmutil/encoding_test.go
+++ b/tpmutil/encoding_test.go
@@ -364,3 +364,17 @@ func TestUnpackHandlesArea(t *testing.T) {
 		t.Errorf("Unpack(%v, %T): %T = %v, want %v", buf, &out, out, out, want)
 	}
 }
+
+func TestUnpackMalformedBytes(t *testing.T) {
+	// buf is malformed because the leading size prefix is illegally large.
+	buf := []byte{0xff, 0xff, 0xff, 0xff, 0x20}
+
+	var u32 U32Bytes
+	if _, err := Unpack(buf, &u32); err != bytes.ErrTooLarge {
+		t.Errorf("Unpack(U32Bytes{}) returned %q, want %q", err, bytes.ErrTooLarge)
+	}
+	var u16 U16Bytes
+	if _, err := Unpack(buf, &u16); err != io.ErrUnexpectedEOF {
+		t.Errorf("Unpack(U16Bytes{}) returned %q, want %q", err, io.ErrUnexpectedEOF)
+	}
+}
--- a/tpmutil/structures.go
+++ b/tpmutil/structures.go
@@ -15,10 +15,18 @@
 package tpmutil
 
 import (
+	"bytes"
 	"encoding/binary"
 	"fmt"
 	"io"
 )
+
+// maxBytesBufferSize sets a sane upper bound on the size of a U32Bytes
+// buffer. This limit exists to prevent a maliciously large size prefix
+// from resulting in a massive memory allocation, potentially causing
+// an OOM condition on the system.
+// We expect no buffer from a TPM to approach 1Mb in size.
+const maxBytesBufferSize = 1024 * 1024 // 1Mb.
 
 // RawBytes is for Pack and RunCommand arguments that are already encoded.
 // Compared to []byte, RawBytes will not be prepended with slice length during
@@ -52,6 +60,7 @@ func (b *U16Bytes) TPMUnmarshal(in io.Reader) error {
 		return err
 	}
 	size := int(tmpSize)
+
 	if len(*b) >= size {
 		*b = (*b)[:size]
 	} else {
@@ -63,7 +72,7 @@ func (b *U16Bytes) TPMUnmarshal(in io.Reader) error {
 		return err
 	}
 	if n != size {
-		return fmt.Errorf("unable to read all contents in to U16Bytes")
+		return io.ErrUnexpectedEOF
 	}
 	return nil
 }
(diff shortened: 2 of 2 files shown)


tpmutil/structures.go
```diff
@@ -52,6 +53,15 @@ func (b *U16Bytes) TPMUnmarshal(in io.Reader) error {
 		return err
 	}
 	size := int(tmpSize)
+
+	// If the reader is a bytes.Buffer, we check there are enough unread
+	// bytes to contain the data section indicated by the size prefix.
+	if bb, ok := in.(*bytes.Buffer); ok {
```
awly: Is this check necessary?
If remaining bytes in bytes.Buffer are fewer than size, in.Read below will fail with the same error

tpmutil/encoding_test.go
```diff
@@ -364,3 +364,16 @@ func TestUnpackHandlesArea(t *testing.T) {
 		t.Errorf("Unpack(%v, %T): %T = %v, want %v", buf, &out, out, out, want)
 	}
 }
+
+func TestUnpackMalformedBytes(t *testing.T) {
+	buf := []byte{0x20, 0xff, 0xff, 0xff, 0x20, 0x20, 0x20, 0xff, 0xff, 0xff}
```
awly: Add a comment explaining why these bytes are malformed

tpmutil/structures.go
```diff
@@ -95,6 +105,20 @@ func (b *U32Bytes) TPMUnmarshal(in io.Reader) error {
...
+		if size > bb.Len() {
+			return io.ErrUnexpectedEOF
+		}
+	} else {
+		// Otherwise, we make sure the size buffer is within sane bounds.
+		if size > (1024 * 1024) { // 1MB
```
awly: pull this limit out into a const

tpmutil/structures.go
```diff
@@ -95,6 +105,20 @@ func (b *U32Bytes) TPMUnmarshal(in io.Reader) error {
 		return err
 	}
 	size := int(tmpSize)
+
+	// If the reader is a bytes.Buffer, we check there are enough unread
+	// bytes to contain the data section indicated by the size prefix.
+	if bb, ok := in.(*bytes.Buffer); ok {
+		if size > bb.Len() {
+			return io.ErrUnexpectedEOF
+		}
```
josephlr: Similar to above, I also don't think we need this check here either, as bytes.Buffer.Read handles the `io.ErrUnexpectedEOF`. We should just unconditionally refuse to unmarshal if `size > 1MiB`.

tpmutil/structures.go
twitchy-jsonp: I can get rid of this, I just thought it would be more correct for us to check the buffer before allocating memory. This isn't technically necessary if we have the 1mb check further on.

tpmutil/encoding_test.go
```diff
@@ -364,3 +364,17 @@ func TestUnpackHandlesArea(t *testing.T) {
...
+func TestUnpackMalformedBytes(t *testing.T) {
+	// buf is malformed because the leading size prefix is illegally large.
+	buf := []byte{0xff, 0xff, 0xff, 0xff, 0x20}
+
+	var u32 U32Bytes
+	if _, err := Unpack(buf, &u32); err != bytes.ErrTooLarge {
+		t.Errorf("Unpack(U32Bytes{}) returned %v, want %v", err, bytes.ErrTooLarge)
+	}
+	var u16 U16Bytes
+	if _, err := Unpack(buf, &u16); err == nil || err.Error() != "unable to read all contents in to U16Bytes" {
```
awly: could you make this a package-level error var or return another known error (like `io.ErrUnexpectedEOF`)?
string comparison is a bit fragile.

tpmutil/encoding_test.go
```diff
@@ -364,3 +364,17 @@ func TestUnpackHandlesArea(t *testing.T) {
...
 	}
 }
+
+func TestUnpackMalformedBytes(t *testing.T) {
+	// buf is malformed because the leading size prefix is illegally large.
+	buf := []byte{0xff, 0xff, 0xff, 0xff, 0x20}
+
+	var u32 U32Bytes
+	if _, err := Unpack(buf, &u32); err != bytes.ErrTooLarge {
+		t.Errorf("Unpack(U32Bytes{}) returned %v, want %v", err, bytes.ErrTooLarge)
```
awly: nit: use `%q` for both returned and want errors

tpmutil/structures.go
```diff
@@ -15,11 +15,14 @@
 package tpmutil
 
 import (
+	"bytes"
 	"encoding/binary"
 	"fmt"
 	"io"
 )
 
+const maxBytesBufferSize = 1024 * 1024 // 1Mb.
```
awly: Add a comment explaining why this limit exists

PR #159, opened 2020-01-14
Adds functions tpm2_clear and tpm2_hierarchychangeauth

neeki-hushyar:
Adds functions to support commands tpm2_clear and tpm2_hierarchychangeauth. Together these commands can be used to take ownership of a tpm.

Using command codes found [here](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p55_Part2_Structures_pubrev.pdf#page=47)

Using parameters found [here](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf)

Tests in tpm2_test.go

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -1511,6 +1511,43 @@ func TestPlainImport(t *testing.T) {
 	}
 }
 
+func TestClear(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	err := Clear(rw, HandleLockout, AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession})
+	if err != nil {
+		t.Fatalf("Clear failed: %v", err)
+	}
+}
+
+func TestHierarchyChangeAuth(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	err := Clear(rw, HandleLockout, AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession})
+	if err != nil {
+		t.Fatalf("Clear failed: %v", err)
+	}
+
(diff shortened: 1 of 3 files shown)


neeki-hushyar:
@googlebot I signed it!

tpm2/test/tpm2_test.go
```diff
@@ -1511,6 +1511,43 @@ func TestPlainImport(t *testing.T) {
 	}
 }
 
+func TestClear(t *testing.T) {
```
twitchy-jsonp: IIRC, this test will mess up a machine if it is run inadvertently?

If so, can you add a new flag and use that to hide running this test?

tpm2/test/tpm2_test.go
```diff
@@ -1511,6 +1511,43 @@ func TestPlainImport(t *testing.T) {
...
+	rw := openTPM(t)
+	defer rw.Close()
+
+	err := Clear(rw, HandleLockout, AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession})
+	if err != nil {
+		t.Fatalf("Clear failed: %v", err)
+	}
+}
+
+func TestHierarchyChangeAuth(t *testing.T) {
```
twitchy-jsonp: Likewise for this test.

tpm2/tpm2.go
```diff
@@ -1060,6 +1060,57 @@ func EvictControl(rw io.ReadWriter, ownerAuth string, owner, objectHandle, persi
...
+	ah, err := tpmutil.Pack(handle)
+	if err != nil {
+		return nil, err
+	}
+	encodedAuth, err := encodeAuthArea(auth)
+	if err != nil {
+		return nil, err
+	}
+
+	newEncodedAuth, err := encodeAuthArea(newAuth)
```
twitchy-jsonp: Please check/handle the error value here as well.

tpm2/test/tpm2_test.go
neeki-hushyar: That's a good point - added flag which will skip test if not set

twitchy-jsonp:
Looks good! Please sync with head (looks like a conflict in the command code table from the PR i just merged) and we should be good to go.

Issue #139, opened 2019-09-24
Application for TPM1.2 Maintainership


ChriMarMe:
Hello everyone,

Our company "9elements Cyber Security" https://9esec.io would like to apply for TPM1.2 maintainership.

Especially following team members @zaolin , @flanfly , @ChriMarMe 

Best Regards,
Christopher Meis

ChriMarMe:
Hey everyone,

some feedback to the application would be nice.
We really want to step in here.

Best Regard
@ChriMarMe

awly:
Thanks for volunteering!
cc @josephlr @twitchy-jsonp @stevenrutherford 

We've agreed to accept your application, with a temporary review shadowing setup:
- after you review and approve a PR, assign one of the current maintainers
- this maintainer will do a second pass on the PR
- after the second reviewer approves a PR, it can be merged

_This ensures we keep a consistent style and structure._

After some time (a few months), when second reviews result in few comments, we'll step back and let you review PRs without additional oversight.

I will put in a [CODEOWNERS](https://help.github.com/en/articles/about-code-owners) file so it's easier to find other maintainer usernames.

**Do you agree to this setup?**

josephlr:
@ChriMarMe we generally use GitHub groups to mange who is admin/writer/etc... is there a specific group from your org that you would want to use here? If you don’t have one, that’s OK, we can just manually add a list of usernames.

ChriMarMe:
@awly I agree to this setup

@josephlr We belong to the "Cyber Security" group of our organisation at github, but only the three people mentioned above will do the reviewing. So I suggest you just add us manually.

awly:
@josephlr sounds like understand github permissions and `CODEOWNERS` well.
Do you mind setting up access for 9elements?

josephlr:
Will do.

zaolin:
Any updates so far?

awly:
Ping @josephlr

hugelgupf:
Hey @josephlr would you mind picking this up again? I'm happy to talk to you internally as well about this.

ChriMarMe:
Hey, whats the status about this?

josephlr:
Hey everyone, very sorry for the delay, this totally fell off my plate, I'm getting all of the groups setup now.

hugelgupf:
Thank you!

josephlr:
[...]

PR #161, opened 2020-01-17
Add CODEOWNERS file

josephlr:
This allows us to have good defaults while allowing different users to
handle TPM 1.2 and TPM 2.0 code.

CC: @zaolin @ChriMarMe @flanfly 

Addresses #139 

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,11 @@
+# Being an admin also gives you global code ownership.
+*                   @google/go-tpm-admin
+
+# Don't add all the admins by default for most things.
+# Note that any admins can still approve any PR.
+*                   @josephlr @awly @twitchy-jsonp
+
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,11 @@
+# Being an admin also gives you global code ownership.
+*                   @google/go-tpm-admin
+
+# Don't add all the admins by default for most things.
+# Note that any admins can still approve any PR.
+*                   @josephlr @awly @twitchy-jsonp
```
awly: This is a subset of go-tpm-admin.
Can we just remove this line?

josephlr: That line doesn't change ownership, but it does change the default attention set on a PR. This prevents every admin from getting automatically added to every PR.

See: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax

PR #162, opened 2020-01-22
TPM2: tpm2.GetCapability uses TPMProp type now.

ChriMarMe:


twitchy-jsonp:
Looks good except this is a breaking API change, so I/we need to bump the major version of the go module.

Also, looks like a test broke:

```
# github.com/google/go-tpm/tpm2/test [github.com/google/go-tpm/tpm2/test.test]
tpm2/test/tpm2_test.go:369:60: cannot use i (type uint32) as type tpm2.TPMProp in argument to tpm2.GetCapability
FAIL	github.com/google/go-tpm/tpm2/test [build failed]
```

ChriMarMe:
Alright then. Maybe I find the time to find a solution to fix it in the long run. 
As long as I have no solution I close the pr and open a new one if i find a proper fix.

PR #169, opened 2020-03-20
export tpm.getCapability

chrisfenner:
This change exports getCapability so that arbitrary capabilities may be fetched.

### DIFF
--- a/tpm/commands.go
+++ b/tpm/commands.go
@@ -143,21 +143,6 @@ func getPubKey(rw io.ReadWriter, keyHandle tpmutil.Handle, ca *commandAuth) (*pu
 	return &pk, &ra, ret, nil
 }
 
-// getCapability reads the requested capability and sub-capability from NVRAM
-func getCapability(rw io.ReadWriter, cap, subcap uint32) ([]byte, error) {
-	subCapBytes, err := tpmutil.Pack(subcap)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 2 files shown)


ChriMarMe:
getCapabilities is the only function where we try to implement every subCap as an exported function, so the user don't need to take care of the returned bytearray, if there is one.

It's just a question of who should take care of the returned bytearray of getCapabilities.

chrisfenner:
Thanks for taking a look Christopher,

The [latest version](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Main-Part-2-TPM-Structures_v1.2_rev116_01032011.pdf) of the TPM 1.2 spec defines 20 capabilities, and the current TPM 1.2 library supports 6 of them.

Until the remaining 14 are implemented, it would be good to have something callers can use if they need a capability which is not explicitly handled here.

I totally agree that the raw byte slice here is more difficult to handle. What if I renamed this function to GetCapabilityRaw and added documentation encouraging callers to use the "nice" versions exposed when available?

ChriMarMe:
Just wrap the getCapability-Function with GetCapabilityRaw, hence they take the same arguments, you can just pass them through. Also good idea with the documentation.

chrisfenner:
SGTM, I've renamed to GetCapabilityRaw and passed the args through to getCapability, which I've left unchanged (aside from a wording fix to the documentation - not all capabilities are information fetched from NVRAM)

PR #176, opened 2020-06-12
TPM12: Fix nvReadValue 

ChriMarMe:
now works with no authentication
fix GetNVIndex, GetNVList
rename TestNVReadValue to TestNVReadValueNoAuth

### DIFF
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -1152,31 +1152,28 @@ func GetAlgs(rw io.ReadWriter) ([]Algorithm, error) {
 // GetNVList returns a list of TPM_NV_INDEX values that
 // are currently allocated NV storage through TPM_NV_DefineSpace.
 func GetNVList(rw io.ReadWriter) ([]uint32, error) {
-	var nvList []uint32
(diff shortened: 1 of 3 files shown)


tpm/tpm.go
```diff
@@ -1152,31 +1152,28 @@ func GetAlgs(rw io.ReadWriter) ([]Algorithm, error) {
...
-	var nvList []uint32
 	buf, err := getCapability(rw, capNVList, 0)
 	if err != nil {
 		return nil, err
 	}
-	r := bytes.NewReader(buf)
-	err = binary.Read(r, binary.LittleEndian, &nvList)
+	nvList := make([]uint32, len(buf)/4)
+	for i := range nvList {
+		nvList[i] = uint32(binary.BigEndian.Uint32(buf[i*4 : (i+1)*4]))
```
zaolin, line 1161: Can we solve this in a better way here?

ChriMarMe: I tried and it failed. The binary.Read() just returned a empty buffer. Couldn't figure out why.

PR #178, opened 2020-06-21
Enable signing arbitrary data with attestation keys

### LINKED ISSUE #101: tpm2: Sign with hash ticket
The TPM2_Sign command includes a validation hashcheck ticket parameter, but this isn't exposed via the tpm2.Sign function API. This is relevant to restricted keys (i.e., tpm2.FlagRestricted, which is included in tpm2.FlagSignerDefault), because they need to make sure the hashed data did not start with TPM_GENERATED_VALUE.

chrisfenner:
Fixes #101 

This change update two functions in tpm2:

Hash now accepts the `hierarchy` parameter to TPM2_Hash, and returns `validation`.

Sign now accepts the `validation` parameter to TPM2_Sign. This allows a restricted signing key to sign arbitrary data (that does not begin with TPM_GENERATED_VALUE) if Hash is used to allow the TPM to perform this check on the data.

This change also adds warnings to the Hash function, pointing out that amounts of data over 1,024 bytes may not work on all TPMs.

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -30,6 +30,7 @@ import (
 	"testing"
 
 	"github.com/google/go-tpm-tools/simulator"
+	"github.com/google/go-tpm/tpm2"
 	. "github.com/google/go-tpm/tpm2"
 	"github.com/google/go-tpm/tpmutil"
 )
@@ -359,6 +360,38 @@ func TestHash(t *testing.T) {
 	if !bytes.Equal(got, want[:]) {
 		t.Errorf("Hash(%q) returned %x, want %x", val, got, want)
 	}
+}
+
+func TestHashVerified(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	run := func(t *testing.T, data []byte, hierarchy tpmutil.Handle, wantValidation bool) {
+		gotDigest, gotValidation, err := HashVerified(rw, AlgSHA256, data, hierarchy)
+		if err != nil {
+			t.Fatalf("HashVerified failed: %v", err)
+		}
+		wantDigest := sha256.Sum256(data)
+
+		if !bytes.Equal(gotDigest, wantDigest[:]) {
+			t.Errorf("HashVerified(%q) returned digest %x, want %x", data, gotDigest, wantDigest)
+		}
+		if wantValidation && len(gotValidation.Digest) == 0 {
+			t.Errorf("HashVerified(%q) unexpectedly returned empty validation ticket", data)
+		}
+		if !wantValidation && len(gotValidation.Digest) != 0 {
+			t.Errorf("HashVerified(%q) unexpectedly returned non-empty validation ticket", data)
+		}
+	}
+	t.Run("Null hierarchy", func(t *testing.T) {
+		run(t, []byte("foobarbaz"), tpm2.HandleNull, false)
+	})
+	t.Run("Owner hierarchy", func(t *testing.T) {
+		run(t, []byte("foobarbaz"), tpm2.HandleOwner, true)
+	})
+	t.Run("Starts with TPM_GENERATED_VALUE", func(t *testing.T) {
+		run(t, []byte("\xffTCGbarbaz"), tpm2.HandleOwner, false)
+	})
 }
 
 func skipOnUnsupportedAlg(t testing.TB, rw io.ReadWriter, alg Algorithm) {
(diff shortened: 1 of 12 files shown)


twitchy-jsonp:
Looking good!

With regards to changing the API vs overloading methods with longer names: We now have Go modules, so it should be fine to introduce breaking changes along with a version bump. I'll leave that decision to someone knowledgeable though :)

 @josephlr thoughts?

tpm2/test/tpm2_test.go
```diff
@@ -700,6 +732,114 @@ func TestSign(t *testing.T) {
...
 }
 
+func TestSignVerifiedHash(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	run := func(t *testing.T, data []byte, wantErr bool, pub Public) {
+		signerHandle, signerPub, err := CreatePrimary(rw, HandleOwner, pcrSelection7, emptyPassword, defaultPassword, pub)
+		if err != nil {
+			t.Fatalf("CreatePrimary failed: %s", err)
```
twitchy-jsonp, line 742: Did you mean to use `%v` here?

josephlr:
I talked with @chrisfenner offline, I think that breaking the API here is fine (given that go-modules gives us an easy way to do this). I think we should modify the `Hash` and `Sign` functions to better reflect the actual parameters passed to these TPM commands.

We can then add a high-level function like the following to `go-tpm-tools`:
```go
func (k *Key) SignData([]byte data) (signature []byte, err error) {
  // Call tpm2.Hash and tpm2.Sign with the ticket params
}
```

If we do a breaking change, then we might want to do additional cleanup. I'll leave discussion of that for #180

tpm2/test/tpm2_test.go
chrisfenner: Good catch, thank you!

tpm2/tpm2.go
```diff
@@ -1447,8 +1464,14 @@ func decodeSign(buf []byte) (*Signature, error) {
...
 // algorithm depends on the key type. Used for keys with non-password authorization policies.
-func SignWithSession(rw io.ReadWriter, sessionHandle, key tpmutil.Handle, password string, digest []byte, sigScheme *SigScheme) (*Signature, error) {
-	cmd, err := encodeSign(sessionHandle, key, password, digest, sigScheme)
+// If 'key' references a Restricted Decryption key, 'validation' must be a valid hash verification
+// ticket from the TPM, which can be obtained by using Hash() to hash the data with the TPM.
+// If 'validation' is nil, a NULL ticket is passed to TPM2_Sign.
+func SignWithSession(rw io.ReadWriter, sessionHandle, key tpmutil.Handle, password string, digest []byte, sigScheme *SigScheme, validation *Ticket) (*Signature, error) {
+	if validation == nil {
+		validation = &nullTicket
+	}
```
josephlr, line 1473: Nit: have `encodeSign` take a `*Ticket` instead of a `Ticket`, then handle the `nil` encoding in `encodeSign`. This makes the tests easier to read, and avoids needing `nullTicket` to be declared outside of `encodeSign`.

josephlr: It's also more consistent with how we handle `SigScheme`.

tpm2/tpm2.go
```diff
@@ -1461,8 +1484,11 @@ func SignWithSession(rw io.ReadWriter, sessionHandle, key tpmutil.Handle, passwo
 
 // Sign computes a signature for digest using a given loaded key. Signature
 // algorithm depends on the key type.
-func Sign(rw io.ReadWriter, key tpmutil.Handle, password string, digest []byte, sigScheme *SigScheme) (*Signature, error) {
-	return SignWithSession(rw, HandlePasswordSession, key, password, digest, sigScheme)
+// If 'key' references a Restricted Decryption key, 'validation' must be a valid hash verification
+// ticket from the TPM, which can be obtained by using Hash() to hash the data with the TPM.
+// If 'validation' is nil, a NULL ticket is passed to TPM2_Sign.
+func Sign(rw io.ReadWriter, key tpmutil.Handle, password string, digest []byte, sigScheme *SigScheme, validation *Ticket) (*Signature, error) {
```
josephlr, line 1490: Bikeshed on ordering here, `Hash` returns `(digest []byte, validation *Ticket)`, so should we have `Sign` take the parameters like: `digest []byte, validation *Ticket, sigScheme *SigScheme`

tpm2/test/tpm2_test.go
```diff
@@ -700,6 +716,119 @@ func TestSign(t *testing.T) {
...
+		}
+		defer FlushContext(rw, signerHandle)
+
+		var scheme *SigScheme
+		if pub.RSAParameters != nil {
+			scheme = pub.RSAParameters.Sign
+		}
+		if pub.ECCParameters != nil {
+			scheme = pub.ECCParameters.Sign
+		}
```
josephlr, line 736: Could we just use a `nil` scheme here to make things more simple?

chrisfenner: Good catch, agreed!

tpm2/tpm2.go
chrisfenner: SGTM. I had originally had this order, then swapped it to match the TPM spec, but I agree that this order makes more sense

tpm2/tpm2.go
chrisfenner: Good call, thanks!

tpm2/test/tpm2_test.go
josephlr: I think making this scheme `nil` got missed.

chrisfenner: whoops, fixed one test but not the other. Sorry about this!

PR #184, opened 2020-06-24
TPM12: GetPermanentFlags returns []byte, 

ChriMarMe:
handling of the raw data must be taken care of by user

### DIFF
--- a/tpm/tpm.go
+++ b/tpm/tpm.go
@@ -1121,16 +1121,12 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
 }
 
 // GetPermanentFlags returns the TPM_PERMANENT_FLAGS structure.
-func GetPermanentFlags(rw io.ReadWriter) (PermanentFlags, error) {
-	var ret PermanentFlags
-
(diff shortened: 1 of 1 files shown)


tpm/tpm.go
```diff
@@ -1121,16 +1121,12 @@ func GetManufacturer(rw io.ReadWriter) ([]byte, error) {
...
-	var ret PermanentFlags
-
+func GetPermanentFlags(rw io.ReadWriter) ([]byte, error) {
 	raw, err := getCapability(rw, capFlag, tpmCapFlagPermanent)
 	if err != nil {
-		return ret, err
+		return nil, err
 	}
-
-	_, err = tpmutil.Unpack(raw, &ret)
```
zaolin, line 1132: Unpack isn't needed anymore?

chrisfenner:
If the caller needs the raw data, why not use [GetCapabilityRaw](https://github.com/google/go-tpm/blob/b2eaffd5ed5b23d962b9e628412da225251cf248/tpm/tpm.go#L1182)?

PR #190, opened 2020-07-15
TPM2: Add missing object attribute stClear

ElMostafaIdrassi:
Its valus is 0x00000004 as specified [here](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf), page 63. It allows for the creation of objects that will have their contexts invalidated at each reboot.


ElMostafaIdrassi:
@googlebot I signed it!

Issue #191, opened 2020-07-17
TPM2: DecodeSignature fails to decode CertifyCreation signature by an ECC key


ElMostafaIdrassi:
I've added a [test](https://github.com/ElMostafaIdrassi/go-tpm/commit/fdcd1b95dfa7b0de7f44c9fc6f53128c8666f3ee) for this particular case, as I found `TestCreateAndCertifyCreation` only deals with RSA keys.

As described in the commit, the test fails because the signature does not seem to be valid, mainly because of its small length (32 bytes during my tests).
It should normally consist of both S and R which both need to be, in this test case, 32-bytes long.

**This was tested using the simulator and a real TPM 2.0 device, under Linux.**

I'm providing the test case here as well : 

```
func TestCreateAndCertifyCreationECC(t *testing.T) {
	rw := openTPM(t)
	defer rw.Close()
	params := tpm2.Public{
		Type:       tpm2.AlgECC,
		NameAlg:    tpm2.AlgSHA256,
		Attributes: tpm2.FlagSignerDefault,
		ECCParameters: &tpm2.ECCParams{
			Sign: &tpm2.SigScheme{
				Alg:  tpm2.AlgECDSA,
				Hash: tpm2.AlgSHA256,
			},
			CurveID: tpm2.CurveNISTP256,
		},
	}
	keyHandle, pub, _, creationHash, tix, _, err := CreatePrimaryEx(rw, HandleEndorsement, pcrSelection7, emptyPassword, emptyPassword, params)
	if err != nil {
		t.Fatalf("CreatePrimaryEx failed: %s", err)
	}
	defer FlushContext(rw, keyHandle)

	scheme := SigScheme{Alg: AlgECDSA, Hash: AlgSHA256, Count: 0}
	attestation, signature, err := CertifyCreation(rw, emptyPassword, keyHandle, keyHandle, nil, creationHash, scheme, tix)
	if err != nil {
		t.Fatalf("CertifyCreation failed: %s", err)
	}
	att, err := DecodeAttestationData(attestation)
	if err != nil {
		t.Fatalf("DecodeAttestationData(%v) failed: %v", attestation, err)
	}
	if att.Type != TagAttestCreation {
		t.Errorf("Got att.Type = %v, want TagAttestCreation", att.Type)
	}
	p, err := DecodePublic(pub)
	if err != nil {
		t.Fatalf("DecodePublic failed: %v", err)
	}
[...]

PR #192, opened 2020-07-18
Fix `decodeCertify` always returning a truncated signature in case of `ECDSA` signature.

ElMostafaIdrassi:
This PR is a follow up to the issue #191.

The bug is mainly due to the fact that the old code called `tpmutil.UnpackBuf` on only one element that is `signature`.
It worked for `RSA` signatures as their corresponding structure `TPMS_SIGNATURE_RSASSA` contains only 1 field `TPM2B_PUBLIC_KEY_RSA`, in addition to `TPMI_ALG_HASH` field.
However, it failed for `ECDSA` signatures because their corresponding structure is `TPMS_SIGNATURE_ECDSA`,
which contains 2 fields R and S - instead of just 1 - which are both `TPM2B_ECC_PARAMETER`, in addition to `TPMI_ALG_HASH` field.

All this information can be found [here](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.07-2014-03-13.pdf).

Therefore, the bug can be fixed by calling `tpmutil.UnpackBuf` on 2 elements, that is R followed by S, in the case of `ECDSA` signature, which results in retrieving the whole signature.


ElMostafaIdrassi:
Any chance this, #190 and #193 can get merged ?

PR #197, opened 2020-07-23
Fix missing error checks, typos

brandonweeks:


tpm2/tpm2.go
```diff
@@ -1753,6 +1756,9 @@ func encryptDecryptBlockSymmetric(rw io.ReadWriteCloser, keyAuth string, key tpm
 				return nil, nil, err
 			}
 			resp, err = runCommand(rw, TagSessions, CmdEncryptDecrypt, tpmutil.RawBytes(Cmd))
+			if err != nil {
```
twitchy-jsonp, line 1759: This one should be covered by the guard clause outside the loop, right?

brandonweeks: Fixed to be true to the intent of the code per the comment.

Issue #202, opened 2020-08-01
ReadPCRs returns a maximum of 8 results


iolivergithub:
Here is the code snipped

```go
    resp,err := tpm2.ReadPCRs(rwc,tpm2.PCRSelection{ tpm2.AlgSHA256 , []int{10,11,12,13,14,15,16,17,18,19,20} })

        fmt.Println("Err is *",err,"*")
        fmt.Println("length  ",len(resp))
        fmt.Println("resp  ",resp)
```

The result is:

```bash
Err is * <nil> *
length   8
resp   map[10:[233 78 123 248 213 96 73 135 44 139 238 32 208 223 134 159 56 215 159 209 85 238 17 216 216 249 145 194 75 85 102 54] 11:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] 12:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] 13:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] 14:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] 15:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] 16:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] 17:[255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255]]
```

Same result, ie maximum of 8 PCRs returned for any combination of PCRs and PCR bank.

ElMostafaIdrassi:
[...]

PR #203, opened 2020-08-02
TPM2: Fixed ReadPCRs() only returning the first 8 PCRs digests when more than 8 PCRs are selected.

ElMostafaIdrassi:
This is a follow up to issue #202.

The bug was due to the fact that the old code called the `TPM2_PCR_Read` command only once. This worked fine when 8 PCRs at most were selected, but returned only the first 8 PCRs digests if more. This can be explained by the fact that one single `TPM2_PCR_Read` command can only return 1 single `TPML_DIGEST`, which consists of, at most, 8 digests. Therefore, if we select PCRs `{0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20}`, one call to `TPM2_PCR_Read` would only return the digests of PCRs `{0,1,2,3,4,5,6,7}`.

The bug is fixed by calling the `TPM2_PCR_Read` command as many times as needed, and skipping, at each call, the 8 PCRs that were selected by the previous call.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -178,16 +178,29 @@ func decodeReadPCRs(in []byte) (map[int][]byte, error) {
 
 // ReadPCRs reads PCR values from the TPM.
 func ReadPCRs(rw io.ReadWriter, sel PCRSelection) (map[int][]byte, error) {
-	Cmd, err := encodeTPMLPCRSelection(sel)
-	if err != nil {
-		return nil, err
+	ret := make(map[int][]byte)
+	for i := 0; i <= len(sel.PCRs)/8; i++ {
+		selPCRs := sel.PCRs[i*8:]
+		if len(selPCRs) == 0 {
+			break
+		}
+		Cmd, err := encodeTPMLPCRSelection(tpm2.PCRSelection{Hash: sel.Hash, PCRs: selPCRs})
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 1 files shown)


twitchy-jsonp:
Given go-tpm at this layer is meant to be a mild shim over the TPM commands, I think we want to keep the existing behavior and let callers of `ReadPCRs()` call the command in a loop.

twitchy-jsonp:
Perhaps we can update the function doc to make it clear this command only wraps `TPM2_PCR_Read`, and if the caller doesn't care for semantics and just wants dump them all, they should use a loop or the wrapper in `go-tpm-tools`.

ElMostafaIdrassi:
Apologies, now I understand. I didn't know about the `ReadPCRs` function in `go-tpm-tools` or the `allPCRs12` function in `go-attestation`.
I like the idea of updating the function's doc, maybe also adding an example or a test to show how to use the function in a loop.

twitchy-jsonp:
All good, thank you for raising this!

Feel free to revise this PR or send another for the function doc update / test, or I'm happy to do it as well :)

ElMostafaIdrassi:
Thank you for your insight! I think it'll be best to close this one and reopen a new fresh PR for the test and doc update.

tomoveu:
@twitchy-jsonp good point. go-tpm is just a mild layer on top of the TPM interface.

In wolfTPM we use wrappers. Can we also have go-tpm wrapper layer for reading all of the PCRs?

Link to the wolftpm/tpm2_wrap.h - https://github.com/wolfSSL/wolfTPM/blob/master/wolftpm/tpm2_wrap.h

PR #204, opened 2020-08-02
TPM2: Added check when parsing TPMLPCRSelection in case of selection of a PCR's index which exceeds max value permitted

ElMostafaIdrassi:
The current implementation makes use of `sizeOfPCRSelect` set to `3`. This means that the maximum size of the `pcrSelect` byte array in `tpmsPCRSelection` is `3`, which means it can only deal with the selection of PCRs from 0 to 23. Attempting to select PCR 24 for example (which might exist on some TPMs) leads to a `panic: runtime error: index out of range [3] with length 3` since `ts.PCRs[byteNum] |= bytePos` tries to access index 3 of the byte array, which is out of range.

This can be fixed by adding a check beforehand, and skipping the PCR if it's index exceeds the maximum value permitted, independently of the value of `sizeOfPCRSelect`.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -73,6 +73,9 @@ func encodeTPMLPCRSelection(sel ...PCRSelection) ([]byte, error) {
 
 		// s[i].PCRs parameter is indexes of PCRs, convert that to set bits.
 		for _, n := range s.PCRs {
+			if n >= 8*sizeOfPCRSelect {
+				continue 
+			}
 			byteNum := n / 8
 			bytePos := byte(1 << byte(n%8))
 			ts.PCRs[byteNum] |= bytePos
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -74,7 +74,7 @@ func encodeTPMLPCRSelection(sel ...PCRSelection) ([]byte, error) {
 		// s[i].PCRs parameter is indexes of PCRs, convert that to set bits.
 		for _, n := range s.PCRs {
 			if n >= 8*sizeOfPCRSelect {
-				continue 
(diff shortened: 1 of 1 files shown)


twitchy-jsonp:
Have you seen a TPM with more than 24 PCRs in the wild?

We should definitely revise PCR selection generation code if the PCR 0-23 assumption is no longer correct. I think the spec permits more than 24 PCRs, but I've never heard of any device doing something like that (and no use-case for additional PCRs).

ElMostafaIdrassi:
@twitchy-jsonp Nope, I've never seen a TPM with more than 24 PCRs. I think the spec specifies the minimum number of PCRs to be 24. However, I find it to be safer to add the check and skip all the PCRs that currently cannot be selected and avoid a panic error.

twitchy-jsonp:
Thats a good point. What do you think about returning an error from `encodeTPMLPCRSelection()` if there are more than 24 PCRs?

ElMostafaIdrassi:
I believe the problem stems more from feeding `encodeTPMLPCRSelection` PCRs that are out of the currently supported range (that is PCRs that are not in the range 0 to 23), rather than from giving it a list of PCRs which exceeds the length of 24. 

For example, the function will do just fine if you give it the following 25 PCRs `{0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,0}` but will fail with a panic if you give it `{24}`.

This is why I believe the check should be done on an index-basis rather than a length one.

Concerning how to deal with it, I'm totally fine with returning an error instead of just skipping.

langbeck:
I agree that the check should be done per index and an error should be returned instead of just skipping.
What do you think @twitchy-jsonp?

twitchy-jsonp:
Yepp that makes sense!

langbeck:
@ElMostafaIdrassi, can you implement these changes?

langbeck (approved):

PR #206, opened 2020-08-10
Change TPMLDigest field "Digests" from slice of tpmutil.U16Bytes to n…

ChriMarMe:
…on slice version.

Signed-off-by: Christopher Meis <christopher.meis@9elements.com>

PR #207, opened 2020-08-11
change TPMLDigest field "Digests" from slice of tpmutil.U16Bytes to s…

ChriMarMe:
…lice of new type BDigest


ChriMarMe:
This is a much cleaner solution compared to #206 
Also this should prevent any hickups while serializing to bytes because it only utilizes native types.

ChriMarMe:
This also does not work.

PR #208, opened 2020-08-11
TPM2: Fix PolicyOR 

ChriMarMe:
Add Encode/Decode function for TPMLDigest struct to generate valid byte slice and parse byte slice into structure
Add test for PolicyOR
Add test for Encode/Decode

### DIFF
--- a/tpm2/encoding_test.go
+++ b/tpm2/encoding_test.go
@@ -24,6 +24,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-tpm/tpmutil"
 )
 
@@ -476,4 +477,45 @@ func TestSignEncode(t *testing.T) {
 			t.Fatalf("encoded Sign message did not match golden value. got: %s, want: %s", hex.EncodeToString(cmdBytes), hex.EncodeToString(testCmdBytes))
 		}
 	})
+}
+
+func TestTPMLDigestEncode(t *testing.T) {
+	hashA := []byte{0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255}
+	hashB := []byte{0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
+
+	hl := TPMLDigest{Digests: []tpmutil.U16Bytes{
+		hashA, hashB},
+	}
+
+	want := []byte{0, 0, 0, 2, 0, 20, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255,
+		0, 20, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
+
+	got, err := hl.Encode()
+	if err != nil {
+		t.Errorf("hl.Encode() failed: %v", err)
+	}
+	if !bytes.Equal(want, got) {
+		t.Errorf("TPMLDigest{%v}.Encode() = %v, want: %v", hl, got, want)
+	}
+}
+
+func TestTPMLDigestDecode(t *testing.T) {
+	hashA := []byte{0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255}
+	hashB := []byte{0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
+
+	want := &TPMLDigest{Digests: []tpmutil.U16Bytes{
+		hashA, hashB},
+	}
+
+	b := []byte{0, 0, 0, 2, 0, 20, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255,
+		0, 20, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
+
+	got, err := DecodeTPMLDigest(b)
+	if err != nil {
+		t.Errorf("DecodeTPMDigest(b) failed: %v", err)
+	}
+
+	if !cmp.Equal(want, got) {
+		t.Errorf("Digests are not the same.\n Have:\n %v\n - Want:\n %v", got.Digests, want.Digests)
+	}
 }
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -19,6 +19,7 @@ import (
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/rsa"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"math/big"
@@ -1023,8 +1024,45 @@ type AuthCommand struct {
 
 // TPMLDigest represents the TPML_Digest structure
 // It is used to convey a list of digest values.
-//This type is used in TPM2_PolicyOR() and in TPM2_PCR_Read()
+// This type is used in TPM2_PolicyOR() and in TPM2_PCR_Read()
 type TPMLDigest struct {
-	Count   uint32
 	Digests []tpmutil.U16Bytes
+}
+
+// Encode converts the TPMLDigest structure into a byte slice
+func (list *TPMLDigest) Encode() ([]byte, error) {
+	res, err := tpmutil.Pack(uint32(len(list.Digests)))
+	if err != nil {
+		return nil, err
+	}
+	for _, item := range list.Digests {
+		b, err := tpmutil.Pack(item)
+		if err != nil {
+			return nil, err
+		}
+		res = append(res, b...)
+
+	}
+	return res, nil
+}
+
+// DecodeTPMLDigest decodes a TPML_Digest part of a message.
+func DecodeTPMLDigest(buf []byte) (*TPMLDigest, error) {
+	in := bytes.NewBuffer(buf)
+	var tpmld TPMLDigest
+	var count uint32
+	if err := binary.Read(in, binary.BigEndian, &count); err != nil {
+		return nil, fmt.Errorf("decoding TPML_Digest: %v", err)
+	}
+	for in.Len() > 0 {
+		var hash tpmutil.U16Bytes
+		if err := hash.TPMUnmarshal(in); err != nil {
+			return nil, err
+		}
+		tpmld.Digests = append(tpmld.Digests, hash)
+	}
+	if count != uint32(len(tpmld.Digests)) {
+		return nil, fmt.Errorf("expected size and read size does not match")
+	}
+	return &tpmld, nil
 }
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -2080,3 +2080,34 @@ func TestDictionaryAttackLockReset(t *testing.T) {
 		t.Fatalf("got %d, expected 0", caps[0].(TaggedProperty).Value)
 	}
 }
+
+func TestPolicyOr(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+	hashAlg, err := AlgSHA256.Hash()
+	if err != nil {
+		t.Errorf("looking up hash algo with error: %v", err)
+	}
+	zeroHash := make([]byte, hashAlg.Size())
+	customHash := make([]byte, hashAlg.Size())
+
+	_, err = rand.Read(customHash)
+	if err != nil {
+		t.Errorf("random read into byte slice with error: %v", err)
+	}
+
+	tpml := TPMLDigest{Digests: []tpmutil.U16Bytes{
+		zeroHash, customHash},
+	}
+
+	sessHandle, _, err := StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionPolicy, AlgNull, AlgSHA256)
+	defer FlushContext(rw, sessHandle)
+
+	if err != nil {
+		t.Errorf("StartAuthSession failed: %v", err)
+	}
+	err = PolicyOr(rw, sessHandle, tpml)
+	if err != nil {
+		t.Errorf("PolicyOr with error: %v", err)
+	}
+}
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/google/go-tpm
 go 1.12
 
 require (
+	github.com/google/go-cmp v0.5.2
 	github.com/google/go-tpm-tools v0.0.0-20190906225433-1614c142f845
 	golang.org/x/sys v0.0.0-20181205085412-a5c9d58dba9a
 )
--- a/go.sum
+++ b/go.sum
@@ -8,6 +8,8 @@ github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSs
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
 github.com/golang/protobuf v1.3.2 h1:6nsPYzhq5kReh6QImI3k5qWzO4PEbvbIW2cwSfR/6xs=
 github.com/golang/protobuf v1.3.2/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
+github.com/google/go-cmp v0.5.2 h1:X2ev0eStA3AbceY54o37/0PQ/UWqKEiiO2dKL5OPaFM=
+github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-tpm v0.1.2-0.20190725015402-ae6dd98980d4/go.mod h1:H9HbmUG2YgV/PHITkO7p6wxEEj/v5nlsVWIwumwH2NI=
 github.com/google/go-tpm-tools v0.0.0-20190906225433-1614c142f845 h1:2WNNKKRI+a5OZi5xiJVfDoOiUyfK/BU1D4w+N6967F4=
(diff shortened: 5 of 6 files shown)


tpm2/structures.go
```diff
@@ -1021,10 +1021,37 @@ type AuthCommand struct {
 	Auth       tpmutil.U16Bytes
 }
 
+// BDigest implements the structure of TPM2B_Digest as described in Part 2 Structures, Page 97
+type BDigest struct {
+	Size   uint16
```
chrisfenner, line 1026: What does this type add that is missing from U16Bytes? For example, Size here is len(Digest).

tpm2/structures.go
```diff
@@ -1021,10 +1021,37 @@ type AuthCommand struct {
...
 // It is used to convey a list of digest values.
 //This type is used in TPM2_PolicyOR() and in TPM2_PCR_Read()
 type TPMLDigest struct {
 	Count   uint32
-	Digests []tpmutil.U16Bytes
+	Digests []BDigest
+}
+
+// ToBytes converts the TPMLDigest structure into a byte slice
+func (list *TPMLDigest) ToBytes() ([]byte, error) {
```
chrisfenner, line 1039: This looks like it is meant to be the Encode() method for serializing a TPMLDigest to the wire format. I think (but am not 100% sure) that if you implement a custom Encode() for a struct, tpmutil.Pack will "just work" and you won't have to touch PolicyOR as part of this PR.

tpm2/structures.go
```diff
@@ -1021,10 +1021,37 @@ type AuthCommand struct {
...
+	if uint32(len(list.Digests)) != list.Count {
```
chrisfenner, line 1040: If you implement a custom Encode() function for this struct, why not eliminate the Count field altogether? That way, there isn't an inconsistency that could exist.

tpm2/structures.go
chrisfenner: Note that if you implement a special Encode(), you probably also need to implement DecodeTPMLDigest.

tpm2/test/tpm2_test.go
```diff
@@ -2080,3 +2080,29 @@ func TestDictionaryAttackLockReset(t *testing.T) {
 		t.Fatalf("got %d, expected 0", caps[0].(TaggedProperty).Value)
 	}
 }
+
+func TestTPMLDigest(t *testing.T) {
```
chrisfenner, line 2084: This test case seems like it may belong in encoding_test.go instead. This test doesn't call out to the TPM.

tpm2/test/tpm2_test.go
```diff
@@ -2080,3 +2080,29 @@ func TestDictionaryAttackLockReset(t *testing.T) {
...
+	hashA := make([]byte, 20)
```
chrisfenner, line 2085: I think this test might be more thorough if it put some sample data into hashA and hashB, such as
0102030405... and fffefdfcfe..., and then asserted the expected full serialization with the correct order of the correct sample hashes.

ChriMarMe:
#209

tpm2/structures.go
ChriMarMe: Yep. I agree, but spec disagrees.

tpm2/test/tpm2_test.go
ChriMarMe: agreed

tpm2/structures.go
ChriMarMe: Yes, but the field exists because the spec says so. My point is: you read the spec, you see the code and there is no magic happening in between....for example the whole concept of U16Bytes and U32Bytes is a nice solution programming wise, but it is a huge pain in the ass to handle and it obfuscates parts of the specification.

tpm2/test/tpm2_test.go
ChriMarMe: agreed

tpm2/structures.go
chrisfenner: The spec wire format is `size uint16 || data []byte`, but that doesn't mean that the in-memory representation in go-tpm should have a separate size field [^1] - part of the job of go-tpm is to translate to and from a higher level representation (like []byte) to a lower level representation (like a correctly serialized TPM2B_DIGEST).

Indeed the entire purpose of tpmutil.U16Bytes is to allow Go code to treat as a boring byte slice something that needs to be serialized into the TPM2B format. So I maintain that the proper fix to the serialization errors you're seeing here is not to create a new type for TPM2B_DIGEST [^2], but to figure out why the full TPM2_PolicyOR is not serializing as expected [^3].

[^1]: If we have to track a separate size field that is set independently of the data, now we need consistency checks in **many** places
[^2]: I'd actually be totally shocked if this fix was the first time go-tpm has needed to change up the way it represents a TPM2B_DIGEST. Those are everywhere.
[^3]: I think TPMLDigest just needs an override `Encode`/`Decode...` pair because the automatic one just isn't doing the trick.

tpm2/structures.go
chrisfenner: I'd like to discuss further: what specifically is painful about U16Bytes/U32Bytes and family? A TSS's job is to make it easier to write TPM commands, otherwise we'd all be assembling raw TPM commands byte by byte. If you walk me through why every user should have to copy len(someSlice) into a field to properly call a command, maybe we can find the source of the general problem with the API that we need to solve.

Re: your general point that the TSS and the TPM API should be well (and completely) mapped to each other - well heard, and I think go-tpm has some room to grow in that aspect. But, IMHO, U16Bytes is not the root cause of the dissonance here.

tpm2/encoding_test.go
```diff
@@ -477,3 +477,36 @@ func TestSignEncode(t *testing.T) {
...
+	} //sum 48
+
+	exp := []byte{0, 0, 0, 2, 0, 20, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255,
+		0, 20, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
+
+	b, err := hl.Encode()
+	if err != nil {
+		t.Errorf("Conversion to []byte failed")
+	}
+	if len(b) != int(48) {
```
chrisfenner, line 506: Should we use len(exp) instead of hardcoded 48 here?

tpm2/structures.go
ChriMarMe: The tpmutil.Pack() function can't handle slices of U16Bytes, and my inability to change that, brought me to that statement mostly. Was not my best statement, nor most elaborate. I see the point.

tpm2/encoding_test.go
ChriMarMe: agreed

ChriMarMe:
No API constraints, only rookie coder who tries to play with the big guys. I code Go for a year now, and sometimes I'm just missing the experience and/or patience to solve a problem and I'm not very familiar with tpm2 spec. :P 

Anyway, thanks for your input and review.

tpm2/structures.go
ChriMarMe: I found a way to handle that. Thank you :)

tpm2/encoding_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	"testing"
 
 	"github.com/google/go-tpm/tpmutil"
+	"github.com/stretchr/testify/assert"
```
chrisfenner, line 28: I'd like @josephlr or @twitchy-jsonp input on introducing this (and other) dependencies here

tpm2/structures.go
```diff
@@ -1028,3 +1029,70 @@ type TPMLDigest struct {
...
+func DecodeTPMLDigest(buf []byte) (*TPMLDigest, error) {
+	in := bytes.NewBuffer(buf)
+	var tpmld TPMLDigest
+	var err error
+	var hashlen uint16
+	if err = binary.Read(in, binary.BigEndian, &tpmld.Count); err != nil {
+		return &tpmld, fmt.Errorf("decoding TPML_Digest: %v", err)
+	}
+
+	for i := uint32(0); i < tpmld.Count; i++ {
```
chrisfenner, line 1065: I think this code could be simplified to take advantage of U16Bytes.TPMUnmarshal, which handles the mechanics of reading the size and byte array:
https://github.com/google/go-tpm/blob/240a3433a7a53049ab3386b7dace1e4d90dba2ce/tpmutil/structures.go#L56-L78

Here, I'd propose only checking the `Count` that you read at the beginning against the total number of `U16Bytes`es that you deserialized from the slice, something like:

```
for in.Len() > 0 {
  var d tpmutil.U16Bytes
  if err = d.TPMUnmarshal(in); err != nil {
    return nil, err
  }
  tpmld.Digests = append(tpmld.Digsts, d)
}
if Count != len(tpmld.Digests) {
  return nil, fmt.Errorf("blah blah blah")
}

File: tpm2/structures.go [hunk 9]
```diff
@@ -1028,3 +1029,70 @@ type TPMLDigest struct {
...
+
+// DecodeTPMLDigest decodes a TPML_Digest part of a message. No error is returned if
+// the input has extra trailing data.
+func DecodeTPMLDigest(buf []byte) (*TPMLDigest, error) {
+	in := bytes.NewBuffer(buf)
+	var tpmld TPMLDigest
+	var err error
+	var hashlen uint16
+	if err = binary.Read(in, binary.BigEndian, &tpmld.Count); err != nil {
+		return &tpmld, fmt.Errorf("decoding TPML_Digest: %v", err)
```
chrisfenner, line 1062: Should this function return nil instead of a partly initialized TPMLDigest if there is an error? (And again in a few other places in the current commit)

tpm2/encoding_test.go
ChriMarMe: :+1:

tpm2/structures.go
```diff
@@ -1025,6 +1026,57 @@ type AuthCommand struct {
...
+		}
+		tpmld.Digests = append(tpmld.Digests, hash)
+	}
+	if count != uint32(len(tpmld.Digests)) {
+		return nil, fmt.Errorf("expected size and read size does not match")
+	}
+	return &tpmld, nil
+}
+
+// NewTPMLDigest creates a new TPMLDigests structure by providing a list of tpmutil.U16Bytes
```
chrisfenner, line 1071: I actually think we don't need this. Since we got rid of Count (and the need for the consistency check), a TPMLDigest can be cleanly constructed with struct literals, like
```
l := TPMLDigest{
  Digests: []tpmutil.U16Bytes{
    digest1,
    digest2,
    ...,
  }
}
```

tpm2/structures.go
```diff
@@ -1025,6 +1026,57 @@ type AuthCommand struct {
 // It is used to convey a list of digest values.
 //This type is used in TPM2_PolicyOR() and in TPM2_PCR_Read()
```
chrisfenner, line 1027: nit: space between // and This

tpm2/structures.go
```diff
@@ -1025,6 +1026,57 @@ type AuthCommand struct {
...
+	}
+	if count != uint32(len(tpmld.Digests)) {
+		return nil, fmt.Errorf("expected size and read size does not match")
+	}
+	return &tpmld, nil
+}
+
+// NewTPMLDigest creates a new TPMLDigests structure by providing a list of tpmutil.U16Bytes
+func NewTPMLDigest(hashes ...[]byte) (*TPMLDigest, error) {
+	if len(hashes) < 2 {
```
chrisfenner, line 1073: If you decide to keep this function: technically a TPML_Digest may even contain 0 or 1 digest (and in fact the 0 case is needed sometimes)

tpm2/test/tpm2_test.go
```diff
@@ -2080,3 +2080,35 @@ func TestDictionaryAttackLockReset(t *testing.T) {
...
+	hashAlg, err := AlgSHA256.Hash()
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+	zeroHash := make([]byte, hashAlg.Size())
+	customHash := make([]byte, hashAlg.Size())
+
+	_, err = rand.Read(customHash)
+	if err != nil {
+		t.Errorf("%v", err)
```
chrisfenner, line 2096: Could you make these test errors more descriptive? e.g., "Error reading random data: %v" and so on

ChriMarMe:
I have to thank you for your patience, guidance and quick responses @chrisfenner :+1:

chrisfenner:
It's my pleasure to help how I can! You're the one doing the actual work to implement a new command in go-tpm, so I have to thank you back for that 😄

ChriMarMe:
@twitchy-jsonp @josephlr Would be nice to get the last bits correct :) can you check it please?

tpm2/encoding_test.go
twitchy-jsonp: Assert in Go is discouraged over simple if statements. Perhaps a bunch of the bits below can be refactored into:

```go

if got, want := thing.Output(), "wanted"; got != want {
  t.Errorf("thing.Output() = %v, want %v", got, want)
}
```

ChriMarMe: Alright. Thank you :)

tpm2/encoding_test.go
```diff
@@ -477,3 +477,45 @@ func TestSignEncode(t *testing.T) {
...
+
+	want := []byte{0, 0, 0, 2, 0, 20, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255,
+		0, 20, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
+
+	have, err := hl.Encode()
+	if err != nil {
+		t.Errorf("Conversion to []byte failed")
+	}
+	if !bytes.Equal(want, have) {
+		t.Errorf("Have: %v \n Want: %v", have, want)
```
twitchy-jsonp, line 497: Nit: The term `got` is more frequently used than `have`. Its also good to describe what specifically the value represents in the error string, so you can tell whats going on just by the test output without looking at the test code. IE:

```go
hl := TPMLDigest{Digests: []tpmutil.U16Bytes{
	hashA, hashB},
}

want := []byte{0, 0, 0, 2, 0, 20, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255,
	0, 20, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
got, err := hl.Encode()
if err != nil {
	t.Errorf("hl.Encode() failed: %v", err)
}
if !bytes.Equal(got, want) {
	t.Errorf("TPMLDigest{%v}.Encode() = %v, want %v", h1, got, want)
}
```

tpm2/structures.go
```diff
@@ -1023,8 +1025,57 @@ type AuthCommand struct {
...
-//This type is used in TPM2_PolicyOR() and in TPM2_PCR_Read()
+// This type is used in TPM2_PolicyOR() and in TPM2_PCR_Read()
 type TPMLDigest struct {
-	Count   uint32
 	Digests []tpmutil.U16Bytes
 }
+
+func (list *TPMLDigest) String() string {
+	var s strings.Builder
+	s.WriteString("TPMLDigest{")
```
twitchy-jsonp, line 1035: TPMLDigest and each Digest has an opening brace `{` - do we want to emit a closing brace `}` at the end as well to balance them out?

ChriMarMe: fixed it

ChriMarMe:
What do I miss? I can't tell :(

tpm2/encoding_test.go
```diff
@@ -477,3 +477,45 @@ func TestSignEncode(t *testing.T) {
...
+
+	b := []byte{0, 0, 0, 2, 0, 20, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255,
+		0, 20, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
+
+	got, err := DecodeTPMLDigest(b)
+	if err != nil {
+		t.Errorf("DecodeTPMDigest(b) failed: %v", err)
+	}
+
+	if got.String() != want.String() {
```
chrisfenner, line 517: nit: if there were a bug in String() (for example, always returning the empty string) it would hide a bug in Decode. Can we instead test for equality between want and got's `Digests` fields?

tpm2/structures.go
```diff
@@ -1023,8 +1025,58 @@ type AuthCommand struct {
...
+	for iterator, item := range list.Digests {
+		s.WriteString("Digest")
+		s.WriteString(string(iterator) + "{")
+		s.WriteString(string(item) + "\n")
```
chrisfenner, line 1039: `item` is likely to contain non-printable characters - did you mean to encode each digest to something printable (like hexadecimal) here?

tpm2/structures.go
chrisfenner: I see an outer opening brace on line 1035, and then a pair of braces for each item (1038, 1040), and no outer closing brace - am I reading it correctly?

tpm2/structures.go
ChriMarMe: I got rid of the String function

tpm2/encoding_test.go
```diff
@@ -477,3 +477,49 @@ func TestSignEncode(t *testing.T) {
...
+
+	b := []byte{0, 0, 0, 2, 0, 20, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 255,
+		0, 20, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 255}
+
+	got, err := DecodeTPMLDigest(b)
+	if err != nil {
+		t.Errorf("DecodeTPMDigest(b) failed: %v", err)
+	}
+
+	if len(got.Digests) != len(want.Digests) {
```
chrisfenner, line 517: nit: the comparison between want and got will panic below if `len(got.Digests) != len(want.Digests)` - if we keep this design for this test, then the `for iterator, item...` test ought to be in an `else` clause below this `if`.

I think I made this comment before, but it got lost because I didn't attach it to a review. Is it possible to use `cmp.Equal(got, want)` to simplify the check and keep it panic-free?

chrisfenner (changes requested):

tpm2/encoding_test.go
ChriMarMe: done

chrisfenner (approved):
Thanks for the change! Tests looking very nice 😊

PR #211, opened 2020-08-15
TPM2: Added CertifyEx and encodeCertifyEx

### LINKED ISSUE #210: TPM2: Certify makes use of hardcoded signature scheme
Hello everyone,

The implementation of `Certify` does not give the caller the ability to set the signature scheme but instead, it calls `encodeCertify` which sets it to the hardcoded value `AlgRSASSA, AlgSHA256`.

This works if the signing key has a signature scheme set to exactly `AlgRSASSA, AlgSHA256`, but will fail if it is anything else (i.e. 
`AlgRSASSA, AlgSHA1`, `AlgECDSA, AlgSHA256`...)

Here is a quick test that showcases this failure (it's the same as `TestCertify`, I just changed `AlgSHA256` to `AlgSHA1`)

```
func TestCertifySHA1(t *testing.T) {
	rw := openTPM(t)
	defer rw.Close()

	params := Public{
		Type:       AlgRSA,
		NameAlg:    AlgSHA256,
		Attributes: FlagSignerDefault,
		RSAParameters: &RSAParams{
			Sign: &SigScheme{
				Alg:  AlgRSASSA,
				Hash: AlgSHA1,
			},
			KeyBits: 2048,
		},
	}
	signerHandle, signerPub, err := CreatePrimary(rw, HandleOwner, pcrSelection7, emptyPassword, defaultPassword, params)
	if err != nil {
		t.Fatalf("CreatePrimary(signer) failed: %s", err)
	}
	defer FlushContext(rw, signerHandle)

	subjectHandle, subjectPub, err := CreatePrimary(rw, HandlePlatform, pcrSelection7, emptyPassword, defaultPassword, params)
	if err != nil {
		t.Fatalf("CreatePrimary(subject) failed: %s", err)
	}
	defer FlushContext(rw, subjectHandle)

	attest, sig, err := Certify(rw, defaultPassword, defaultPassword, subjectHandle, signerHandle, nil)
	if err != nil {
		t.Errorf("Certify failed: %s", err)
		return
	}

	attestHash := sha256.Sum256(attest)
	if err := rsa.VerifyPKCS1v15(signerPub.(*rsa.PublicKey), crypto.SHA256, attestHash[:], sig); err != nil {
		t.Errorf("Signature verification failed: %v", err)
	}

	t.Run("DecodeAttestationData", func(t *testing.T) {
		ad, err := DecodeAttestationData(attest)
		if err != nil {
			t.Fatal("DecodeAttestationData:", err)
		}
		params := Public{
			Type:       AlgRSA,
			NameAlg:    AlgSHA256,
			Attributes: FlagSignerDefault,
			RSAParameters: &RSAParams{
				Sign: &SigScheme{
					Alg:  AlgRSASSA,
					Hash: AlgSHA1,
				},
				KeyBits: 2048,
				// Note: we don't include Exponent because CreatePrimary also
				// returns Public without it.
				ModulusRaw: subjectPub.(*rsa.PublicKey).N.Bytes(),
			},
		}
		matches, err := ad.AttestedCertifyInfo.Name.MatchesPublic(params)
		if err != nil {
			t.Fatalf("AttestedCertifyInfo.Name.MatchesPublic error: %v", err)
		}
		if !matches {
			t.Error("Name in AttestationData doesn't match Public structure of subject")
		}
	})
}
```

The test fails with the error `parameter 2, error code 0x12 : unsupported or incompatible scheme`

Can we modify `Certify` so that it takes the sigScheme as an argument and gives the caller the ability to set it accordingly rather than harcode it to `AlgRSASSA, AlgSHA256` in `encodeCertify` ? I know this would break the API, but I can't think of anything better. I tried setting the hash alg of the sig scheme to `AlgNull` in vain, I kept getting `parameter 2, error code 0x3 : hash algorithm not supported or not appropriate`.

Cheers,

ElMostafaIdrassi:
These functions differ from `Certify` and `encodeCertify` in that they take the `scheme` to be used as an additional argument.
Should close #210.

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -581,49 +581,49 @@ func TestCertifyEx(t *testing.T) {
 
 	restrictedKeySignerFlags := FlagSignerDefault
 	unrestrictedKeySignerFlags := FlagSign | FlagFixedTPM | FlagFixedParent | FlagSensitiveDataOrigin | FlagUserWithAuth
-	testFlags := []struct {
+	testCases := []struct {
 		description  string
 		attributes   KeyProp
-		keyScheme    *SigScheme
+		keyScheme    SigScheme
 		passedScheme SigScheme
 		shouldPass   bool
 	}{
-		{"Null-SHA1", unrestrictedKeySignerFlags, &SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA1}, true},
-		{"Null-SHA256", unrestrictedKeySignerFlags, &SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, true},
-		{"Null-Null", unrestrictedKeySignerFlags, &SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgNull}, false},
-		{"SHA256-Null", restrictedKeySignerFlags, &SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgNull}, true},
-		{"SHA256-SHA256", restrictedKeySignerFlags, &SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, true},
-		{"SHA256-SHA1", restrictedKeySignerFlags, &SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA1}, false},
-	}
-	params := Public{
-		Type:    AlgRSA,
-		NameAlg: AlgSHA256,
-		RSAParameters: &RSAParams{
-			KeyBits: 2048,
-		},
+		{"Null-SHA1", unrestrictedKeySignerFlags, SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA1}, true},
+		{"Null-SHA256", unrestrictedKeySignerFlags, SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, true},
+		{"Null-Null", unrestrictedKeySignerFlags, SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgNull}, false},
+		{"SHA256-Null", restrictedKeySignerFlags, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgNull}, true},
+		{"SHA256-SHA256", restrictedKeySignerFlags, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, true},
+		{"SHA256-SHA1", restrictedKeySignerFlags, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA1}, false},
 	}
 
-	for _, flag := range testFlags {
-		params.Attributes = flag.attributes
-		params.RSAParameters.Sign = flag.keyScheme
+	for _, testCase := range testCases {
+		params := Public{
+			Type:       AlgRSA,
+			NameAlg:    AlgSHA256,
+			Attributes: testCase.attributes,
+			RSAParameters: &RSAParams{
+				Sign:    &testCase.keyScheme,
+				KeyBits: 2048,
+			},
+		}
 
-		t.Run(flag.description, func(t *testing.T) {
-			signerHandle, _, err := CreatePrimary(rw, HandleOwner, pcrSelection7, emptyPassword, defaultPassword, params)
+		t.Run(testCase.description, func(t *testing.T) {
+			signerHandle, _, err := CreatePrimary(rw, HandleOwner, PCRSelection{}, emptyPassword, defaultPassword, params)
 			if err != nil {
 				t.Fatalf("CreatePrimary(signer) failed: %s", err)
 			}
 			defer FlushContext(rw, signerHandle)
 
-			subjectHandle, _, err := CreatePrimary(rw, HandlePlatform, pcrSelection7, emptyPassword, defaultPassword, params)
+			subjectHandle, _, err := CreatePrimary(rw, HandlePlatform, PCRSelection{}, emptyPassword, defaultPassword, params)
 			if err != nil {
 				t.Fatalf("CreatePrimary(subject) failed: %s", err)
 			}
 			defer FlushContext(rw, subjectHandle)
 
-			_, _, err = CertifyEx(rw, defaultPassword, defaultPassword, subjectHandle, signerHandle, nil, flag.passedScheme)
-			if err != nil && flag.shouldPass {
+			_, _, err = CertifyEx(rw, defaultPassword, defaultPassword, subjectHandle, signerHandle, nil, testCase.passedScheme)
+			if err != nil && testCase.shouldPass {
 				t.Errorf("CertifyEx expected to succeed but failed: %s", err)
-			} else if err == nil && !flag.shouldPass {
+			} else if err == nil && !testCase.shouldPass {
 				t.Errorf("CertifyEx expected to fail but succeeded")
 			}
 		})
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -579,138 +579,55 @@ func TestCertifyEx(t *testing.T) {
 	rw := openTPM(t)
 	defer rw.Close()
 
+	restrictedKeySignerFlags := FlagSignerDefault
+	unrestrictedKeySignerFlags := FlagSign | FlagFixedTPM | FlagFixedParent | FlagSensitiveDataOrigin | FlagUserWithAuth
+	testFlags := []struct {
+		description  string
+		attributes   KeyProp
+		keyScheme    *SigScheme
+		passedScheme SigScheme
+		shouldPass   bool
+	}{
+		{"Null-SHA1", unrestrictedKeySignerFlags, &SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA1}, true},
+		{"Null-SHA256", unrestrictedKeySignerFlags, &SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, true},
+		{"Null-Null", unrestrictedKeySignerFlags, &SigScheme{Alg: AlgNull}, SigScheme{Alg: AlgNull}, false},
+		{"SHA256-Null", restrictedKeySignerFlags, &SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgNull}, true},
+		{"SHA256-SHA256", restrictedKeySignerFlags, &SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, true},
+		{"SHA256-SHA1", restrictedKeySignerFlags, &SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256}, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA1}, false},
+	}
 	params := Public{
-		Type:       AlgRSA,
-		NameAlg:    AlgSHA256,
-		Attributes: FlagSignerDefault,
+		Type:    AlgRSA,
+		NameAlg: AlgSHA256,
 		RSAParameters: &RSAParams{
-			Sign: &SigScheme{
-				Alg:  AlgRSASSA,
-				Hash: AlgSHA256,
-			},
 			KeyBits: 2048,
 		},
 	}
-	signerHandle, signerPub, err := CreatePrimary(rw, HandleOwner, pcrSelection7, emptyPassword, defaultPassword, params)
-	if err != nil {
-		t.Fatalf("CreatePrimary(signer) failed: %s", err)
+
+	for _, flag := range testFlags {
+		params.Attributes = flag.attributes
+		params.RSAParameters.Sign = flag.keyScheme
+
+		t.Run(flag.description, func(t *testing.T) {
+			signerHandle, _, err := CreatePrimary(rw, HandleOwner, pcrSelection7, emptyPassword, defaultPassword, params)
+			if err != nil {
+				t.Fatalf("CreatePrimary(signer) failed: %s", err)
+			}
+			defer FlushContext(rw, signerHandle)
+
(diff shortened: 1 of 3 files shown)


tpm2/tpm2.go
```diff
@@ -1546,6 +1546,26 @@ func encodeCertify(parentAuth, ownerAuth string, object, signer tpmutil.Handle,
...
+	if err != nil {
+		return nil, err
+	}
+
+	auth, err := encodeAuthArea(AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession, Auth: []byte(parentAuth)}, AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession, Auth: []byte(ownerAuth)})
+	if err != nil {
+		return nil, err
+	}
+
+	params, err := tpmutil.Pack(qualifyingData, scheme)
```
chrisfenner, line 1562: Can you update this to correctly pack TPM_ALG_NULL (with no following hash alg) for the NULL case of TPMT_SCHEME? Feel free to file a GitHub issue for "tpmtSigScheme does not match TPMT_SIG_SCHEME" and reference it with a TODO comment explaining that this special-case command packing would be a workaround and not needed after the bug is fixed.

tpm2/tpm2.go
```diff
@@ -1546,6 +1546,26 @@ func encodeCertify(parentAuth, ownerAuth string, object, signer tpmutil.Handle,
 	return concat(ha, auth, params)
 }
 
+// This function differs from encodeCertify in that it takes the scheme
+// to be used as an additional argument.
+func encodeCertifyEx(parentAuth, ownerAuth string, object, signer tpmutil.Handle, qualifyingData tpmutil.U16Bytes, scheme tpmtSigScheme) ([]byte, error) {
```
chrisfenner, line 1551: As discussed in #210, can you rename these to `objectAuth` and `signerAuth` as you suggested, both here and in Certify?

tpm2/tpm2.go
```diff
@@ -1594,6 +1614,24 @@ func Certify(rw io.ReadWriter, parentAuth, ownerAuth string, object, signer tpmu
 	return decodeCertify(resp)
 }
 
+// CertifyEx generates a signature of a loaded TPM object with a signing key
+// signer. This function differs from Certify in that it takes the scheme
```
chrisfenner, line 1618: Can you also add a comment to `Certify` that explains it will hardcode `RSAPSS` and `SHA256` for the signing scheme?

tpm2/tpm2.go
ElMostafaIdrassi: Do you mean when the passed `scheme` is `nil` ? or when the `Scheme` of `tpmtSigScheme` is `AlgNull` ? or both ?

tpm2/tpm2.go
chrisfenner: Oops, I miswrote. Thank you for correctly using `RSASSA` in the actual change 😃

tpm2/tpm2.go
chrisfenner: I think either one would be workable.

My 2 cents: I prefer continuing to take `scheme` by value, and special-casing `if scheme.Scheme == tpmAlgNull` as a workaround (making sure to reference the GitHub issue in a comment, so we can clean it up later when fixed), as personally I might wonder as a user if the API will substitute some default value if I provide `nil` for `scheme`. AlgNull is very clear and helps guide the user to expectations based in the TPM spec.

ElMostafaIdrassi:
I think it will be good to add a test case for `CertifyEx`, but for that, `tpmtSigScheme` needs to be exported. What's your opinion on that, everyone ?

chrisfenner:
I think we should export `tpmtSigScheme` but name it something nicer, like `SigScheme`. Except, oops... `SigScheme` is already taken by a type that expresses `TPMT_RSA_SCHEME`...  oof.

The good news is that the current `SigScheme` actually looks compatible with what we need here, and it already handles the Null case. Do you want to try just eliminating `tpmtSigScheme` and replacing its usage with `SigScheme`?

tpm2/tpm2.go
ElMostafaIdrassi: No worries, I figured it was just a typo ;).

ElMostafaIdrassi:
@chrisfenner I'll give it a try but on a separate PR just to be clean.

tpm2/tpm2.go
ElMostafaIdrassi: Does this look good now ?

ElMostafaIdrassi:
I don't see how I can add `CertifyEx` tests without exporting `tpmtSigScheme` first. You have suggested before that I replace all occurrences of `tpmtSigScheme` with `SigScheme`, and I decided to create a separate PR for that, but now that I think of it, maybe I'll just push the changes here.

tpm2/test/tpm2_test.go
```diff
@@ -575,6 +575,144 @@ func TestCertify(t *testing.T) {
...
+	}
+	defer FlushContext(rw, signerHandle)
+
+	subjectHandle, subjectPub, err := CreatePrimary(rw, HandlePlatform, pcrSelection7, emptyPassword, defaultPassword, params)
+	if err != nil {
+		t.Fatalf("CreatePrimary(subject) failed: %s", err)
+	}
+	defer FlushContext(rw, subjectHandle)
+
+	attest, sig, err := CertifyEx(rw, defaultPassword, defaultPassword, subjectHandle, signerHandle, nil, SigScheme{Alg: AlgRSASSA, Hash: AlgSHA256, Count: 0})
```
chrisfenner, line 606: Nit: The `Count` member of SigScheme is not used in this case and not relevant to this call, and omitting it (letting it default to the same value, 0) seems more readable.

tpm2/test/tpm2_test.go
```diff
@@ -575,6 +575,144 @@ func TestCertify(t *testing.T) {
...
+func TestCertifyEx(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	params := Public{
+		Type:       AlgRSA,
+		NameAlg:    AlgSHA256,
+		Attributes: FlagSignerDefault,
+		RSAParameters: &RSAParams{
+			Sign: &SigScheme{
```
chrisfenner, line 587: This test should set the NULL signing scheme on the created key, and use it to certify with two different algorithms - this proves that the expected data is being sent to the TPM (make a key with no opinion about which signing scheme to use, and use it with two different signing schemes specified by the call)

tpm2/test/tpm2_test.go
```diff
@@ -575,6 +575,144 @@ func TestCertify(t *testing.T) {
...
+		t.Errorf("Certify failed: %s", err)
+		return
+	}
+
+	attestHash := sha256.Sum256(attest)
+	if err := rsa.VerifyPKCS1v15(signerPub.(*rsa.PublicKey), crypto.SHA256, attestHash[:], sig); err != nil {
+		t.Errorf("Signature verification failed: %v", err)
+	}
+
+	t.Run("DecodeAttestationData", func(t *testing.T) {
```
chrisfenner, line 617: This is awesome, but I don't think it's necessary to test that the TPM properly signed the right thing for the Certify. I think it's sufficient to verify that the TPM returned some data that was signed according to the scheme that was passed in (for each scheme that we pass in)

tpm2/test/tpm2_test.go
```diff
@@ -575,6 +575,144 @@ func TestCertify(t *testing.T) {
...
+	}
+	defer FlushContext(rw, signerHandle)
+
+	subjectHandle, subjectPub, err := CreatePrimary(rw, HandlePlatform, pcrSelection7, emptyPassword, defaultPassword, params)
+	if err != nil {
+		t.Fatalf("CreatePrimary(subject) failed: %s", err)
+	}
+	defer FlushContext(rw, subjectHandle)
+
+	attest, sig, err := CertifyEx(rw, defaultPassword, defaultPassword, subjectHandle, signerHandle, nil, SigScheme{Alg: AlgNull, Hash: AlgSHA256, Count: 0})
```
chrisfenner, line 675: AlgNull/AlgSHA256 doesn't really make sense as a combination of values to SigScheme - we should rely on default initialization being good enough for the members that don't matter to the use-case and have simply SigScheme{Alg: AlgNull}

tpm2/test/tpm2_test.go
```diff
@@ -575,6 +575,144 @@ func TestCertify(t *testing.T) {
...
+		t.Errorf("Certify failed: %s", err)
+		return
+	}
+
+	attestHash := sha256.Sum256(attest)
+	if err := rsa.VerifyPKCS1v15(signerPub.(*rsa.PublicKey), crypto.SHA256, attestHash[:], sig); err != nil {
+		t.Errorf("Signature verification failed: %v", err)
+	}
+
+	t.Run("DecodeAttestationData", func(t *testing.T) {
```
chrisfenner, line 686: As above, I don't think it's necessary to have this much code to test that the TPM implementation itself correctly produced the signature.

tpm2/test/tpm2_test.go
ElMostafaIdrassi: I made sure `TestCertifyEx` replicates the same code that is in `TestCertify`, that's why this test is there. Are you sure you want this piece of code gone ?

tpm2/test/tpm2_test.go
ElMostafaIdrassi: Not following here. So you want `params` to have `SigScheme{ Alg: AlgNull }` as `Sign` ? Are you sure this is possible ? Because I just tried and it gives : `CreatePrimary(signer) failed: parameter 2, error code 0x12 : unsupported or incompatible scheme`. Maybe you meant to have the key with a set ``SigScheme` (non NULL), then try to certify with the same `SigScheme` and with the NULL one.

tpm2/test/tpm2_test.go
chrisfenner: That's my thinking, @twitchy-jsonp or @josephlr what do you think?

Philosophically: this library is for making it easy to talk to the TPM. The unit test for this library should ensure that it successfully performs the intended action with the TPM, not that the TPM implementation itself is correct. IMHO, this code from the original Certify test makes it harder to read/maintain (for example, to add another iteration where we sign with another padding scheme)

tpm2/test/tpm2_test.go
chrisfenner: Yes, but there is a hidden problem here 😄 `FlagSignerDefault` specifies a restricted key. So this test is creating a restricted key for use in the Certify command. This is normal, but you can use unrestricted keys and this library should not prevent this. If needed for testing, the test should create an unrestricted key for the test.

The TPM spec does not allow you to create a restricted signing key with a `NULL` signing scheme. You have to pick the one scheme that is allowed for that key. This is the TPM forcing good hygiene (and the error from the TPM you mentioned above).

If you created a non-restricted signing key, you could set `NULL` on the signing scheme. This means: allow the user to specify the signing scheme at use time later.

As I see it, the point of CertifyEx is to allow the caller to pass the signing scheme to the command (either null or some specific scheme). This implies a matrix of good test cases:

1.  Key scheme is NULL, pass some scheme (should pass)
2. - Key scheme is NULL, pass another scheme (should pass)
3. - Key scheme is NULL, pass NULL (should fail)
4. - Key scheme is some scheme, pass NULL (should pass)
5. - Key scheme is some scheme, pass same scheme (should pass)
6. - Key scheme is some scheme, pass another scheme (should fail)

First 3 cases require creating an unrestricted signing key for the test.

chrisfenner: BTW, you may already be familiar with [Table Driven Testing](https://github.com/golang/go/wiki/TableDrivenTests) but if you incorporate this strategy here, you can have a very readable and thorough test that covers these 6 cases 😊

ElMostafaIdrassi: Aha, now I understand, thank you very much! I'm just starting my journey learning these TPM shenanigans, so forgive my ignorance 😄. I'll make sure to give the test another try using the technique you've mentionned. BTW, I think there will be no need for `TestCertifyExAlgNull`, as `TestCertifyEx` will incorporate everything.

chrisfenner: I'm happy to provide any info you need! Welcome to TPM - I promise that these details start to become second nature after you spend enough time around the spec 😆

tpm2/test/tpm2_test.go
```diff
@@ -579,138 +579,55 @@ func TestCertifyEx(t *testing.T) {
...
-			t.Error("Name in AttestationData doesn't match Public structure of subject")
-		}
-	})
-}
-
-func TestCertifyExAlgNull(t *testing.T) {
-	rw := openTPM(t)
-	defer rw.Close()
-
 	params := Public{
```
chrisfenner, line 598: You could initialize this inside the loop and pass along the array member so you don't have to update later

tpm2/test/tpm2_test.go
```diff
@@ -579,138 +579,55 @@ func TestCertifyEx(t *testing.T) {
...
-		t.Errorf("Certify failed: %s", err)
-		return
-	}
-
-	attestHash := sha256.Sum256(attest)
-	if err := rsa.VerifyPKCS1v15(signerPub.(*rsa.PublicKey), crypto.SHA256, attestHash[:], sig); err != nil {
-		t.Errorf("Signature verification failed: %v", err)
+	restrictedKeySignerFlags := FlagSignerDefault
+	unrestrictedKeySignerFlags := FlagSign | FlagFixedTPM | FlagFixedParent | FlagSensitiveDataOrigin | FlagUserWithAuth
+	testFlags := []struct {
```
chrisfenner, line 584: nit: can this be called `cases` or `testCases` since it holds more than just flags?

tpm2/test/tpm2_test.go
```diff
@@ -579,138 +579,55 @@ func TestCertifyEx(t *testing.T) {
...
+		description  string
+		attributes   KeyProp
+		keyScheme    *SigScheme
```
chrisfenner, line 587: nit: It's probably easier to read if `keyScheme` is a value (just type SigScheme) here and you take the address of it with &keyScheme on line 608, where you need to pass a *SigScheme to params.RSAParameters.Sign

tpm2/test/tpm2_test.go
```diff
@@ -579,138 +579,55 @@ func TestCertifyEx(t *testing.T) {
...
+		params.Attributes = flag.attributes
+		params.RSAParameters.Sign = flag.keyScheme
 
-	attest, sig, err := CertifyEx(rw, defaultPassword, defaultPassword, subjectHandle, signerHandle, nil, SigScheme{Alg: AlgNull, Hash: AlgSHA256, Count: 0})
-	if err != nil {
-		t.Errorf("Certify failed: %s", err)
-		return
-	}
+		t.Run(flag.description, func(t *testing.T) {
+			signerHandle, _, err := CreatePrimary(rw, HandleOwner, pcrSelection7, emptyPassword, defaultPassword, params)
```
chrisfenner, line 611: nit: I don't think this test cares too much about the PCR selection passed to CreatePrimary, could probably leave it out

chrisfenner (dismissed):
This is looking good to me (with just a few nits to pick). Since I'm still a newer reviewer in this repo, I'd prefer for @twitchy-jsonp or @josephlr to take a final pass at reviewing this PR, in cased I missed something. But thank you for the change and for all the incorporations to my feedback!

PR #212, opened 2020-08-18
TPM2: Introduce NVWriteEx for more flexibility

ChriMarMe:
Add NVWriteEx function for more flexibility
Reworked NVWrite to work on top of NVWriteEx

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -1235,29 +1235,34 @@ func NVDefineSpaceEx(rw io.ReadWriter, owner tpmutil.Handle, ownerAuth, authVal 
 	return err
 }
 
-func encodeWriteNV(owner, handle tpmutil.Handle, authString string, data tpmutil.U16Bytes, offset uint16) ([]byte, error) {
-	auth, err := encodeAuthArea(AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession, Auth: []byte(authString)})
-	if err != nil {
-		return nil, err
-	}
-	out, err := tpmutil.Pack(owner, handle)
-	if err != nil {
-		return nil, err
-	}
-	buf, err := tpmutil.Pack(data, offset)
-	if err != nil {
-		return nil, err
-	}
-	return concat(out, auth, buf)
+// NVWrite writes data into the TPM's NV storage.
+func NVWrite(rw io.ReadWriter, authHandle, nvIndex tpmutil.Handle, authString string, data tpmutil.U16Bytes, offset uint16) error {
+	auth := AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession, Auth: []byte(authString)}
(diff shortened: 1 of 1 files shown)


ChriMarMe:
I can't figure out why the test fails.

FAIL: TestNVReadWriteAndLocks (0.00s)
    tpm2_test.go:1212: (not a failure) NVUndefineSpace at index 0x1500000 failed: handle 2, error code 0xb : the handle is not correct for the use
    tpm2_test.go:1243: NVWrite succeeded after NVWriteLock

The NVWrite works on top of NVWriteEx and succeeds, i haven't changed NVWriteLock, but it seems it doesn't lock the Index anymore. I'm clueless here :(

tpm2/tpm2.go
```diff
@@ -1235,30 +1235,35 @@ func NVDefineSpaceEx(rw io.ReadWriter, owner tpmutil.Handle, ownerAuth, authVal
...
-func NVWrite(rw io.ReadWriter, owner, handle tpmutil.Handle, authString string, data tpmutil.U16Bytes, offset uint16) error {
-	Cmd, err := encodeWriteNV(owner, handle, authString, data, offset)
+	b, err := concat(h, authEnc, d)
 	if err != nil {
 		return err
 	}
-	_, err = runCommand(rw, TagSessions, CmdWriteNV, tpmutil.RawBytes(Cmd))
-	return err
+	_, err = runCommand(rw, TagSessions, CmdWriteNV, tpmutil.RawBytes(b))
+	return nil
```
chrisfenner, line 1266: Should this line have `return err` instead? I think this is what is causing the test to fail - your code is doing the right thing but accidentally leaving the final error value behind.

tpm2/tpm2.go
```diff
@@ -1235,30 +1235,35 @@ func NVDefineSpaceEx(rw io.ReadWriter, owner tpmutil.Handle, ownerAuth, authVal
 	return err
 }
 
-func encodeWriteNV(owner, handle tpmutil.Handle, authString string, data tpmutil.U16Bytes, offset uint16) ([]byte, error) {
-	auth, err := encodeAuthArea(AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession, Auth: []byte(authString)})
+// NVWrite writes data into the TPM's NV storage.
+func NVWrite(rw io.ReadWriter, owner, handle tpmutil.Handle, authString string, data tpmutil.U16Bytes, offset uint16) error {
```
chrisfenner, line 1239: nit: can you use the TPM spec names for these handles: `authHandle` and `nvIndex`? Calling the first handle `owner` (which is a problem that predates this PR) is unnecessarily prescriptive. The NV index itself could be the source of the authorization value, e.g., if the NV index has some authorization policy on it.

ChriMarMe:
Thank you. I looked over the code 5 times and didn't see it. Much obliged!

PR #217, opened 2020-08-24
TPM2: Change parameter type from tpmutil.handle to AuthCommand.

### LINKED ISSUE #216: TPM2: NVUndefineSpaceSpecial has wrong parameter type
And another mistake I made, because I didn't understood the spec properly.

NVUndefineSpaceSpecial gets two handles insteat of two AuthCommands. This is wrong. The function can't create AuthCommand structures without further information about the nvIndex.

Possible solution:
It is the users responsibility to create the right conditions for the authentication against the nvIndex authValue or policy, as well as supply the right values for platform authorization to delete the index.
-> change type handle to type AuthCommand

ChriMarMe:
This enables the user to create the right conditions for deleting the nvIndev if TPMA_NV_POLICY_DELETE is set.

This may fix #216 

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -2111,3 +2111,67 @@ func TestPolicyOr(t *testing.T) {
 		t.Errorf("PolicyOr with error: %v", err)
 	}
 }
+
+func TestNVUndefineSpaceSpecial(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	// Set up trial policy session for creation of PolicyCommandCode policy. The policy will be bound as 'delete policy' to the nv index.
+	sess, _, err := StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionTrial, AlgNull, AlgSHA256)
+	if err != nil {
+		t.Errorf("StartAuthSession() for policy generation failed: %v", err)
+	}
+	err = PolicyCommandCode(rw, sess, CmdNVUndefineSpaceSpecial)
+	if err != nil {
+		t.Errorf("PolicyCommandCode() failed: %v", err)
+	}
+	pol, err := PolicyGetDigest(rw, sess)
+	if err != nil {
+		t.Errorf("PolicyGetDigest() failed: %v", err)
+	}
+	err = FlushContext(rw, sess)
+
+	// Test index with
+	pubIndex := NVPublic{
+		NVIndex:    0x1500000,
+		NameAlg:    AlgSHA256,
+		Attributes: AttrPlatformCreate | AttrPolicyDelete | AttrPolicyWrite | AttrNoDA | AttrAuthRead,
+		AuthPolicy: pol,
+		DataSize:   10,
+	}
+	// AuthCommand for platform hierarchy
+	platformAuthCmd := AuthCommand{
+		Session:    HandlePasswordSession,
+		Attributes: AttrContinueSession,
+		Auth:       EmptyAuth,
+	}
+
+	// Create the index
+	err = NVDefineSpaceEx(rw, HandlePlatform, "", pubIndex, platformAuthCmd)
+	if err != nil {
+		t.Errorf("NVDefineSpaceEx() failed: %v", err)
+	}
+
+	// Start new auth session for authentication of nv index access
+	sess, _, err = StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionPolicy, AlgNull, AlgSHA256)
+	if err != nil {
+		t.Errorf("StartAuthSession() for nv index auth failed: %v", err)
+	}
+
+	// Authorize UndefineSpaceSpecial by policy since it can't be authorized by auth value due to the attribute POLICY_DELETE on the index
+	err = PolicyCommandCode(rw, sess, CmdNVUndefineSpaceSpecial)
+	if err != nil {
+		t.Errorf("PolicyCommandCode() or nv index auth failed: %v", err)
+	}
+
+	indexAuthCmd := AuthCommand{
+		Session:    sess,
+		Attributes: AttrContinueSession,
+		Auth:       EmptyAuth,
+	}
+	// Delete the index
+	err = NVUndefineSpaceSpecial(rw, pubIndex.NVIndex, indexAuthCmd, platformAuthCmd)
+	if err != nil {
+		t.Errorf("NVUndefineSpaceSpecial() failed: %v", err)
+	}
+}
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -1194,12 +1194,19 @@ func NVUndefineSpaceEx(rw io.ReadWriter, owner, index tpmutil.Handle, authArea A
 }
 
 // NVUndefineSpaceSpecial This command allows removal of a platform-created NV Index that has TPMA_NV_POLICY_DELETE SET.
-func NVUndefineSpaceSpecial(rw io.ReadWriter, index, platformPP tpmutil.Handle) error {
-	cmd, err := tpmutil.Pack(index, platformPP)
+// The policy to authorize NV index access needs to be created with PolicyCommandCode(rw, sessionHandle, CmdNVUndefineSpaceSpecial) function
+// nvAuthCmd takes the session handle for the policy and the AuthValue (which can be emptyAuth) for the authorization.
+// platformAuth takes either a sessionHandle for the platform policy or HandlePasswordSession and the platformAuth value for authorization.
+func NVUndefineSpaceSpecial(rw io.ReadWriter, nvIndex tpmutil.Handle, nvAuth, platformAuth AuthCommand) error {
+	authBytes, err := encodeAuthArea(nvAuth, platformAuth)
 	if err != nil {
 		return err
 	}
-	_, err = runCommand(rw, TagSessions, CmdNVUndefineSpaceSpecial, tpmutil.RawBytes(cmd))
+	auth, err := tpmutil.Pack(authBytes)
+	if err != nil {
+		return err
+	}
+	_, err = runCommand(rw, TagSessions, CmdNVUndefineSpaceSpecial, nvIndex, HandlePlatform, tpmutil.RawBytes(auth))
 	return err
 }
 


ChriMarMe:
With this:

`NVUndefineSpaceSpecial() failed: handle 2, error code 0x4 : value is out of range or is not correct for the context`

Needs some fixing still

ChriMarMe:
Still no clue

tpm2/tpm2.go
```diff
@@ -1187,12 +1187,18 @@ func NVUndefineSpace(rw io.ReadWriter, ownerAuth string, owner, index tpmutil.Ha
 }
 
 // NVUndefineSpaceSpecial This command allows removal of a platform-created NV Index that has TPMA_NV_POLICY_DELETE SET.
-func NVUndefineSpaceSpecial(rw io.ReadWriter, index, platformPP tpmutil.Handle) error {
-	cmd, err := tpmutil.Pack(index, platformPP)
+func NVUndefineSpaceSpecial(rw io.ReadWriter, nvIndex, sessionIndex tpmutil.Handle, pAuth []byte) error {
```
chrisfenner, line 1190: This should probably take an AuthCommand for each of the authorizations on `nvIndex` and `platform`

chrisfenner, line 1190: Should `sessionIndex` be renamed to `platform` to match the spec?

tpm2/tpm2.go
```diff
@@ -1187,12 +1187,18 @@ func NVUndefineSpace(rw io.ReadWriter, ownerAuth string, owner, index tpmutil.Ha
...
+	indexAuth := AuthCommand{Session: sessionIndex, Attributes: AttrContinueSession, Auth: []byte(EmptyAuth)}
```
chrisfenner, line 1191: There's a couple of things at play here,

- This function assumes that the auth value on the index is the empty auth, but that's not going to be universally true.
- UndefineSpaceSpecial requires authorization by policy that contains an explicit `PolicyCommandCode(CC_UndefineSpaceSpecial)`
- I think the solution is to take an AuthCommand for this and document the function to remind callers that they need to use a specially constructed policy to be allowed to perform this command.

tpm2/tpm2.go
```diff
@@ -1187,12 +1187,18 @@ func NVUndefineSpace(rw io.ReadWriter, ownerAuth string, owner, index tpmutil.Ha
...
+	platformAuth := AuthCommand{Session: HandlePasswordSession, Attributes: AttrContinueSession, Auth: []byte(pAuth)}
```
chrisfenner, line 1192: Platform hierarchy may set up a special policy that allows certain `POLICY_DELETE` NV indices to be deleted, without revealing the full PlatformAuth (which is likely risky) - if we take this as an `AuthCommand` as well, this will help make that possible.

tpm2/tpm2.go
```diff
@@ -1187,12 +1187,18 @@ func NVUndefineSpace(rw io.ReadWriter, ownerAuth string, owner, index tpmutil.Ha
 }
 
 // NVUndefineSpaceSpecial This command allows removal of a platform-created NV Index that has TPMA_NV_POLICY_DELETE SET.
```
chrisfenner, line 1189: Could you please add a unit test for this function? That will make it easier to help debug what you're seeing if there are still problems.

tpm2/tpm2.go
ChriMarMe: I have not figured out how to get access to platform policy, when I only have the PlatformAuth, if you can help me out here I happily implement that.

ChriMarMe: This would be not correct, because the parameter in the function call does not reflect the spec. It is the session handle which authorize the access to the nv index.

ChriMarMe: Nevermind, changed it anyway related to the comments below :)

ChriMarMe:
```
tpm2_test.go:2101: PolicyOr() for policy generation failed: binary.Write: invalid type []tpmutil.U16Bytes
tpm2_test.go:2146: PolicyOr() for nv index auth failed: binary.Write: invalid type []tpmutil.U16Bytes
```

I can't progress until #208 is done.

tpm2/tpm2.go
ChriMarMe: Done

tpm2/tpm2.go
```diff
@@ -1191,12 +1191,19 @@ func NVUndefineSpace(rw io.ReadWriter, ownerAuth string, owner, index tpmutil.Ha
...
-func NVUndefineSpaceSpecial(rw io.ReadWriter, index, platformPP tpmutil.Handle) error {
-	cmd, err := tpmutil.Pack(index, platformPP)
+// The policy to authorize NV index access needs to be created with PolicyCommandCode(rw, sessionHandle, CmdNVUndefineSpaceSpecial) function
+// nvAuthCmd takes the session handle for the policy and the AuthValue (which can be emptyAuth) for the authorization.
+// platformAuth takes either a sessionHandle for the platform policy or HandlePasswordSession and the platformAuth value for authorization.
+func NVUndefineSpaceSpecial(rw io.ReadWriter, nvIndex tpmutil.Handle, nvAuthCmd, platformAuth AuthCommand) error {
```
chrisfenner, line 1197: naming consistency: either these parms should both be named `nvAuthCmd` and `platformAuthCmd` or `nvAuth` and `platformAuth`

chrisfenner (changes requested):
The command looks good to me, just have some feedback on the test which I think can make it even simpler.

tpm2/test/tpm2_test.go
```diff
@@ -2111,3 +2111,87 @@ func TestPolicyOr(t *testing.T) {
 		t.Errorf("PolicyOr with error: %v", err)
 	}
 }
+
+func TestNVUndefineSpaceSpecial(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+
+	// Set up policy session for creation (Trail session)
```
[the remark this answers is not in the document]
ChriMarMe: done

tpm2/test/tpm2_test.go
```diff
@@ -2111,3 +2111,87 @@ func TestPolicyOr(t *testing.T) {
...
+	sess, _, err := StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionTrial, AlgNull, AlgSHA256)
+	if err != nil {
+		t.Errorf("StartAuthSession() for policy generation failed: %v", err)
+	}
+	// Use well known SHA256 values to generate a known policy.
+	hashVal := make([]byte, 32)
+	for iterator := range hashVal {
+		hashVal[iterator] = byte(iterator)
+	}
+	hashNull := make([]byte, 32)
```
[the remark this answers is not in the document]
ChriMarMe: done

tpm2/test/tpm2_test.go
```diff
@@ -2111,3 +2111,87 @@ func TestPolicyOr(t *testing.T) {
...
+	sess, _, err = StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionPolicy, AlgNull, AlgSHA256)
+	if err != nil {
+		t.Errorf("StartAuthSession() for nv index auth failed: %v", err)
+	}
+	// Use PolicyOr to manipulate the session with well known hash values
+	err = PolicyOr(rw, sess, TPMLDigest{Digests: []tpmutil.U16Bytes{tpmutil.U16Bytes(hashVal), tpmutil.U16Bytes(hashNull)}})
+	if err != nil {
+		t.Errorf("PolicyOr() for nv index auth failed: %v", err)
+	}
+	// Use PolicyCommandCode(rw, sess, CmdNVUndefineSpaceSpecial) to gain access to NVUndefineSpaceSpecial
```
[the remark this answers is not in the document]
ChriMarMe: done

tpm2/tpm2.go
ChriMarMe: done

tpm2/test/tpm2_test.go
```diff
@@ -2111,3 +2111,87 @@ func TestPolicyOr(t *testing.T) {
...
+		t.Errorf("StartAuthSession() for policy generation failed: %v", err)
+	}
+	// Use well known SHA256 values to generate a known policy.
+	hashVal := make([]byte, 32)
+	for iterator := range hashVal {
+		hashVal[iterator] = byte(iterator)
+	}
+	hashNull := make([]byte, 32)
+	// Use well known hash for policy generation with PolicyOr
+	err = PolicyOr(rw, sess, TPMLDigest{Digests: []tpmutil.U16Bytes{tpmutil.U16Bytes(hashVal), tpmutil.U16Bytes(hashNull)}})
```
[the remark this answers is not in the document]
ChriMarMe: If I understand the spec correctly PolicyOR hashes all the digests and resets the policyHash of the session to ZeroDigest.
See TPM2 Spec, Part3 - Commnds, Page 227. Especially `Note 2`

It looks like because the policyDigest is set so zeroDigest I have full controll about the policyDigest of the session.

tpm2/test/tpm2_test.go
```diff
@@ -2111,3 +2111,87 @@ func TestPolicyOr(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("NVDefineSpaceEx() failed: %v", err)
+	}
+
+	// Start new auth session for authentication of nv index access
+	sess, _, err = StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionPolicy, AlgNull, AlgSHA256)
+	if err != nil {
+		t.Errorf("StartAuthSession() for nv index auth failed: %v", err)
+	}
+	// Use PolicyOr to manipulate the session with well known hash values
```
chrisfenner, line 2175: So I am unable to reply to the earlier conversation here as it was on an old commit that got force-pushed over. I'll try to paraphrase what you said as now I can't even find the comment (oops!)

> The TPM resets the policy hash to the zeroPolicy, giving me control of the policy so I can use it for this test. See Part 3: Commands, section 23.6, PolicyOr: 'Reset policyDigest to a Zero Digest.' (and also Note 2)

To restate my main concern here: I don't think you need PolicyOR at all in order to test `UndefineSpaceSpecial`. But as TPM 2.0 policies in general (and PolicyOR in particular) are not explained super well in the spec, I'd like to also discuss this a bit as it may help.

Quick summary on policies (sorry if you don't need it 😊 ):

- Policies are expressed as simple hashes that can be extended, much like PCRs.
- Calling a `PolicySomething` command makes one or more assertions about the state of the TPM, and extends the current Policy hash
  - Example: `PolicyPCR` checks to see that the named PCR values are as expected, then extends some info about the PCRS that were checked into the current policy hash (and you can only get the TPM to extend this info by performing this check)
  - Example: `PolicyAuthValue` writes a bit for a future assertion that the command being authorized is also authorized by passing the object's auth value, and extends some info about the check into the current policy hash (and you can only get the TPM to extend this info by performing this check)
- Objects are authorized with policy by executing the exactly-correct policy in the exactly-correct order, and having all those commands succeed.

More details about `PolicyOR`:

- `PolicyOR` turns policies from a 1D array of ANDed-together assertions into a [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree) of policies.
  - As with anything using a Merkle tree, in order to trace the path from the leaf (the particular policy you wanted to use this time) to the root (the actual policy hash on the object), you need some metadata about the rest of the tree. You don't need the whole rest of the tree, but you need all the peer nodes of all the nodes in your leaf-to-root traversal of the tree. i.e., you always have to pass the hashes of all the policies you COULD have satisfied, in order to satisfy one of them and end up with the final policy hash that matches the object whose policy is this tree you are trying to conform to.
- `PolicyOR` is an assertion-then-update-hash like any other `PolicySomething`. But in this case, the assertion is "the current policy hash is one of these up-to-8 values". And the extension logic is special (your note about the spec), because the whole point is to be able to get to the same hash value whether you used branch A or branch B of the `PolicyOR`.

So, the policy hash is _reset to the zeroPolicy_ and **then** extended with the information about **all** the policy branches (uniquely identifying the complete policy tree). That way, the same PolicyOR command executed on top of branchA or branchB results in the same hash - this wouldn't be possible if it simply extended the current policy hash value (or else you'd have to have a collision in your hash algorithm, which would make TPM policies useless garbage).

Hopefully that clarifies why PolicyOR does this reset. Now, to your PR 😁

This test is supposed to test `UndefineSpaceSpecial`. `UndefineSpaceSpecial` says:
> Because administrative role is required, the policy must contain a command that sets the policy command code to `TPM_CC_NV_UndefineSpaceSpecial`.

i.e., you are supposed to use `PolicyCommandCode` or `PolicyCpHash` to specifically announce "This policy is intended for use in authorizing `UndefineSpaceSpecial`! I'm OK with it!". The `Duplicate` command is similarly guarded against shooting yourself in the foot.

So, all you need to do to authorize this command for test is to have your test NV index have a simple policy that is just `PolicyCommandCode(TPM_CC_NV_UndefineSpaceSpecial)` (along with the needed attributes to disallow unspecial `UndefineSpace`.

The PolicyOr in this test is really just making the policy unnecessarily complex. It corresponds to this plain-language policy:

- You can authorize this command if:
  - Either
    - Nothing (the zero policy)
    - Something impossible (no known policy hashes to 0x000102030405...)
  - And
    - The command is UndefineSpaceSpecial (the PolicyCommandCode policy)

The whole "either nothing or something impossible" clause adds to the complexity of the test without adding test coverage (as this is not intended to be a PolicyOR test)

chrisfenner (changes requested):

tpm2/test/tpm2_test.go
ChriMarMe: Thank you good sir for that long explanation. I think i actually learned a lot about policies by it :)

tpm2/test/tpm2_test.go
```diff
@@ -2111,3 +2111,72 @@ func TestPolicyOr(t *testing.T) {
...
+	rw := openTPM(t)
+	defer rw.Close()
+
+	// Set up policy session for creation in a trial session
+	sess, _, err := StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionTrial, AlgNull, AlgSHA256)
+	if err != nil {
+		t.Errorf("StartAuthSession() for policy generation failed: %v", err)
+	}
+
+	// Use PolicyCommandCode(rw, sess, CmdNVUndefineSpaceSpecial) to manipulate the session
```
chrisfenner, line 2125: I think the policy is now so simple that this comment is not needed 😄

I would actually propose replacing the comments on lines 2119, 2125, 2130, and 2135 with one comment before the StartAuthSession that just says "use a trial session to calculate the PolicyCommandCode policy" (or similar), then removing blank lines so that that logical step is grouped nicely into a paragraph of one logical step.

tpm2/test/tpm2_test.go
```diff
@@ -2111,3 +2111,72 @@ func TestPolicyOr(t *testing.T) {
...
+		t.Errorf("StartAuthSession() for nv index auth failed: %v", err)
+	}
+
+	// Authorize UndefineSpaceSpecial by policy since it can't be authorized by auth value due to the attribute POLICY_DELETE on the index
+	err = PolicyCommandCode(rw, sess, CmdNVUndefineSpaceSpecial)
+	if err != nil {
+		t.Errorf("PolicyCommandCode() or nv index auth failed: %v", err)
+	}
+	//
+	// indexAuthCmd for nv index authorization
```
chrisfenner, line 2171: This comment is probably redundant with the code it's about, too

chrisfenner (approved):

tpm2/test/tpm2_test.go
ChriMarMe: done

tpm2/test/tpm2_test.go
ChriMarMe: done

ChriMarMe:
@josephlr Or @twitchy-jsonp can I haz approval from you too?

twitchy-jsonp (approved):

PR #231, opened 2020-12-30
tpm2: fix encoding of name returned by Load()

### LINKED ISSUE #230: Load should return TPM2B_NAME
I have some code that currently looks like this:

```go
privBlob, pubBlob, _, _, _, err := tpm2.CreateKey(f, srk, tpm2.PCRSelection{}, "", "", tmpl)
if err != nil {
    log.Fatalf("create aik: %v", err)
}
aik, nameData, err := tpm2.Load(f, srk, "", pubBlob, privBlob)
if err != nil {
    log.Fatalf("load aik: %v", err)
}
```

I'm currently trying to parse nameData to pass to credactivation.Generate(). But the size prefix is lopped off.

https://github.com/google/go-tpm/blob/fea513a1089448420dc92aae6094e23232e72622/tpm2/tpm2.go#L647

So parsing with DecodeName fails:

```go
tpm2.DecodeName(nameData)
```

That returns

```
unpacking name: decoding Digest: hash algorithm not supported: 0x6e27
```

Okay if I send a fix?

ericchiang:
TPM2_Load() returns a TPM2B_NAME which includes a size prefix. However,
the currently unpacking removes this. Re-encode the name value so it can
be parsed by DecodeName().

Fixes #230

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -199,12 +199,15 @@ func TestCombinedKeyTest(t *testing.T) {
 		t.Fatalf("CreateKey failed: %s", err)
 	}
 
-	keyHandle, _, err := Load(rw, parentHandle, defaultPassword, publicBlob, privateBlob)
+	keyHandle, name, err := Load(rw, parentHandle, defaultPassword, publicBlob, privateBlob)
 	if err != nil {
 		t.Fatalf("Load failed: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


ericchiang:
Searching GitHub, the only users who are using this are just displaying it. They're not actually using the bytes. Most just ignore the return value.

https://github.com/search?l=Go&p=1&q=tpm2.Load%28&type=Code

brandonweeks (approved):

PR #232, opened 2021-01-18
TPM2: Add String function for tpm2.Algorithm

ChriMarMe:
Add String function for tpm2.Algorithm to brighten everyones parsing experience


twitchy-jsonp (changes requested):
Can you make the implementation implement `Stringer`? https://golang.org/pkg/fmt/#Stringer

For the case where the alg is not recognized, can you return something like `fmt.Sprintf("Alg?<%d>", int(a))` ?

Not ideal, but better than the empty string imo.

twitchy-jsonp (approved):
My parsing experience is indeed brightened. Thanks!

PR #235, opened 2021-02-25
Various fixes for Remote Attestation implementations

zaolin:
Summary:

* Fix the wrong usage of signature decoding by using DecodeSignature function (Remote Attestation)
* Fix decodeCertify for wrong signature decoding
* Make EncodeCreationData public
* Introduce OutsideInfo to key creation for key labels
* Improve tpmutil runner with retries to stabilize virtualized TPMs usage

### DIFF
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -674,7 +674,7 @@ func DecodeAttestationData(in []byte) (*AttestationData, error) {
 			return nil, fmt.Errorf("decoding AttestedQuoteInfo: %v", err)
 		}
 	default:
-		return nil, fmt.Errorf("only Ouote, Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
+		return nil, fmt.Errorf("only Quote, Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
 	}
 
 	return &ad, nil
@@ -710,7 +710,7 @@ func (ad AttestationData) Encode() ([]byte, error) {
 			return nil, fmt.Errorf("encoding AttestedQuoteInfo: %v", err)
 		}
 	default:
-		return nil, fmt.Errorf("only Ouote, Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
+		return nil, fmt.Errorf("only Quote, Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
 	}
 
 	return concat(head, signer, tail, info)
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -674,7 +674,7 @@ func DecodeAttestationData(in []byte) (*AttestationData, error) {
 			return nil, fmt.Errorf("decoding AttestedQuoteInfo: %v", err)
 		}
 	default:
-		return nil, fmt.Errorf("only Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
+		return nil, fmt.Errorf("only Ouote, Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
 	}
 
 	return &ad, nil
@@ -705,8 +705,12 @@ func (ad AttestationData) Encode() ([]byte, error) {
 		if info, err = ad.AttestedCreationInfo.encode(); err != nil {
 			return nil, fmt.Errorf("encoding AttestedCreationInfo: %v", err)
 		}
+	case TagAttestQuote:
+		if info, err = ad.AttestedQuoteInfo.encode(); err != nil {
+			return nil, fmt.Errorf("encoding AttestedQuoteInfo: %v", err)
+		}
 	default:
-		return nil, fmt.Errorf("only Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
+		return nil, fmt.Errorf("only Ouote, Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
 	}
 
 	return concat(head, signer, tail, info)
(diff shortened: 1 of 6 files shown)


chrisfenner:
Hi zaolin, I apologize in advance for my slow response time as it is a busy week for me. I would be happy to review this PR and I'm targeting getting back to you by 2021-03-05.

zaolin:
Hi @chrisfenner, no problem at all! :)

tpm2/structures.go
```diff
@@ -674,7 +674,7 @@ func DecodeAttestationData(in []byte) (*AttestationData, error) {
 			return nil, fmt.Errorf("decoding AttestedQuoteInfo: %v", err)
 		}
 	default:
-		return nil, fmt.Errorf("only Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
+		return nil, fmt.Errorf("only Ouote, Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
```
twitchy-jsonp, line 677: s/Ouote/Quote ?

tpm2/structures.go
```diff
@@ -705,8 +705,12 @@ func (ad AttestationData) Encode() ([]byte, error) {
 		if info, err = ad.AttestedCreationInfo.encode(); err != nil {
 			return nil, fmt.Errorf("encoding AttestedCreationInfo: %v", err)
 		}
+	case TagAttestQuote:
+		if info, err = ad.AttestedQuoteInfo.encode(); err != nil {
+			return nil, fmt.Errorf("encoding AttestedQuoteInfo: %v", err)
+		}
 	default:
-		return nil, fmt.Errorf("only Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
+		return nil, fmt.Errorf("only Ouote, Certify & Creation attestation structures are supported, got type 0x%x", ad.Type)
```
twitchy-jsonp, line 713: s/Ouote/Quote ?

tpm2/structures.go
```diff
@@ -827,7 +845,8 @@ type CreationData struct {
 	OutsideInfo         tpmutil.U16Bytes
 }
 
-func (cd *CreationData) encode() ([]byte, error) {
+// EncodeCreationData encodes byte array to TPMS_CREATION_DATA message.
+func (cd *CreationData) EncodeCreationData() ([]byte, error) {
```
twitchy-jsonp, line 849: Why does this need to be exported?

(Not saying its wrong, but would like to understand the use-case. go-attestation package has been able to handle creation data by simply keeping it as bytes and only decoding when necessary, rather than doing round-trips through encode/decode).

tpmutil/run.go
```diff
@@ -41,32 +42,51 @@ func RunCommand(rw io.ReadWriter, tag Tag, cmd Command, in ...interface{}) ([]by
 		return nil, 0, err
 	}
 
-	if _, err := rw.Write(inb); err != nil {
-		return nil, 0, err
-	}
+	// f(t) = (2^t)ms, up to 2s
```
twitchy-jsonp, line 45: This is awesome :D

twitchy-jsonp (changes requested):

tpm2/structures.go
flanfly: We use it to generate TPMS_CREATE_DATA structures for various tests.

zaolin: Also, the DecodeCreationData is already public. :)

zaolin:
@twitchy-jsonp Any update. Does it look good to go now?

tpm2/tpm2.go
```diff
@@ -1746,35 +1753,17 @@ func encodeCertifyEx(objectAuth, signerAuth string, object, signer tpmutil.Handl
...
-	var sigAlg, hashAlg Algorithm
+	var attest tpmutil.U16Bytes
 
 	buf := bytes.NewBuffer(resp)
 	if err := tpmutil.UnpackBuf(buf, &paramSize); err != nil {
 		return nil, nil, err
 	}
 	buf.Truncate(int(paramSize))
-	if err := tpmutil.UnpackBuf(buf, &attest, &sigAlg); err != nil {
+	if err := tpmutil.UnpackBuf(buf, &attest); err != nil {
```
chrisfenner, line 1763: Thanks for this fix. I filed #236 to track the issue you've found and resolved here.

@twitchy-jsonp, @josephlr this unfortunately constitutes a breaking change to the API. How do we want to handle it?

chrisfenner (commented):
This change looks good to me, thank you @zaolin for all these fixes.

Unfortunately since this looks like a breaking change, we will need to strategize about how to merge. @twitchy-jsonp, @josephlr, do you think we should rev a version number?

<shudder> we could introduce another flavor of the Certify and CertifyCreation commands that returns the whole signature. CertifyCreationEx and CertifyEx2? 🤢

twitchy-jsonp:
I'm okay with just bumping a minor version number for this and asking users to make the (small) adjustment.

chrisfenner (approved):
This change looks good to me. If @twitchy-jsonp is OK merging as-is releasing a new minor version, then I support that plan (as opposed to some solution that involves either defining new "fixed" functions)

twitchy-jsonp (approved):

tomoveu:
I just found out about this change and I have a question.

@chrisfenner Is this a "breaking change" or a "small adjustment"? 

Because I thought go-tpm maintains API compatibility. And in my opinion, breaking existing API is a major thing.

@zaolin adding new API **CreateKeyWithOutsideInfo** that calls **create** with the new argument does not help. Because the underlying **create** now brakes API compatibility.

chrisfenner:
Hey @tomoveu - Sorry for the inconvenience! I hope I can help.

**First, a note about versioning and compatibility.**
In general, go-tpm is on semantic major version 0 which means that from time to time, we have to make changes that break the existing API. Whether we use the words "breaking change" or "small adjustment", there's unfortunately change, so we don't do it except when needed (for example, to keep from having to create a lot of new flavors of existing APIs). I'd recommend depending on a specific released version of go-tpm, rather than HEAD of go-tpm main/master branch, so that such changes only affect you when you actively update dependencies, rather than transparently at inconvenient times.

**The breaking API change**
As far as I can see, there is no API compatibility issue with the public `Create` functions. Are you depending on the private `create` function, which took a refactor in this change? The breaking I was referring to was that code that depends on Certify* functions needs to deserialize the second parameter differently, into the [structure](https://github.com/google/go-tpm/blob/4347b8d6b884fa39ac65a6bac410ea1aa50e671c/tpm2/structures.go#L561) that maps to `TPMT_SIGNATURE` (via `DecodeSignature`) instead of a raw RSA/ECC signature with scheme and hash algorithm unknown, to be passed directly to something like `ecdsa.Verify`. I'd be very interested to know if this change causes problems for users of public `Create*` APIs.

PR #241, opened 2021-04-08
Remove Invalid Pack test for ioutil.Discard

### LINKED ISSUE #240: Tests fail on curent master
Right now, running `go test -p 1 ./...` results in the following failure (on go 1.16.2):

```
--- FAIL: TestEncodingPackTypeInvalid (0.00s)
    encoding_test.go:54: The packing function incorrectly succeeds for a non-packable value
```

This looks to be caused by `ioutil.Discard` being a valid value for packing. As of Go 1.16, this value was changed to be `io.Discard` which is just an empty struct (which is a valid value for packing).

josephlr:
Fixes #240

The internal definition of ioutil.Discard is not stable, so we should
not be testing packing with such a value.

Also, the `testEncodingInvalidSlices` function is only used in a
single place, so I just moved its implementation into
`TestEncodingPackTypeInvalid`.

### DIFF
--- a/tpmutil/encoding_test.go
+++ b/tpmutil/encoding_test.go
@@ -27,15 +27,15 @@ type invalidPacked struct {
 	B uint32
 }
 
-func testEncodingInvalidSlices(t *testing.T, f func(io.Writer, interface{}) error) {
+func TestEncodingPackTypeInvalid(t *testing.T) {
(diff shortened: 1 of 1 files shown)


chrisfenner (approved):
Nit: Typo in PR subject: Inavlid->Invalid

PR #247, opened 2021-05-09
Add support for PolicyNV

chrisfenner:
This PR contains two commits:
* one to refactor the current NV test into a bunch of small tests
* one to introduce support for PolicyNV

In the course of testing this change, I found a [reference code bug](https://github.com/microsoft/ms-tpm-20-ref/issues/57). As a result, there are some test cases that can be uncommented-out when that bug is fixed and the change is ingested into the copy of the simulator code used by go-tpm-tools.

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -457,6 +457,7 @@ const (
 	CmdShutdown                   tpmutil.Command = 0x00000145
 	CmdActivateCredential         tpmutil.Command = 0x00000147
 	CmdCertify                    tpmutil.Command = 0x00000148
+	CmdPolicyNV                   tpmutil.Command = 0x00000149
 	CmdCertifyCreation            tpmutil.Command = 0x0000014A
 	CmdReadNV                     tpmutil.Command = 0x0000014E
 	CmdReadLockNV                 tpmutil.Command = 0x0000014F
@@ -547,10 +548,10 @@ var permMap = map[NVAttr]string{
 	AttrAuthRead:       "AuthRead",
 	AttrPolicyRead:     "PolicyRead",
 	AttrNoDA:           "No Do",
-	AttrOrderly:        "Oderly",
+	AttrOrderly:        "Orderly",
 	AttrClearSTClear:   "ClearSTClear",
 	AttrReadLocked:     "ReadLocked",
-	AttrWritten:        "Writte",
+	AttrWritten:        "Written",
 	AttrPlatformCreate: "PlatformCreate",
 	AttrReadSTClear:    "ReadSTClear",
 }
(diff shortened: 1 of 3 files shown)


tpm2/constants.go
```diff
@@ -547,10 +548,10 @@ var permMap = map[NVAttr]string{
 	AttrAuthRead:       "AuthRead",
 	AttrPolicyRead:     "PolicyRead",
 	AttrNoDA:           "No Do",
```
chrisfenner, line 550: I'll fix this, too, along with any requested changes by reviewers.

tpm2/test/tpm2_test.go
```diff
@@ -1341,97 +1341,288 @@ func TestCreateAndCertifyCreationECC(t *testing.T) {
...
+		t.Fatalf("NVDefineSpace failed: %v", err)
+	}
+	defer nvUndefine(rw)
 
-	sessHandle, _, err := StartAuthSession(rw, HandleNull, HandleNull, make([]byte, 16), nil, SessionPolicy, AlgNull, AlgSHA256)
+	if err := NVWriteLock(rw, HandleOwner, nvIdx, emptyPassword); err != nil {
+		t.Fatalf("NVWriteLock failed: %v", err)
+	}
+	if err := nvWrite(rw); err == nil {
+		t.Errorf("NVRead succeeded after NVReadLock")
```
chrisfenner, line 1540: Will fix this error string (should be NVWrite) along with any changes suggested by reviewers.

josephlr:
> In the course of testing this change, I found a [reference code bug](https://github.com/microsoft/ms-tpm-20-ref/issues/57). As a result, there are some test cases that can be uncommented-out when that bug is fixed and the change is ingested into the copy of the simulator code used by go-tpm-tools.

See my comment on your proposed upstream change. If the upstream changes fix this bug, we can just bump `go-tpm-tools/simulator` to current master and then you can have the regression tests un commented out.

josephlr:
@chrisfenner I need to read the spec sections on PolicyNV to understand what sort of API we would want here. So there might be a delay in review (as opposed to the previous delay in review, which was caused by me missing this in my inbox).

tpm2/tpm2.go
```diff
@@ -1456,6 +1456,33 @@ func NVReadLock(rw io.ReadWriter, owner, handle tpmutil.Handle, authString strin
...
 }
 
+// PolicyNV sets NV object state binding for authorization on a session.
+// auth and password must represent a handle and authorization value which is
+// allowed to read the NV index. For example, if the NV index has the
+// "OwnerRead" and "AuthRead" attributes, you can pass either RH_OWNER and the
+// Owner password, or the NV index and the NV's auth value.
+// This function does not support binding to NV indices which cannot be read
+// with an auth value.
+func PolicyNV(rw io.ReadWriter, auth, index, session tpmutil.Handle, password string, operandB tpmutil.U16Bytes, offset uint16, operation EO) error {
```
josephlr, line 1466: Question about parameter ordering, we currently aren't super consistent with how we handle parameter ordering for commands that Read/Write NVData. Per the spec, you always need two handles (`authHandle` and `nvIndex`) and an authorization area corresponding to `authHandle`. Here's the variants we currently have:

- 1 param: `nvIndex`
  - assumes `authHandle = nvIndex`, assumes empty password authorization
  - [`NVRead`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVRead)
- 2 params: `nvIndex`, `authString`
  - assumes `authHandle = nvIndex`, assumes password authorization
  - [`NVIncrement`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVIncrement)
- 3 params: `nvIndex`, `authHandle`, `authString`
  - assumes password authorization
  - [`NVReadEx`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVReadEx)
- 3 params: `authHandle`, `nvIndex`, `authString`
  - assumes password authorization
  - [`NVWrite`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVWrite)
  - [`NVReadLock`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVReadLock)
  - [`NVWriteLock`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVWriteLock)
- 3 params: `authString`, `authHandle`, `nvIndex`
  - assumes password authorization
  - [`NVUndefineSpace`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVUndefineSpace)
- 3 params: `authHandle`, `nvIndex`, `authArea`
  - assumes nothing!
  - [`NVUndefineSpaceEx`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVUndefineSpaceEx)
  - [`NVWriteEx`](https://pkg.go.dev/github.com/google/go-tpm@v0.3.2/tpm2#NVWriteEx)

Obviously, this is something we should clean up for the next breaking change.

So I think the best signature would be something like:
```go
func PolicyNV(rw io.ReadWriter, authHandle, nvIndex tpmutil.Handle, authArea AuthCommand, session tpmutil.Handle, ...) error
```

josephlr: Could we just use `[]byte` for `operandB`? And only use the `tpmutil.U16Bytes` internally?

josephlr: Could we make the last three parameters here a helper struct? They only ever appear together.

chrisfenner:
This got stale :) Work is being done to add 100% coverage of TPM commands in the tpmdirect branch instead.

PR #248, opened 2021-06-01
Add SealEx, an extension of Seal() that accepts 'attributes' argument

### LINKED ISSUE #187: TPM2 Seal command supporting NoDA attribute
The [Seal](https://github.com/google/go-tpm/blob/89fb7a400ac7763655d11809afc01352dcacb22f/tpm2/tpm2.go#L562) function doesn't currently allow the caller to specify TPMA_OBJECT.NoDA to exempt the sealed blob from dictionary attacks. If the caller knows the auth value is not a low-entropy secret like a PIN or human-generated password, the caller should exempt their sealed data from dictionary attack prevention to improve reliability even if the TPM loses power (which counts as a dictionary attack attempt to the TPM)

A `SealEx` or similar command ought to be constructed, allowing the caller to pass the attributes to TPM2_Create.

anatol:
The functionality is similar to the command-line 'tpm2_create -a'
parameter and allows to specify extra sealing attributes.

Closes #187

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -572,10 +572,18 @@ func CreateKeyWithOutsideInfo(rw io.ReadWriter, owner tpmutil.Handle, sel PCRSel
 // password and auth policy. Access to the parent must be available with a simple password.
 // Returns private and public portions of the created object.
 func Seal(rw io.ReadWriter, parentHandle tpmutil.Handle, parentPassword, objectPassword string, objectAuthPolicy []byte, sensitiveData []byte) ([]byte, []byte, error) {
+	return SealEx(rw, parentHandle, parentPassword, objectPassword, FlagSealDefault, objectAuthPolicy, sensitiveData)
+}
+
+// SealEx creates a data blob object that seals the sensitive data under a parent and with a
+// password and auth policy. Access to the parent must be available with a simple password.
+// It also allows to specify the object attributes.
+// Returns private and public portions of the created object.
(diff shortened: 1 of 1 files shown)


josephlr:
Is there a reason to add this method rather than just using `CreateKeyWithSensitive`? If a user wants to provide specific attributes, then I think it makes sense to have them provide the entiere tpm2.Public structure.

anatol:
@josephlr I was not aware of `CreateKeyWithSensitive`.

So I tried following snippet:
```
inPublic := tpm2.Public{
	Type:       tpm2.AlgKeyedHash,
	NameAlg:    tpm2.AlgSHA256,
	Attributes: tpm2.FlagSealDefault | tpm2.FlagUserWithAuth,
	AuthPolicy: policy,
}
privateArea, publicArea, _, _, _, err := tpm2.CreateKeyWithSensitive(dev, srkHandle, tpm2.PCRSelection{}, "", "", inPublic, toSeal)
if err != nil {
	return nil, err
}
```

that corresponds to what I am looking for.

Though for some reason if I run it against `swtpm` it returnes `RCRetry` code. I had to pull `master` branch (ecde5837c1f8 in particular) that adds retry logic to `RunCommand`. With this fix the snippet above works fine for me.

josephlr:
> @josephlr I was not aware of `CreateKeyWithSensitive`.

Ya that's my bad, I should have posted more info on the original bug. @anatol are you ok with me closing this PR (and the Bug)?

> Though for some reason if I run it against `swtpm` it returnes `RCRetry` code.

Ya, that can sometimes happen with certain policies and certain simulators. Real hardware drivers (like the Linux kernel) generally handle this under the hood, so it's not directly visible to applications.

anatol:
> @anatol are you ok with me closing this PR (and the Bug)?

Yes it is fine with me.

chrisfenner:
OK with me too (since I filed that bug).

PR #250, opened 2021-06-10
quote: Remove internal usage of ownerPassword with Quote

josephlr:
Quote only takes one authorization area, so our functions shouldn't take
two passwords. To avoid a breaking change, the function signature of
`Quote` and `QuoteRaw` are note changed, but `encodeQuote` drops the
parameter.

We should remove the superflous param from the public functions when we
next make a breaking change.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -917,12 +917,12 @@ func UnsealWithSession(rw io.ReadWriter, sessionHandle, itemHandle tpmutil.Handl
 	return decodeUnseal(resp)
 }
 
-func encodeQuote(signingHandle tpmutil.Handle, parentPassword string, toQuote tpmutil.U16Bytes, sel PCRSelection, sigAlg Algorithm) ([]byte, error) {
+func encodeQuote(signingHandle tpmutil.Handle, signerAuth string, toQuote tpmutil.U16Bytes, sel PCRSelection, sigAlg Algorithm) ([]byte, error) {
 	ha, err := tpmutil.Pack(signingHandle)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


tpm2/tpm2.go
```diff
@@ -917,7 +917,7 @@ func UnsealWithSession(rw io.ReadWriter, sessionHandle, itemHandle tpmutil.Handl
 	return decodeUnseal(resp)
 }
 
-func encodeQuote(signingHandle tpmutil.Handle, parentPassword, ownerPassword string, toQuote tpmutil.U16Bytes, sel PCRSelection, sigAlg Algorithm) ([]byte, error) {
+func encodeQuote(signingHandle tpmutil.Handle, parentPassword string, toQuote tpmutil.U16Bytes, sel PCRSelection, sigAlg Algorithm) ([]byte, error) {
```
alexmwu, line 920: signerAuth may be a better name, in keeping with Certify and CertifyEx

alexmwu (approved):

tpm2/tpm2.go
```diff
@@ -970,7 +970,8 @@ func Quote(rw io.ReadWriter, signingHandle tpmutil.Handle, parentPassword, owner
 // QuoteRaw is very similar to Quote, except that it will return
 // the raw signature in a byte array without decoding.
 func QuoteRaw(rw io.ReadWriter, signingHandle tpmutil.Handle, parentPassword, ownerPassword string, toQuote []byte, sel PCRSelection, sigAlg Algorithm) ([]byte, []byte, error) {
-	Cmd, err := encodeQuote(signingHandle, parentPassword, ownerPassword, toQuote, sel, sigAlg)
+	// TODO: Remove "ownerPassword" parameter on next breaking change.
```
alexmwu, line 973: also, add this TODO to `Quote` as well

tpm2/tpm2.go
josephlr: Good idea, done.

PR #256, opened 2021-06-29
tpm2: Allow for tpm2.OpenTPM() to use default paths on Linux

josephlr:
This allows users to easily to "the right thing" on Linux, which is to try to open `/dev/tpmrm0` and then fall back to `/dev/tpm0` if the RM doesn't exist. This is what tpm2-tss does.

### DIFF
--- a/tpm2/open_other.go
+++ b/tpm2/open_other.go
@@ -17,8 +17,10 @@
 package tpm2
 
 import (
+	"errors"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/google/go-tpm/tpmutil"
 )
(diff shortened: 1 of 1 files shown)


tpm2/open_other.go
```diff
@@ -17,26 +17,41 @@
...
-func OpenTPM(path string) (io.ReadWriteCloser, error) {
-	rwc, err := tpmutil.OpenTPM(path)
+//
+// This function may also be invoked with no paths, as tpm2.OpenTPM(). In this
+// case, the default paths on Linux (/dev/tpmrm0 and /dev/tpm0), will be used.
+func OpenTPM(path ...string) (tpm io.ReadWriteCloser, err error) {
+	switch len(path) {
+	case 0:
+		tpm, err = tpmutil.OpenTPM("/dev/tpmrm0")
+		if os.IsNotExist(err) {
```
alexmwu, line 38: we should use `errors.Is(err, os.ErrNotExist)` https://golang.org/src/os/error.go?s=3221:3252#L82

alexmwu (approved):

brandonweeks (approved):

alexmwu:
This PR is technically a breaking change, if anyone depends on the signature of OpenTPM on non Windows GOOSes. See https://blog.golang.org/module-compatibility#TOC_2. We may want to update the minor version.

josephlr:
Discussed offline w/ Alex and Maintainers, we will have a minor breaking release of go-tpm to include this and other changes.

PR #258, opened 2021-07-14
Added NVReadLockEx, NVWriteLockEx and changed NVReadEx to support authorization flexibility

p-avaznejad:
Usecase: 
When the disk encryption key is saved in the NV entity, the key can be read during the boot process if the PCRs values are as expected. OS can lock the entity after reading the encryption key to prevent accessing the secret during runtime.

Changes:
- Added NVWriteLockEx, NVReadLockEx, and changed encodeLockNV to support authorization flexibility like Policy PCR.
- Changed encodeNVRead and NVReadEx to support more authorization methods, instead of being limited to the owner's password.
- Changed NVReadEx to enable the user to define the number of octets to be read.
- Added a testcase,TestNVReadWriteAndLockWithPCR, in order to test the features

### DIFF
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -32,7 +32,6 @@ import (
 	"testing"
 
 	"github.com/google/go-tpm-tools/simulator"
-	"github.com/google/go-tpm/tpm2"
 	. "github.com/google/go-tpm/tpm2"
 	"github.com/google/go-tpm/tpmutil"
 )
@@ -1443,14 +1442,14 @@ func TestNVReadWriteAndLockWithPCR(t *testing.T) {
 	}
 
 	// startAuthSession for getting policy digest
-	createPolicySession, _, err := tpm2.StartAuthSession(
+	createPolicySession, _, err := StartAuthSession(
 		rw,               /*rw io.ReadWriter*/
-		tpm2.HandleNull,  /*tpmKey tpmutil.Handle*/
-		tpm2.HandleNull,  /*bindKey tpmutil.Handle*/
+		HandleNull,       /*tpmKey tpmutil.Handle*/
+		HandleNull,       /*bindKey tpmutil.Handle*/
 		make([]byte, 16), /*nonceCaller []byte*/
 		nil,              /*secret []byte*/
 		SessionTrial,     /*se SessionType*/
-		tpm2.AlgNull,     /*sym Algorithm*/
+		AlgNull,          /*sym Algorithm*/
 		AlgSHA256)        /*hashAlg Algorithm*/
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


josephlr:
@p-avaznejad sorry for the delay in reviewing this. It's a breaking change, so we need to figure out how to handle this sort of thing and keep it consistent with the rest of our commands. 

The added tests are great!

The NV commands are a bit of a mess right now, see https://github.com/google/go-tpm/pull/247/files#r660386720 . Ideally we would like to clean them up in such a way that the changes in this PR would also be included.

One potential idea is to define interfaces `Auth`, `AuthHandle`, and `AuthProvision`
```go
type Auth interface {
	AuthCommand() AuthCommand
}

type AuthHandle interface {
	Auth
	AuthHandle() tpmutil.Handle
}

// TPMI_RH_PROVISION w/ Auth
type AuthProvision interface {
	AuthHandle
	isProvision()
}
```

Then we would define types for the different types of authorization:
```go
// These all implement Auth
type PasswordAuth struct{ Password string }
type SessionAuth struct{ Session tpmutil.Handle }
// AuthCommand also implements Auth
```
and the different types of authorized handle:
```go
// All these implement AuthHandle
type EmptyAuth struct{ Handle tpmutil.Handle }
type HandleWithAuth struct {
	Handle tpmutil.Handle
	Auth   Auth
}
// If passed Auth is nil, we just use empty password auth
type OwnerAuth struct{ Auth Auth }
type EndorsementAuth struct{ Auth Auth }
type PlatformAuth struct{ Auth Auth }
```
additionally `OwnerAuth` and `PlatformAuth` would implement `AuthProvision`.

Then we could have the following TPM methods:
```go
func NVDefineSpace(rw io.ReadWriter, owner AuthProvision, authVal string, pubInfo NVPublic) error
func NVUndefineSpace(rw io.ReadWriter, owner AuthProvision, index tpmutil.Handle) error
func NVRead(rw io.ReadWriter,  nvIndex tpmutil.Handle, auth Auth) ([]byte, error)
func NVWrite(rw io.ReadWriter, nvIndex tpmutil.Handle, auth Auth, data []byte, offset uint16) error
func NVReadLock(rw io.ReadWriter, nvIndex tpmutil.Handle, auth Auth) error
func NVWriteLock(rw io.ReadWriter, nvIndex tpmutil.Handle, auth Auth) error
```
The NV methods that take an `Auth`, have some additional checks for `nil` and if the passed `Auth` implements `AuthProvison`.

Then we can use the methods like:
```go
// Define our new space
NVDefineSpace(rw, OwnerAuth{}, "", nvPubWithSessionAuthPolicy)
// Defer undefining our space
NVUndefineSpace(rw, OwnerAuth{}, idx)
// Write to NV with session
NVWrite(rw, idx, SessionAuth{sessHandle}, data, 0)
// Read from NV with session
data, err := NVRead(rw, idx, SessionAuth{sessHandle}) 
// Write to NV with empty auth
NVWrite(rw, idx2, nil, data, 0)
// Read from NV with empty auth
data2, err := NVRead(rw, idx2, nil)
```

PR #268, opened 2022-01-21
Add PCRReset command for tpm2

jkl73:
Add a command to reset a PCR

This will allow us to reset the debug pcr 16 and pcr 23 in a test that requires a clean pcr.

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -454,6 +454,7 @@ const (
 	CmdDictionaryAttackLockReset  tpmutil.Command = 0x00000139
 	CmdDictionaryAttackParameters tpmutil.Command = 0x0000013A
 	CmdPCREvent                   tpmutil.Command = 0x0000013C
+	CmdPCRReset                   tpmutil.Command = 0x0000013D
 	CmdSequenceComplete           tpmutil.Command = 0x0000013E
 	CmdStartup                    tpmutil.Command = 0x00000144
 	CmdShutdown                   tpmutil.Command = 0x00000145
--- a/tpm2/test/tpm2_test.go
+++ b/tpm2/test/tpm2_test.go
@@ -1102,6 +1102,39 @@ func TestReadPCR(t *testing.T) {
 	}
 }
 
+func TestPCRReset(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+	allZeroBytes := make([]byte, 32)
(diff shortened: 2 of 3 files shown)


tpm2/test/tpm2_test.go
```diff
@@ -1102,6 +1102,38 @@ func TestReadPCR(t *testing.T) {
 	}
 }
 
+func TestPCRReset(t *testing.T) {
+	rw := openTPM(t)
+	defer rw.Close()
+	allZeroBytes := make([]byte, 32)
+
+	var fakeHashSum [32]byte
+	err := PCRExtend(rw, 16, AlgSHA256, fakeHashSum[:], "")
```
josephlr, line 1111: We should have PCR 16 defined as a constant here, instead of repeating over and over.

tpm2/test/tpm2_test.go
```diff
@@ -1102,6 +1102,38 @@ func TestReadPCR(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pcrVal, err := ReadPCR(rw, 16 /*pcr*/, AlgSHA256)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if reflect.DeepEqual(allZeroBytes, pcrVal) {
```
josephlr, line 1120: Here and elsewhere, I think this can just be bytes.Equal, they are just byte slices after all

josephlr (approved):

tpm2/constants.go
```diff
@@ -493,6 +493,7 @@ const (
 	CmdPolicyPCR                  tpmutil.Command = 0x0000017F
 	CmdReadClock                  tpmutil.Command = 0x00000181
 	CmdPCRExtend                  tpmutil.Command = 0x00000182
+	CmdPCRReset                   tpmutil.Command = 0x0000013D
```
alexmwu, line 496: nit: this appears to be in sorted order by the CC value. (should go under CmdPCREvent)

alexmwu (approved):

tpm2/test/tpm2_test.go
jkl73: Right. Done.

tpm2/constants.go
jkl73: yes

PR #274, opened 2022-03-10
Make tpm2 quote functions accept PCR selections for multiple banks

hansinator:
The TPM quote commands accept an array of PCR selections but the go-tpm library did not reflect this in its public API, so I added some brackets and dots here and there.

### DIFF
--- a/tpm2/encoding_test.go
+++ b/tpm2/encoding_test.go
@@ -296,7 +296,7 @@ func TestEncodeQuote(t *testing.T) {
 		t.Fatal(err)
 	}
 	toQuote := []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 0xa, 0xb, 0xc, 0xd, 0xe, 0xf, 0x10}
-	cmdBytes, err := encodeQuote(tpmutil.Handle(0x80000001), defaultPassword, toQuote, pcrSelection7, 0x0010)
+	cmdBytes, err := encodeQuote(tpmutil.Handle(0x80000001), defaultPassword, toQuote, []PCRSelection{pcrSelection7}, 0x0010)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/tpm2/structures.go
+++ b/tpm2/structures.go
@@ -823,13 +823,13 @@ func (ci CertifyInfo) encode() ([]byte, error) {
 
 // QuoteInfo represents a TPMS_QUOTE_INFO structure.
 type QuoteInfo struct {
-	PCRSelection PCRSelection
+	PCRSelection []PCRSelection
 	PCRDigest    tpmutil.U16Bytes
 }
 
 func decodeQuoteInfo(in *bytes.Buffer) (*QuoteInfo, error) {
 	var out QuoteInfo
-	sel, err := decodeOneTPMLPCRSelection(in)
+	sel, err := decodeTPMLPCRSelection(in)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


jkl73:
Hi, `PCRSelection` struct already contains an array of PCR banks, you can use it with multiple PCRs like
`tpm2.PCRSelection{Hash: tpm2HashAlgo, PCRs: []int{1,2,3}}`

hansinator:
@jkl73 Are you sure? Your example does contain an array of PCR indices, but no array for PCR banks.

jkl73:
Ah I see, I misunderstood. I was thinking multiple PCR indices... (instead of banks/hashes).
The Change LGTM, but it may break some downstream packages which use `Quote` or `QuoteRaw`

hansinator:
Quote and QuoteRaw already have an unused parameter and accompanying comments that say to remove them on the next breaking API change. Release history shows that this library doesn't forever keep its API stable and the last point release had a breaking change. So I thought I'd propose a more tidy solution first.

I could also change the PR and introduce something like QuoteEx and QuoteRawEx having the new signatures and let the other two functions forward to these.

josephlr:
I agree that matching the spec here would be reasonable when we next make a breaking change to Quote/QuoteRaw/etc... I'm just not sure when that will be. @chrisfenner do we think we'll have another breaking change before the TPM-direct stuff is ready?

josephlr:
> I agree that matching the spec here would be reasonable when we next make a breaking change to Quote/QuoteRaw/etc... I'm just not sure when that will be. @chrisfenner do we think we'll have another breaking change before the TPM-direct stuff is ready?

hansinator:
We are using the library in a production setting with customers. Unfortunately, we can't use a dev branch.

In practice our customers are having systems where
a) multiple PCR banks are active and hashed into
b) multiple banks are active but not with the same set of registers, some Dell servers f.e. have SHA2 active with no registers (thus no hashes) and SHA1 active with a full set of registers

We can advise people to configure their UEFI settings correctly, but they typically won't want to when they deploy our software to their servers. So we have to work with what we get and that is the use-case for quoting multiple PCR banks. We have >10MB of extra data in a quote and we don't want round-trips with our remote attestation servers, so we need to quote everything there is in one go.

Personally, I feel a library should act as a tool and not impose policies, so I'd reason that way when thinking about making it possible to quote multiple PCR banks.

chrisfenner:
@hansinator that seems reasonable, and we get you need to work with what you have. We would be in favor of making this change the next time we release a breaking change for go-tpm. I think that it will at least be a little bit before the Direct API is ready, so I don't want you to have to wait on that.

So I can get a sense of urgency here, what is the cost to calling `tpm2.Quote` twice (first with all the SHA1 PCRs, second with all the SHA256 PCRs)? It seemed like you had some additional costs on the verification side?

Re: this:

> b) multiple banks are active but not with the same set of registers, some 2021 produced Dell servers f.e. have SHA2 active with no registers (thus no hashes, not an empty unextented register, but literally the registers are completely missing) and SHA1 active with a full set of registers

That's very weird to me, and it implies a firmware bug if not exactly the same issue I was worried about ([CVE-2021-42299](https://github.com/google/security-research/blob/master/pocs/bios/tpm-carte-blanche/writeup.md)). I'd like to follow up about it, would you be willing to shoot me a [mail](cfenn@google.com)?

hansinator:
@chrisfenner We are prepared to use a soft-fork in the meantime, even if it takes a while.

PR #277, opened 2022-05-24
Implement and tested go-tpm direct function TPM2_GetRandom

matt-tsai:


direct/tpm2/get_random_test.go
```diff
@@ -0,0 +1,24 @@
...
+	defer thetpm.Close()
+
+	grc := GetRandom{
+		BytesRequested: 16,
+	}
+
+	if _, err := grc.Execute(thetpm); err != nil {
+		t.Fatalf("GetRandom failed: %v", err)
+	}
+
```
chrisfenner, line 23: nit: can delete this blank line

direct/tpm2/tpm2.go
```diff
@@ -355,6 +355,33 @@ type CreateLoadedResponse struct {
...
+// Execute executes the command and returns the response.
+func (cmd *GetRandom) Execute(t transport.TPM, s ...Session) (*GetRandomResponse, error) {
+	var rsp GetRandomResponse
+	if err := execute(t, cmd, &rsp, s...); err != nil {
+		return nil, err
+	}
+	return &rsp, nil
+}
+
+// GetRandomReponse is the reponse from TPM2_GetRandom
```
chrisfenner, line 376: nit: end this sentence in a period

chrisfenner (approved):
Looks great, thanks for the change! I have two tiny nits.

direct/tpm2/tpm2.go
```diff
@@ -355,6 +355,33 @@ type CreateLoadedResponse struct {
 // Response implements the Response interface.
 func (*CreateLoadedResponse) Response() tpm.CC { return tpm.CCCreateLoaded }
 
+// GetRandom is the input to TPM2_GetRandom.
+// See definition in Part 3, Commands, section 16.1
+type GetRandom struct {
+	BytesRequested uint16
```
alexmwu, line 361: nit: add the description "number of octets to return" from the spec as a comment

direct/tpm2/get_random_test.go
```diff
@@ -0,0 +1,24 @@
+package tpm2
```
alexmwu, line 1: (not a blocking comment) how do we want to organize tests? having a test per command may make the directory quite large. we could have a test directory like in `tpm2` or we could split by use case.

alexmwu (approved):

direct/tpm2/get_random_test.go
chrisfenner: My mild opinion is that I like having more and smaller test files. A test directory with e.g., one file per command tested would be groovy.

alexmwu: I generally agree this makes tests easier to find, and we should go for this. I do wonder about tests that use multiple commands like a potential policy auth test with multiple policies.

chrisfenner: Right, and we'd have a lot of helpers too. So we might have some per-command tests as well as some "integration-y" tests that really cover a bunch of commands. That's a good reason not to be rigid on "one file per command" but in general I think it's ok to have a lot of files.

chrisfenner (approved):

chrisfenner:
LGTM - Matt can you update your branch with a rebase?

PR #279, opened 2022-05-25
Read public

matt-tsai:
Implemented and Tested ReadPublic

### DIFF
--- a/direct/tpm2/read_public_test.go
+++ b/direct/tpm2/read_public_test.go
@@ -1,0 +1,76 @@
+package tpm2
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/google/go-tpm/direct/structures/tpm"
+	"github.com/google/go-tpm/direct/structures/tpm2b"
+	"github.com/google/go-tpm/direct/structures/tpma"
+	"github.com/google/go-tpm/direct/structures/tpms"
+	"github.com/google/go-tpm/direct/structures/tpmt"
+	"github.com/google/go-tpm/direct/structures/tpmu"
+	"github.com/google/go-tpm/direct/transport/simulator"
+)
+
+func TestReadPublicKey(t *testing.T) {
+
+	thetpm, err := simulator.OpenSimulator()
+	if err != nil {
+		t.Fatalf("could not connect to TPM simulator: %v", err)
+	}
+	defer thetpm.Close()
+
+	// CreatePrimary Struct Params copied from policy_test.go
+	createPrimary := CreatePrimary{
+		PrimaryHandle: tpm.RHOwner,
+		InPublic: tpm2b.Public{
+			PublicArea: tpmt.Public{
+				Type:    tpm.AlgECC,
+				NameAlg: tpm.AlgSHA256,
+				ObjectAttributes: tpma.Object{
+					FixedTPM:            true,
+					FixedParent:         true,
+					SensitiveDataOrigin: true,
+					UserWithAuth:        true,
+					SignEncrypt:         true,
+				},
+				Parameters: tpmu.PublicParms{
+					ECCDetail: &tpms.ECCParms{
+						Scheme: tpmt.ECCScheme{
+							Scheme: tpm.AlgECDSA,
+							Details: tpmu.AsymScheme{
+								ECDSA: &tpms.SigSchemeECDSA{
+									HashAlg: tpm.AlgSHA256,
+								},
+							},
+						},
+						CurveID: tpm.ECCNistP256,
+					},
+				},
+			},
+		},
+	}
+
+	rspCP, err := createPrimary.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("CreatePrimary failed: %v", err)
+	}
+
+	readPublic := ReadPublic{
+		ObjectHandle: rspCP.ObjectHandle,
+	}
+
+	rspRP, err := readPublic.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("ReadPublic failed: %v", err)
+	}
+
+	rspCPXBuffer := rspCP.OutPublic.PublicArea.Unique.ECC.X.Buffer
+	rspRPXBuffer := rspRP.OutPublic.PublicArea.Unique.ECC.X.Buffer
+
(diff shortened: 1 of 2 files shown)


direct/tpm2/tpm2.go
```diff
@@ -290,6 +290,38 @@ type LoadExternalResponse struct {
...
+	var rsp ReadPublicResponse
+	if err := execute(t, cmd, &rsp, s...); err != nil {
+		return nil, err
+	}
+	return &rsp, nil
+}
+
+// ReadPublicResponse is the response from TPM2_ReadPublic.
+type ReadPublicResponse struct {
+	// public area of an object
```
chrisfenner, line 314: nit: can these comments be verbatim copies of the "description" column in the spec?

direct/tpm2/read_public_test.go
```diff
@@ -0,0 +1,76 @@
...
+									HashAlg: tpm.AlgSHA256,
+								},
+							},
+						},
+						CurveID: tpm.ECCNistP256,
+					},
+				},
+			},
+		},
+	}
```
chrisfenner, line 53: Can we `defer` a `FlushContext` similar to what we do here: https://github.com/google/go-tpm/blob/7cd0fbd6e8fe921362249ce85533e0839dffb68a/direct/tpm2/policy_test.go#L53-L61
Something like:
```go
defer func() { 
 	flush := FlushContext{ 
 		FlushHandle: rsp.ObjectHandle, 
 	} 
 	if err := flush.Execute(thetpm); err != nil { 
 		t.Errorf("could not flush signing key: %v", err) 
 	} 
 }()
```

direct/tpm2/read_public_test.go
```diff
@@ -0,0 +1,76 @@
...
+	rspRP, err := readPublic.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("ReadPublic failed: %v", err)
+	}
+
+	rspCPXBuffer := rspCP.OutPublic.PublicArea.Unique.ECC.X.Buffer
+	rspRPXBuffer := rspRP.OutPublic.PublicArea.Unique.ECC.X.Buffer
+
+	// not comprehensive as the function to encode public area is not implemented yet
+	if !bytes.Equal(rspCPXBuffer, rspRPXBuffer) {
```
chrisfenner, line 73: How about using `cmp.Equal` to compare the two `Unique`s?

chrisfenner (changes requested):
Thanks for the change! This looks pretty solid, I just have some minor notes (mostly about the test)

direct/tpm2/read_public_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestReadPublicKey(t *testing.T) {
+
+	thetpm, err := simulator.OpenSimulator()
+	if err != nil {
+		t.Fatalf("could not connect to TPM simulator: %v", err)
+	}
+	defer thetpm.Close()
+
+	// CreatePrimary Struct Params copied from policy_test.go
```
alexmwu, line 24: nit: unnecessary comment here

direct/tpm2/read_public_test.go
alexmwu: @chrisfenner 
Sorry to make this more convoluted. Common commands like FlushContext take up more lines of code than `go-tpm/tpm2` currently does. Users likely wouldn't want the clutter of management commands like FlushContext taking too much space; they care about the business logic of what they're putting into commands like CreatePrimary or ReadPublic.

Would having the idiom below be easier to read?
```
defer func() {
  if err := FlushContext{FlushHandle: rsp.ObjectHandle}.Execute(thetpm); err != nil {
    t.Errorf("...")
  }
}
```

we could also make wrappers for commands like FlushContext, although I think it seems unnecessary for a small command like FlushContext.
e.g. in a util pkg or something:
```
direct/tpm2/directutil/context
func FlushContext(t transport.TPM, flushHandle handle) error {
  return FlushContext{ FlushHandle: flushHandle}.Execute(t)
}

direct/tpm2/read_public_test.go:
defer func() {
  if err := directutil.FlushContext(thetpm, rsp.ObjectHandle); err != nil {
    t.Errorf("...")
  }
}
```

alexmwu (approved):

chrisfenner:
Not at all @alexmwu, I totally agree with the feedback. It would be awesome to have a one-line idiom for "defer flush this thing" and even awesomer to have a test version that asserts that the flush succeeded.

I'm in email/meeting mode all day today so I can't remember if this would compile but can we have something along the lines of
```go
defer t.Run("🚽", &FlushContext{rsp.ObjectHandle}.Execute(thetpm) == nil)
```
in the tests and
```go
defer &FlushContext{rsp.ObjectHandle}.Execute(thetpm))
```
in sample code?

If that's not possible with the current reflection logic, are there tweaks we could make to make it possible?

direct/tpm2/read_public_test.go
chrisfenner: We discussed offline - we don't have to assert that every flush succeeded during the tests (the current tests don't; we just need a FlushContext test that checks it once).

So, I'd like to update my suggestion to just: can we add
```go
defer &FlushContext{rsp.ObjectHandle}.Execute(thetpm))
```
here?

matt-tsai: I was getting 
`function must be invoked in defer statement` for 

> `defer &FlushContext{rsp.ObjectHandle}.Execute(thetpm))`

So my work around was
```
flushContext := FlushContext{FlushHandle: rspCP.ObjectHandle}
defer flushContext.Execute(thetpm)
```
which seems to work just as well

alexmwu: Can you try `defer (&FlushContext{rsp.ObjectHandle}).Execute(thetpm))`

matt-tsai: The same error occurs.

chrisfenner: I'm fine with the 2 line version here, and if we think of a neat idiom later it's easy to change.

chrisfenner (approved):

PR #280, opened 2022-05-26
ContextSave and ContextLoad

matt-tsai:
Added necessary Structures for ContextSave and ContextLoad. Implemented and tested both functions.

### DIFF
--- a/direct/structures/internal/structures.go
+++ b/direct/structures/internal/structures.go
@@ -397,6 +397,10 @@ type TPMISHPolicy = TPMHandle
 // See definition in Part 2: Structures, section 9.11.
 type TPMIDHContext = TPMHandle
 
+// TPMIDHSaved represents a TPMI_DH_SAVED.
+// See definition in Part 2: Structures, section 9.12.
+type TPMIDHSaved = TPMHandle
+
 // TPMIRHHierarchy represents a TPMI_RH_HIERARCHY.
 // See definition in Part 2: Structures, section 9.13.
 type TPMIRHHierarchy = TPMHandle
@@ -1610,6 +1614,42 @@ type TPMSNVPublic struct {
 // See definition in Part 2: Structures, section 13.6.
 type TPM2BNVPublic struct {
 	NVPublic TPMSNVPublic `gotpm:"sized"`
+}
+
+// TPM2BContextSensitive represents a TPM2B_CONTEXT_SENSITIVE
+// See definition in Part 2: Structures, section 14.2.
+type TPM2BContextSensitive struct {
+	Size uint16
+	// the sensitive data
+	Buffer []byte
+}
+
+// TPMSContextData represents a TPMS_CONTEXT_DATA
+// See definition in Part 2: Structures, section 14.3.
+type TPMSContextData struct {
+	// the integrity value
+	Integrity TPM2BDigest
+	// the sensitive area
+	Encrypted TPM2BContextSensitive
+}
+
+// TPM2BContextData represents a TPM2B_CONTEXT_DATA
+// See definition in Part 2: Structures, section 14.4.
+type TPM2BContextData struct {
+	Buffer TPMSContextData `gotpm:"sized"`
+}
+
+// TPMSContext represents a TPMS_CONTEXT
+// See definition in Part 2: Structures, section 14.5.
+type TPMSContext struct {
+	// the sequence number of the context
+	Sequence uint64
+	// a handle indicating if the context is a session, object, or sequence object
+	SavedHandle TPMIDHSaved
+	// the hierarchy of the context
+	Hierarchy TPMIRHHierarchy
+	// the context data and integrity HMAC
+	ContextBlob TPM2BContextData
 }
 
 // TPM2BCreationData represents a TPM2B_CREATION_DATA.
--- a/direct/structures/tpms/tpms.go
+++ b/direct/structures/tpms/tpms.go
@@ -187,6 +187,14 @@ type ECCParms = internal.TPMSECCParms
 // See definition in Part 2: Structures, section 13.5.
 type NVPublic = internal.TPMSNVPublic
 
+// ContextData represents a TPMS_CONTEXT_DATA
+// See definition in Part 2: Structures, section 14.3.
+type ContextData = internal.TPMSContextData
+
+// Context represents a TPMS_CONTEXT
+// See definition in Part 2: Structures, section 14.5.
+type Context = internal.TPMSContext
+
 // CreationData represents a TPMS_CREATION_DATA.
 // See definition in Part 2: Structures, section 15.1.
 type CreationData = internal.TPMSCreationData
--- a/direct/tpm2/combined_context_test.go
+++ b/direct/tpm2/combined_context_test.go
@@ -7,44 +7,73 @@ import (
 	"github.com/google/go-tpm/direct/structures/tpm"
 	"github.com/google/go-tpm/direct/structures/tpm2b"
 	"github.com/google/go-tpm/direct/structures/tpma"
+	"github.com/google/go-tpm/direct/structures/tpml"
 	"github.com/google/go-tpm/direct/structures/tpms"
 	"github.com/google/go-tpm/direct/structures/tpmt"
 	"github.com/google/go-tpm/direct/structures/tpmu"
+	"github.com/google/go-tpm/direct/transport"
 	"github.com/google/go-tpm/direct/transport/simulator"
 )
 
-func CombinedContextTest(t *testing.T) {
+func ReadPublicName(t *testing.T, handle tpm.Handle, thetpm transport.TPM) tpm2b.Name {
+	readPublic := ReadPublic{
+		ObjectHandle: handle,
+	}
+
+	rspRP, err := readPublic.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("Failed to read public: %v", err)
+	}
+
+	return rspRP.Name
+}
+
+func TestCombinedContext(t *testing.T) {
 	thetpm, err := simulator.OpenSimulator()
 	if err != nil {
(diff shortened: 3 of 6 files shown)


direct/structures/internal/structures.go
```diff
@@ -1591,6 +1591,44 @@ type TPM2BNVPublic struct {
...
 
+// TPM2BContextSensitive represents a TPM2B_CONTEXT_SENSITIVE
+// See definition in Part 2: Structures, section 14.2.
+type TPM2BContextSensitive struct {
+	Size uint16
+	// the sensitive data
+	Buffer []byte
+}
+
+// TPMSContextData represents a TPMs_CONTEXT_DATA
```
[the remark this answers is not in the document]
matt-tsai: Fixed.

direct/structures/internal/structures.go
```diff
@@ -1591,6 +1591,44 @@ type TPM2BNVPublic struct {
...
+}
+
+// TPMSContext represents a TPMS_CONTEXT
+// See definition in Part 2: Structures, section 14.5.
+type TPMSContext struct {
+	// the sequence number of the context
+	// Transient object contexts and session contexts used different counters.
+	Sequence uint64
+	// a handle indicating if the context is a session, object, or sequence object
+	SavedHandle TPMIDHContext
```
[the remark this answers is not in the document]
matt-tsai: Just to confirm, I should add TPMI_DH_SAVED even though it is not in Part 2? If so, under which section should I add to the comments?

direct/structures/internal/structures.go
```diff
@@ -1591,6 +1591,44 @@ type TPM2BNVPublic struct {
...
+	Integrity TPM2BDigest
+	// the sensitive area
+	Encrypted TPM2BContextSensitive
+}
+
+// TPM2BContextData represents a TPM2B_CONTEXT_DATA
+// See definition in Part 2: Structures, section 14.4.
+type TPM2BContextData struct {
+	Size   uint16
+	Buffer []byte
```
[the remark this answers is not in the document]
matt-tsai: Fixed.

direct/structures/tpms/tpms.go
```diff
@@ -183,6 +183,14 @@ type ECCParms = internal.TPMSECCParms
 // See definition in Part 2: Structures, section 13.5.
 type NVPublic = internal.TPMSNVPublic
 
+// ContextData represents a TPMS_CONTEXT_DATA
+// See definition in Part 2: Structures, section 14.3.
+type ContextData = internal.TPMSContextData
+
+// Context represents a TPMS_CONTEXT
+// See definition in Part 2: Structures, section 14.5.
+type Context = internal.TPMSContextData
```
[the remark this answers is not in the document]
matt-tsai: Fixed.

direct/structures/internal/structures.go
```diff
@@ -1591,6 +1591,44 @@ type TPM2BNVPublic struct {
 	NVPublic TPMSNVPublic `gotpm:"sized"`
 }
 
+// TPM2BContextSensitive represents a TPM2B_CONTEXT_SENSITIVE
+// See definition in Part 2: Structures, section 14.2.
+type TPM2BContextSensitive struct {
```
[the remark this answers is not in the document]
matt-tsai: Fixed

direct/structures/internal/structures.go
```diff
@@ -1591,6 +1591,44 @@ type TPM2BNVPublic struct {
...
+type TPM2BContextData struct {
+	Size   uint16
+	Buffer []byte
+}
+
+// TPMSContext represents a TPMS_CONTEXT
+// See definition in Part 2: Structures, section 14.5.
+type TPMSContext struct {
+	// the sequence number of the context
+	// Transient object contexts and session contexts used different counters.
```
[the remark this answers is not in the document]
chrisfenner: I agree, usually they aren't critical.

A user using this code probably has the spec and has seen it (they know the names of the commands etc). Including one line of copied descriptions for things from the spec is supposed to help refresh/remind users so they don't have to go back and forth *as* much, but for nuanced things (where notes, parentheticals matter) - they have to look at the spec anyway.

direct/structures/internal/structures.go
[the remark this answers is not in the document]
chrisfenner: Looks like that was added for 1.59

direct/tpm2/combined_context_test.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		t.Fatalf("CreatePrimary failed: %v", err)
+	}
+
+	flushContextCP := FlushContext{FlushHandle: rspCP.ObjectHandle}
+	defer flushContextCP.Execute(thetpm)
+
+	cl := CreateLoaded{
+		ParentHandle: tpm.RHOwner,
+		InPublic: tpm2b.Template{
```
chrisfenner, line 63: if you want, you can create a template as a local variable and pass it both times

direct/tpm2/combined_context_test.go
```diff
@@ -0,0 +1,118 @@
...
+	contextLoad := ContextLoad{
+		Context: rspCS.Context,
+	}
+
+	rspCoL, err := contextLoad.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("ContextLoad failed: %v", err)
+	}
+
+	if !cmp.Equal(rspCoL.LoadedHandle, rspCrL.ObjectHandle) {
```
chrisfenner, line 115: Probably it's more correct to call ReadPublic here and confirm that the same object is loaded both times. We might not get the same handle value when we loaded it again (the TPM gets to decide that)

It's totally sufficient to compare just the Names, don't have to check every field

chrisfenner (changes requested):
The only change that matters here is the "read public" suggestion (just to make the test safe for future use with "real" tpms

direct/structures/internal/structures.go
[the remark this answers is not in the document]
matt-tsai: Non-critical comments removed.

direct/tpm2/combined_context_test.go
matt-tsai: The keys are different enough where they might be useful for future references.

direct/tpm2/combined_context_test.go
matt-tsai: This fix is taking longer than usual as I encountered a bug that I haven't seen before.

matt-tsai: Fixed to compare names.

direct/tpm2/combined_context_test.go
```diff
@@ -0,0 +1,97 @@
...
+	contextLoad := ContextLoad{
+		Context: rspCS.Context,
+	}
+
+	rspCL, err := contextLoad.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("ContextLoad failed: %v", err)
+	}
+
+	if !cmp.Equal(rspCL.LoadedHandle.KnownName(), rspCP.ObjectHandle.KnownName()) {
```
chrisfenner, line 94: KnownName is for when the caller knows the expected name of the object; calling it on a loaded handle should produce nil.

Instead, can you call ReadPublic on the `rspCL.LoadedHandle` and compare it to `rspCP.Name`?

direct/tpm2/reflect.go
```diff
@@ -20,7 +20,7 @@ import (
 const (
 	// Chosen based on MAX_DIGEST_BUFFER, the length of the longest
```
chrisfenner, line 21: Can you update this comment explaining that maxListLength needs to be greater than MAX_CONTEXT_SIZE, which in the reference implementation is 1344?

chrisfenner (changes requested):
Overall looks great. I have only a couple minor concerns.

direct/tpm2/combined_context_test.go
matt-tsai: Alright I will make that change now.

direct/tpm2/reflect.go
matt-tsai: Updated!

direct/tpm2/combined_context_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	contextLoad := ContextLoad{
+		Context: rspCS.Context,
+	}
+
+	rspCL, err := contextLoad.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("ContextLoad failed: %v", err)
+	}
```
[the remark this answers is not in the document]
matt-tsai: Flushed.

direct/tpm2/combined_context_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/google/go-tpm/direct/structures/tpml"
+	"github.com/google/go-tpm/direct/structures/tpms"
+	"github.com/google/go-tpm/direct/structures/tpmt"
+	"github.com/google/go-tpm/direct/structures/tpmu"
+	"github.com/google/go-tpm/direct/transport"
+	"github.com/google/go-tpm/direct/transport/simulator"
+)
+
+func ReadPublicName(t *testing.T, handle tpm.Handle, thetpm transport.TPM) tpm2b.Name {
+	readPublic := ReadPublic{
```
chrisfenner, line 19: Femto-nit: You can use [t.Helper()](https://pkg.go.dev/testing#T.Helper) here so that failures are printed with the line number of whatever test called this helper instead of the line number instead of the helper.

(I don't actually care too much about this one test, but for future reference this is a handy tool)

chrisfenner (approved):

direct/tpm2/combined_context_test.go
matt-tsai: Okay, I will be sure to keep this in mind for future test writing. Thank you.

chrisfenner (approved):

PR #283, opened 2022-06-10
Supporting functionalities for direct migration

matt-tsai:
Main changes include: ReadWrite interface for transport TPMs, adding util to grab HandleValue.

### DIFF
--- a/direct/tpm2/tpm2.go
+++ b/direct/tpm2/tpm2.go
@@ -660,8 +660,7 @@ type Sign struct {
 	InScheme tpmt.SigScheme
 	// proof that digest was created by the TPM
 	// If keyHandle is not a restricted signing key, then this
-	// may be a NULL Ticket with tag =
-	// TPM_ST_CHECKHASH.
+	// may be a NULL Ticket with tag = TPM_ST_CHECKHASH.
 	Validation tpmt.TKHashCheck
 }
 
--- a/direct/transport/open_other.go
+++ b/direct/transport/open_other.go
@@ -17,12 +17,12 @@ import (
 //
 // This function may also be invoked with no paths, as tpm2.OpenTPM(). In this
 // case, the default paths on Linux (/dev/tpmrm0 then /dev/tpm0), will be used.
-func OpenTPM(path ...string) (TPM, error) {
+func OpenTPM(path ...string) (TPMCloser, error) {
 	rwc, err := legacy.OpenTPM(path...)
 	if err != nil {
 		return nil, err
 	}
-	return &LocalTPM{
+	return &wrappedRWC{
 		transport: rwc,
 	}, nil
 }
--- a/direct/transport/tpm.go
+++ b/direct/transport/tpm.go
@@ -18,50 +18,49 @@ type TPMCloser interface {
 	io.Closer
 }
 
-// The wrappedRW represents a struct that wraps an io.ReadWriter
-// to function like a transport.TPM.
+
+// wrappedRW represents a struct that wraps an io.ReadWriter
+// to a transport.TPM to be compatible with tpmdirect.
 type wrappedRW struct {
 	transport io.ReadWriter
 }
 
-// The wrappedRWC represents a struct that wraps an io.ReadWriteCloser
-// to function like a transport.TPM with the close function.
+// wrappedRWC represents a struct that wraps an io.ReadWriteCloser
+// to a transport.TPM to be compatible with tpmdirect.
 type wrappedRWC struct {
 	transport io.ReadWriteCloser
 }
 
-// The wrappedTPM represents a struct that wraps a transport.TPM 
-// to function like a io.ReadWriter
+// wrappedTPM represents a struct that wraps a transport.TPM's underlying 
+// transport to use with legacy code that expects an io.ReadWriter.
 type wrappedTPM struct {
 	response []byte
 	tpm      TPM
 }
 
-// FromReadWritter takes in a io.ReadWriter TPM and returns a
-// transport.TPm wrapping the io.ReadWriter.
+// FromReadWriter takes in a io.ReadWriter and returns a
+// transport.TPM wrapping the io.ReadWriter.
 func FromReadWriter(rw io.ReadWriter) TPM {
 	return &wrappedRW{transport: rw}
 }
 
-// ToReadWritter takes in a transport TPM and returns an
+// ToReadWriter takes in a transport TPM and returns an
 // io.ReadWriter wrapping the transport TPM.
 func ToReadWriter(tpm TPM) io.ReadWriter {
 	return &wrappedTPM{tpm: tpm}
 }
 
 // Read copies t.response into the p buffer and return the appropriate length.
-func (t *wrappedTPM) Read(p []byte) (n int, err error) {
-	p = t.response
-	return len(p), nil
+func (t *wrappedTPM) Read(p []byte) (int, error) {
+	n := copy(p, t.response)
+	t.response = t.response[n:]
+	if len(t.response) == 0 {
+		return n, io.EOF
+	}
+	return n, nil
 }
 
-// Write writes len(p) bytes from p to the underlying data stream.
-// It returns the number of bytes written from p (0 <= n <= len(p))
-// and any error encountered that caused the write to stop early.
-// Write must return a non-nil error if it returns n < len(p).
-// Write must not modify the slice data, even temporarily.
-//
-// Implementations must not retain p.
+// Write implements the io.ReadWriter interface.
 func (t *wrappedTPM) Write(p []byte) (n int, err error) {
 	t.response, err = t.tpm.Send(p)
 	if err != nil {
--- a/direct/transport/tpm.go
+++ b/direct/transport/tpm.go
@@ -10,20 +10,77 @@ import (
 // TPM represents a logical connection to a TPM.
 type TPM interface {
 	Send(input []byte) ([]byte, error)
-	Close() error
 }
 
-// LocalTPM represents a connection to the local TPM.
-type LocalTPM struct {
+// TPMCloser represents a logical connection to a TPM and you can close it.
+type TPMCloser interface {
+	TPM
+	io.Closer
+}
+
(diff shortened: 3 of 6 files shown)


direct/transport/tpm.go
```diff
@@ -10,14 +10,32 @@ import (
...
 	Send(input []byte) ([]byte, error)
+}
+
+// TPMCloser represents a logical connection to a TPM and you can close it!!!.
+type TPMCloser interface {
+	TPM
 	Close() error
 }
 
+type tempTPM struct {
```
matt-tsai, line 21: Needs a better name...

direct/tpm2/tpm2.go
```diff
@@ -534,7 +534,7 @@ type Sign struct {
 	// If keyHandle is not a restricted signing key, then this
 	// may be a NULL Ticket with tag =
 	// TPM_ST_CHECKHASH.
-	Validation tpmt.TKHashCheck
+	Validation tpmt.TKHashCheck
```
josephlr, line 537: Why whitespace here?

direct/transport/tpm.go
josephlr: `compatTPM`? legacyTPM?

josephlr (commented):
One idea worth considering here is which direction we want to support conversion for the TPM types. We can support:
  1. Using the new TPM type (`transport.TPM`) with the old TPM interface (`io.ReadWriter`/`io.ReadWriteCloser`) and library.
  2. Using the old TPM type (`io.ReadWriter`) with the new TPM interface (`transport.TPM`) and library

This PR implements (2), but I'm wondering if we should implement (1) instead.

matt-tsai:
> One idea worth considering here is which direction we want to support conversion for the TPM types. We can support:
> 
> 1. Using the new TPM type (`transport.TPM`) with the old TPM interface (`io.ReadWriter`/`io.ReadWriteCloser`) and library.
> 2. Using the old TPM type (`io.ReadWriter`) with the new TPM interface (`transport.TPM`) and library
> 
> This PR implements (2), but I'm wondering if we should implement (1) instead.

This PR implements 2 because the old tests still need RWC and the Direct Migration needs a RW in order to be made a transport tpm used for Execute. Correct me if I'm wrong but eventually we will only have 1 interface. 
I think the option (option 2 in this case) where we change RW into transport TPMs might be wiser as it would require less changes to direct and would also be backwards compatible.

direct/tpm2/tpm2.go
matt-tsai: Removed whitespace.

direct/transport/tpm.go
alexmwu: seeing as we already have `LocalTPM`, we could also rename that to `LocalTPMCloser` and name this one `LocalTPM`

chrisfenner: I'll contribute to the bikeshed :)

If I understand @josephlr's out-of-band comments, we should return non-public types that happen to satisfy public interfaces. Therefore we need:

Our interfaces:
`TPM`
`TPMCloser`

Our private type names for each of these:
`localTPM` (implements both TPM and TPMCloser)
`wrappedTPM` (implements only TPM)
No wrappedTPMCloser because we shouldn't close a TPM whose readwriter we're borrowing

Don't need public type names like `WrappedTPM` or `LocalTPM` (so we can delete `LocalTPM` entirely). Just the public functions that return `TPM` or `TPMCloser` (`OpenSimulator`, `OpenTPM`, `WrapReadWriter`)

direct/transport/open_other.go
```diff
@@ -22,7 +22,7 @@ func OpenTPM(path ...string) (TPM, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &LocalTPM{
```
josephlr, line 25: Here (and in open_windows.go) we need the `OpenTPM` function to return `TPMCloser`.

direct/tpm2/tpm2.go
```diff
@@ -532,8 +532,7 @@ type Sign struct {
 	InScheme tpmt.SigScheme
 	// proof that digest was created by the TPM
 	// If keyHandle is not a restricted signing key, then this
-	// may be a NULL Ticket with tag =
-	// TPM_ST_CHECKHASH.
+	// may be a NULL Ticket with tag = TPM_ST_CHECKHASH.
```
josephlr, line 535: Is this change supposed to be in this PR?

direct/transport/tpm.go
matt-tsai: I think Joe's comment makes the most sense to me too. :) Thank you for the suggestions

direct/transport/open_other.go
matt-tsai: Fixed.

direct/transport/tpm.go
```diff
@@ -10,20 +10,77 @@ import (
...
+}
+
+// The wrappedRWC represents a struct that wraps an io.ReadWriteCloser
+// to function like a transport.TPM with the close function.
+type wrappedRWC struct {
 	transport io.ReadWriteCloser
 }
 
+// The wrappedTPM represents a struct that wraps a transport.TPM 
+// to function like a io.ReadWriter
```
chrisfenner, line 34: nit: Maybe rephrase, like:

wrappedTPM represents a struct that wraps a transport.TPM's underlying transport to use with legacy code that expects an io.ReadWriter

direct/transport/tpm.go
```diff
@@ -10,20 +10,77 @@ import (
...
+type wrappedTPM struct {
+	response []byte
+	tpm      TPM
+}
+
+// FromReadWritter takes in a io.ReadWriter TPM and returns a
+// transport.TPm wrapping the io.ReadWriter.
```
chrisfenner, line 41: nit: transport.TPM

direct/transport/tpm.go
```diff
@@ -10,20 +10,77 @@ import (
...
 }
 
+// The wrappedTPM represents a struct that wraps a transport.TPM 
+// to function like a io.ReadWriter
+type wrappedTPM struct {
+	response []byte
+	tpm      TPM
+}
+
+// FromReadWritter takes in a io.ReadWriter TPM and returns a
```
chrisfenner, line 40: Nit: FromReadWriter

chrisfenner (approved):
Thanks for the change, looks great. Only teeny nits

direct/transport/tpm.go
matt-tsai: 100% agreed, I was having trouble describing it myself, thank you!

direct/transport/tpm.go
matt-tsai: Fixed

direct/transport/tpm.go
matt-tsai: Fixed.

direct/transport/tpm.go
```diff
@@ -10,20 +10,74 @@ import (
...
+
+// add a toreadwrite from transport
+func ToReadWriter(tpm TPM) io.ReadWriter {
+	return &wrappedTPM{tpm: tpm}
+}
+
+// Not sure what else to do here? 
+func (t *wrappedTPM) Read(p []byte) (n int, err error) {
+	return len(p), nil
+}
```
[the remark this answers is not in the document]
matt-tsai: Fixed.

direct/transport/tpm.go
```diff
@@ -10,20 +10,77 @@ import (
...
+	return &wrappedTPM{tpm: tpm}
+}
+
+// Read copies t.response into the p buffer and return the appropriate length.
+func (t *wrappedTPM) Read(p []byte) (n int, err error) {
+	p = t.response
+	return len(p), nil
+}
+
+// Write writes len(p) bytes from p to the underlying data stream.
```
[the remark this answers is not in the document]
matt-tsai: Got it, updated this.

chrisfenner (approved):

direct/transport/tpm.go
```diff
@@ -55,13 +56,7 @@ func (t *wrappedTPM) Read(p []byte) (n int, err error) {
...
 }
 
-// Write writes len(p) bytes from p to the underlying data stream.
-// It returns the number of bytes written from p (0 <= n <= len(p))
-// and any error encountered that caused the write to stop early.
-// Write must return a non-nil error if it returns n < len(p).
-// Write must not modify the slice data, even temporarily.
-//
-// Implementations must not retain p.
+// Write implements the io.ReadWriter interface
```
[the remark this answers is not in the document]
matt-tsai: Thanks for letting me know, I will keep this in mind going forward! Also it was been fixed.

direct/transport/tpm.go
```diff
@@ -10,20 +10,72 @@ import (
...
+
+// ToReadWriter takes in a transport TPM and returns an
+// io.ReadWriter wrapping the transport TPM.
+func ToReadWriter(tpm TPM) io.ReadWriter {
+	return &wrappedTPM{tpm: tpm}
+}
+
+// Read copies t.response into the p buffer and return the appropriate length.
+func (t *wrappedTPM) Read(p []byte) (n int, err error) {
+	p = t.response
```
josephlr, line 55: So I don't think this works, using this implementation we won't actually write any bytes to p. This just overwrites p temporarily in the function, but doesn't actually do anything. Example:
https://go.dev/play/p/Lqdl8wT_1n1

I think we will need to either use a `*bytes.Buffer` as the internal response type (example: https://go.dev/play/p/JW4V94-0Blu), or have an implementation that looks like:
```go
func (t *wrappedTPM) Read(p []byte) (int, error) {
	n := copy(p, t.response)
	t.response = t.response[n:]
	if len(t.response) == 0 {
		return n, io.EOF
	}
	return n, nil
}
```
see this example: https://go.dev/play/p/Hppzq-rQETY
which uses [the `copy` builtin](https://pkg.go.dev/builtin@go1.18.3#copy)

matt-tsai: Fixed, should work as intended now.

josephlr:
Nice work on this!!!

PR #289, opened 2022-07-06
PolicyPCR 

matt-tsai:
Finished implementing PolicyPCR. Covered positive, negative, and void test cases.


direct/tpm2/tpm2.go
```diff
@@ -902,6 +902,38 @@ type PolicyOrResponse struct{}
...
+	var rsp PolicyPCRResponse
+	return execute(t, cmd, &rsp, s...)
+}
+
+// Update implements the PolicyCommand interface.
+func (p *PolicyPCR) Update(policy *PolicyCalculator) error {
+	return policy.Update(tpm.CCPolicyPCR, p.Pcrs, p.PcrDigest.Buffer)
+}
+
+// PolicyOrResponse is the response from TPM2_PolicyPCR.
```
chrisfenner, line 931: Should be PolicyPCRResponse

matt-tsai: Good catch.

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,189 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+	expectedDigest := sha1.Sum(expectedVal)
+	t.Logf("expectedDigest=%x", expectedDigest)
+
+	// Use the policy helper to calculate the same policy
+	pol, err := NewPolicyCalculator(tpm.AlgSHA1)
+	if err != nil {
+		t.Fatalf("creating policy calculator: %v", err)
+	}
+	
+	pol.Update(tpm.CCPolicyPCR, policyPCR.Pcrs, expectedDigest)
```
matt-tsai, line 459: Work around for void case, feedback is welcome.

chrisfenner, line 459: Instead of calling Update ourselves (which we don't expect callers to actually directly do) how about assembling a PolicyPCR instance with pcrDigest populated, and passing that one to the calculator. The one actually invoked in the trial session should have pcrDigest omitted.

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,189 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+		expectedDigest []byte
+		expectedResult bool
+	}{
+		{desc: "success: expectedDigest passed in expectedDigest", expectedDigest: expectedDigest[:], expectedResult: true},
+		{desc: "fail: wrongDigest passed in expectedDigest", expectedDigest: wrongDigest[:], expectedResult: false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			policyPCR := PolicyPCR{
+				PolicySession: sess.Handle(),
```
chrisfenner, line 350: We probably should call PolicyRestart (or just flush and re-create the session) across iterations of this loop. If the second case was a positive-result case, it'd fail because `PolicyPCR(x) ⋅ PolicyPCR(y)` [where `⋅` represents the implicit AND of chaining multiple policy commands in one session] is a different policy than `PolicyPCR(y)` by itself

chrisfenner (changes requested):

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,189 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+	}
+
+	pcrReadRsp, err := pcrRead.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("Failed to read PCRSelection")
+	}
+
+	var expectedVal []byte
+	for _, digest := range pcrReadRsp.PCRValues.Digests {
+		expectedVal = append(expectedVal, digest.Buffer...)
```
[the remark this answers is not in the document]
josephlr: So this line seems fine an idiomatic to me.

`s1 = append(s1, s2...)` is the standard way to append the elements of s2 to s1. See https://go.dev/blog/slices-intro (search for "append one slice to another")

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,189 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+	}
+	t.Logf("expectedVal=%x", expectedVal)
+
+	// Hash algorithm must match the one in PolicySession.
+	expectedDigest := sha1.Sum(expectedVal)
+	t.Logf("expectedDigest=%x", expectedDigest)
+
+	wrongCase := append(expectedVal, []byte("wrongCase")...)
```
chrisfenner, line 336: Here is the line I meant to attach the above comment to.

direct/tpm2/policy_test.go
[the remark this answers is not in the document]
chrisfenner: Oops, I'm sorry. I made this comment on the wrong append line. Please see below, where we have

wrongCase := append(expectedVal, ...)

Since this will modify expectedVal if it needed to be reallocated, that could lead to problems.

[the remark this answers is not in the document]
matt-tsai: Okay that makes more sense. I will resolve this by making the wrongDigest a Hash of the expectedDigest.

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,189 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+	}()
+
+	PCRs, err := CreatePCRSelection([]int{0, 3, 7})
+	if err != nil {
+		t.Fatalf("Failed to create PCRSelection")
+	}
+	selection := tpml.PCRSelection{
+		PCRSelections: []tpms.PCRSelection{
+			{
+				Hash:      tpm.AlgSHA1,
```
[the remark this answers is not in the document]
matt-tsai: Resolved by having the void testcase test SHA256.

direct/tpm2/policy_test.go
matt-tsai: This has been resolved by having the wrongDigest be the expectedDigest hashed an additional time.

direct/tpm2/policy_test.go
matt-tsai: Resolved by creating a populated PolicyPCR (policyPCRPopulated) for the calculator.

direct/tpm2/policy_test.go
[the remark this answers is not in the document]
alexmwu: One thing could be to add a nested loop that goes through the algorithms we support, so that we get coverage on hash calculations as well. 
something like:
```
        tpmAlgs := []tpm.HashAlg{tpm.AlgSHA1, tpm.AlgSHA256}
	tests := []struct {
		desc           string
		expectedDigest []byte
		expectedResult bool
	}{
		{desc: "success: expectedDigest passed in expectedDigest", expectedDigest: expectedDigest[:], expectedResult: true},
		{desc: "fail: wrongDigest passed in expectedDigest", expectedDigest: wrongDigest[:], expectedResult: false},
	}
	for _, tt := range tests {	tests := []struct {
		desc           string
		expectedDigest []byte
		expectedResult bool
	}{
		{desc: "success: expectedDigest passed in expectedDigest", expectedDigest: expectedDigest[:], expectedResult: true},
		{desc: "fail: wrongDigest passed in expectedDigest", expectedDigest: wrongDigest[:], expectedResult: false},
	}
	for _, tt := range tests {
          for _, tpmAlg := range tpmAlgs {
// test logic here.
```

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,189 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+	}
+
+	var expectedVal []byte
+	for _, digest := range pcrReadRsp.PCRValues.Digests {
+		expectedVal = append(expectedVal, digest.Buffer...)
+	}
+	t.Logf("expectedVal=%x", expectedVal)
+
+	// Hash algorithm must match the one in PolicySession.
+	expectedDigest := sha1.Sum(expectedVal)
```
alexmwu, line 333: This is heavily duplicated in the two tests. Could we pull it out into a helper?

```
func getExpectedPCRDigest(t *testing.T, tpm transport.TPM, pcrs []int, alg tpm.HashAlg) []byte {
```

inside you can also convert the tpm.HashAlg to a crypto.Hash to get the digest

direct/tpm2/policy_test.go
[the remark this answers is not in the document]
chrisfenner: The reason I did not suggest this is that since the purpose of this is to test the software stack and not the TPM, we don't have to be exhaustive about covering every single algorithm in every single case. We just need to make sure that go-tpm marshals commands and unmarshals responses correctly, and IMHO as long as 2 or more hash algorithms are represented across all the tests for a command, we have plenty of coverage of that. That said, if you disagree, I won't spend a lot of effort arguing against even more thorough testing :)

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,189 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+
+	wrongCase := append(expectedVal, []byte("wrongCase")...)
+	wrongDigest := sha1.Sum(wrongCase)
+
+	tests := []struct {
+		desc           string
+		expectedDigest []byte
+		expectedResult bool
+	}{
+		{desc: "success: expectedDigest passed in expectedDigest", expectedDigest: expectedDigest[:], expectedResult: true},
```
alexmwu, line 344: nit: since you're defining an anon struct and using every field, you don't have to use the field names. This makes it a bit cleaner to read. 
```
	tests := []struct {
		desc           string
		expectedDigest []byte
		expectedResult bool
	}{
		{"success: expectedDigest passed in expectedDigest", expectedDigest[:], true},
...
```

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,189 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+			got := pol.Hash()
+
+			if bytes.Equal(got.Digest, want.PolicyDigest.Buffer) != tt.expectedResult {
+				t.Errorf("policyPCR.Hash() = %x,\nwant %x", got.Digest, want.PolicyDigest.Buffer)
+			}
+		})
+	}
+}
+
+// Testing void input for PcrDigest, from TPM 2.0 Part 3, 23.7.
```
alexmwu, line 384: @chrisfenner, do you know in practice how often TPMs follow:
"If the caller does not
provide PCR settings (pcrDigest has a length of zero), the TPM may (and it is preferred to) use the
current TPM PCR settings (digestTPM) in the calculation for the new policyDigest."
vs "may return an error"?

chrisfenner: Great question. I actually learned that TPMs did this because of reviewing this change - I think it is safer and more normal to always specify the hash. That said, go-tpm should let the user do what they want, and if they want to use this feature (and rely on a TPM doing it) we should behave properly. Since the reference code does it, it seems good to me to test it.

direct/tpm2/policy_test.go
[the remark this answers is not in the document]
alexmwu: I do not have strong feelings on the matter except that we should calculate a SHA256 digest somewhere with the PolicyCalculator. If we already have, great! Otherwise in these tests, I believe it's all SHA1 now.

direct/tpm2/policy_test.go
matt-tsai: I will pull out the process needed to produce an expected digest, but I think everything before that is essential to code readability and testing down the line. Ex. selection is used in the table tests.

direct/tpm2/policy_test.go
alexmwu: Yeah this was more for my curiosity. No change needed here

direct/tpm2/policy_test.go
matt-tsai: Fixed.

direct/tpm2/policy_test.go
matt-tsai: Resolved. Pulled out process of reading PCR selection and processing to get PCRDigest.

direct/tpm2/policy_test.go
matt-tsai: Resolved, but ran into trouble when passing in the wrong policy. Ended up deciding to have the policy get executed first then change the digest prior to the policy calculator.

direct/tpm2/policy_test.go
```diff
@@ -282,25 +282,35 @@ func TestPolicyOrUpdate(t *testing.T) {
 	}
 }
 
+func getExpectedPCRDigest(t *testing.T, thetpm transport.TPM, selection tpml.PCRSelection, hashAlg tpm.AlgID) []byte {
+	pcrRead := PCRRead{
```
chrisfenner, line 286: nit: This is a good place to try out `t.Helper()`

direct/tpm2/policy_test.go
```diff
@@ -282,25 +282,35 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+	var expectedVal []byte
+	for _, digest := range pcrReadRsp.PCRValues.Digests {
+		expectedVal = append(expectedVal, digest.Buffer...)
+	}
+	t.Logf("expectedVal=%x", expectedVal)
+
+	cryptoHashAlg, err := hashAlg.Hash()
+	hash := cryptoHashAlg.New()
+	hash.Write(expectedVal)
+	return hash.Sum([]byte{})
```
chrisfenner, line 304: nit: `hash.Sum(nil)`

direct/tpm2/policy_test.go
```diff
@@ -282,25 +282,35 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+	pcrReadRsp, err := pcrRead.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("Failed to read PCRSelection")
+	}
+
+	var expectedVal []byte
+	for _, digest := range pcrReadRsp.PCRValues.Digests {
+		expectedVal = append(expectedVal, digest.Buffer...)
+	}
+	t.Logf("expectedVal=%x", expectedVal)
```
chrisfenner, line 299: nit: This log line may not be needed.

direct/tpm2/policy_test.go
```diff
@@ -314,42 +324,32 @@ func TestPolicyPCRUpdate(t *testing.T) {
...
-
-	// Hash algorithm must match the one in PolicySession.
-	expectedDigest := sha1.Sum(expectedVal)
+	expectedDigest := getExpectedPCRDigest(t, thetpm, selection, tpm.AlgSHA1)
 	t.Logf("expectedDigest=%x", expectedDigest)
 
-	wrongCase := append(expectedVal, []byte("wrongCase")...)
-	wrongDigest := sha1.Sum(wrongCase)
+	wrongDigest := sha1.Sum(expectedDigest[:])
+	t.Logf("wrongDigest=%x", wrongDigest)
```
chrisfenner, line 331: nit: Probably don't need to log this

direct/tpm2/policy_test.go
```diff
@@ -366,6 +366,10 @@ func TestPolicyPCRUpdate(t *testing.T) {
 				t.Fatalf("executing PolicyGetDigest: %v", err)
 			}
 
+			if tt.expectedResult == false {
```
chrisfenner, line 369: nit: `if !tt.expectedResult`

direct/tpm2/policy_test.go
```diff
@@ -314,42 +324,32 @@ func TestPolicyPCRUpdate(t *testing.T) {
...
 
-	wrongCase := append(expectedVal, []byte("wrongCase")...)
-	wrongDigest := sha1.Sum(wrongCase)
+	wrongDigest := sha1.Sum(expectedDigest[:])
+	t.Logf("wrongDigest=%x", wrongDigest)
 
 	tests := []struct {
 		desc           string
 		expectedDigest []byte
 		expectedResult bool
```
chrisfenner, line 336: I'm not really sure what this expected result is? When it is false, the test code seems to check that changing the policyPCR's PCRDigest field causes the policy not to match, but that's not as interesting as making sure that the TPM and the PolicyCalculator always agree on what a policy hash is, for a given policy. IMHO, I think you could take out the failure case and make this a flat, one-test test.

chrisfenner (changes requested):
This is looking really good. Thanks for your thoroughness! I have a few more nits (sorry) and one comment about `TestPolicyPCRUpdate`'s negative test case.

direct/tpm2/policy_test.go
matt-tsai: Resolved.

direct/tpm2/policy_test.go
matt-tsai: Resolved.

direct/tpm2/policy_test.go
matt-tsai: Resolved.

direct/tpm2/policy_test.go
matt-tsai: This was helpful for my debugging, but yes I agree. Resolved.

direct/tpm2/policy_test.go
matt-tsai: Resolved.

direct/tpm2/policy_test.go
matt-tsai: After discussing with Joe, I decided to rewrite the whole test covering 6 cases, Trial correct, incorrect, empty, Real correct, incorrect, empty.

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,139 @@ func TestPolicyOrUpdate(t *testing.T) {
...
 
+func getExpectedPCRDigest(t *testing.T, thetpm transport.TPM, selection tpml.PCRSelection, hashAlg tpm.AlgID) []byte {
+	t.Helper()
+	pcrRead := PCRRead{
+		PCRSelectionIn: selection,
+	}
+
+	pcrReadRsp, err := pcrRead.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("Failed to read PCRSelection")
```
chrisfenner, line 293: (sorry to miss this before) nit: this should be something like "failed to read PCRs"

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,139 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+	}
+
+	var expectedVal []byte
+	for _, digest := range pcrReadRsp.PCRValues.Digests {
+		expectedVal = append(expectedVal, digest.Buffer...)
+	}
+
+	cryptoHashAlg, err := hashAlg.Hash()
```
chrisfenner, line 301: nit: check err

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,139 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+		PCRSelections: []tpms.PCRSelection{
+			{
+				Hash:      tpm.AlgSHA1,
+				PCRSelect: PCRs,
+			},
+		},
+	}
+
+	expectedDigest := getExpectedPCRDigest(t, thetpm, selection, tpm.AlgSHA1)
+	t.Logf("expectedDigest=%x", expectedDigest)
```
chrisfenner, line 328: nit: don't need this log line anymore for the final version of the code. (We should only print when there is a mismatch)

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,139 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+			}
+
+			err = policyPCR.Execute(thetpm)
+			if tt.callShouldSucceed {
+				if err != nil {
+					t.Fatalf("executing PolicyPCR: %v", err)
+				}
+			} else {
+				if err == nil {
+					t.Fatalf("failing test PolicyPCR is passing")
```
chrisfenner, line 369: Might need to rephrase this, something like "expected PolicyPCR to return error, got nil"

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,139 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+				t.Fatalf("executing PolicyGetDigest: %v", err)
+			}
+
+			// If the pcrDigest is empty: see TPM 2.0 Part 3, 23.7.
+			if tt.pcrDigest == nil {
+				expectedDigest := getExpectedPCRDigest(t, thetpm, selection, tpm.AlgSHA1)
+				t.Logf("expectedDigest=%x", expectedDigest)
+
+				// Create a populated policyPCR for the PolicyCalculator
+				policyPCR = PolicyPCR{
```
chrisfenner, line 388: nit: you can simplify by just setting `policyPCR.PcrDigest.Buffer = expectedDigest[:]` instead of recreating the entire object

direct/tpm2/policy_test.go
```diff
@@ -281,6 +282,139 @@ func TestPolicyOrUpdate(t *testing.T) {
...
+				t.Fatalf("creating policy calculator: %v", err)
+			}
+			policyPCR.Update(pol)
+			got := pol.Hash()
+
+			if !bytes.Equal(got.Digest, want.PolicyDigest.Buffer) {
+				t.Errorf("policyPCR.Hash() = %x,\nwant %x", got.Digest, want.PolicyDigest.Buffer)
+			}
+
+			t.Helper()
```
chrisfenner, line 409: nit: shouldn't have Helper() here.

chrisfenner (approved):
This looks great to me! I have some very minor (some optional) nits and then I think we're ready to merge this in.

direct/tpm2/policy_test.go
matt-tsai: Resolved.

PR #293, opened 2022-07-12
Implemented and Tested TPM2_Certify

matt-tsai:


direct/tpm2/certify_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	rspSigner, err := createPrimarySigner.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("Failed to create primary: %v", err)
+	}
+	flushContextSigner := FlushContext{FlushHandle: rspSigner.ObjectHandle}
+	defer flushContextSigner.Execute(thetpm)
+
+	createPrimarySubject := CreatePrimary{
+		PrimaryHandle: tpm.RHPlatform,
```
chrisfenner, line 91: Seems a little strange to use Platform hierarchy here. This will only work against the simulator since a well-behaved BIOS will disable or change the Platform auth. Can we use Owner hierarchy but make sure the keys are different (e.g., by putting some string into InPublic.Unique)?

direct/tpm2/certify_test.go
```diff
@@ -0,0 +1,147 @@
...
+			Name:   rspSubject.Name,
+			Auth:   PasswordAuth(Auth),
+		},
+		SignHandle: AuthHandle{
+			Handle: rspSigner.ObjectHandle,
+			Name:   rspSigner.Name,
+			Auth:   PasswordAuth(Auth),
+		},
+		QualifyingData: tpm2b.Data{
+			Buffer: nil,
```
chrisfenner, line 121: Might be nice to include a random 16-byte nonce here as an example.

chrisfenner (changes requested):
This looks great. I have only a couple very minor suggestions to improve the "example code" qualities of the test.

direct/tpm2/certify_test.go
```diff
@@ -0,0 +1,153 @@
...
+			Name:   rspSubject.Name,
+			Auth:   PasswordAuth(Auth),
+		},
+		SignHandle: AuthHandle{
+			Handle: rspSigner.ObjectHandle,
+			Name:   rspSigner.Name,
+			Auth:   PasswordAuth(Auth),
+		},
+		QualifyingData: tpm2b.Data{
+			Buffer: make([]byte, 16),
```
chrisfenner, line 127: Can we put something in this buffer and then verify that we see that same something in the signed attestation?

direct/tpm2/certify_test.go
```diff
@@ -0,0 +1,153 @@
...
+				UserAuth: tpm2b.Auth{
+					Buffer: Auth,
+				},
+			},
+		},
+		InPublic:    public,
+		CreationPCR: pcrSelection,
+	}
+	uniqueIdentifier := tpmu.PublicID{
+		KeyedHash: &tpm2b.Digest{
```
chrisfenner, line 103: We should be selecting the RSA member of this union since the type of public indicates RSA, not KeyedHash.

chrisfenner (changes requested):

direct/tpm2/certify_test.go
matt-tsai: Resolved.

direct/tpm2/certify_test.go
matt-tsai: Resolved.

direct/tpm2/certify_test.go
```diff
@@ -0,0 +1,162 @@
...
+	createPrimarySubject.InPublic.PublicArea.Unique = unique
+
+	rspSubject, err := createPrimarySubject.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("Failed to create primary: %v", err)
+	}
+	flushContextSubject := FlushContext{FlushHandle: rspSubject.ObjectHandle}
+	defer flushContextSubject.Execute(thetpm)
+
+	originalBuffer := make([]byte, 16)
```
chrisfenner, line 117: nit: could this be

```
originalBuffer := []byte("test nonce")
```

to make it clear that this is for nonces, and decrease the chance of the test accidentally passing despite some mistake in extraData marshaling/unmarshalling

chrisfenner (approved):
Looks great, just one more change (basically a nit about the contents of the nonce for the test, to make it clearer and more likely to catch certain types of bugs)

direct/tpm2/certify_test.go
matt-tsai: Makes sense, resolved.

PR #294, opened 2022-07-19
Implemented and Tested Clear.

matt-tsai:
Implemented and tested Clear. Fixed some code spacing. Clear will be helping in testing other functions so more comprehensive tests will come with the migration of more features.

### DIFF
--- a/direct/tpm2/clear_test.go
+++ b/direct/tpm2/clear_test.go
@@ -1,0 +1,25 @@
+package tpm2
+
+import (
+	"testing"
+
+	"github.com/google/go-tpm/direct/structures/tpm"
+	"github.com/google/go-tpm/direct/transport/simulator"
+)
+
+func TestClear(t *testing.T) {
+
+	thetpm, err := simulator.OpenSimulator()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


direct/tpm2/clear_test.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+	defer thetpm.Close()
+
+	clear := Clear{
+		AuthHandle: tpm.RHLockout,
+	}
+
+	if _, err := clear.Execute(thetpm); err != nil {
+		t.Fatalf("Clear failed: %v", err)
+	}
```
josephlr, line 24: Is it possible to have some observable affect here? I know TPM2_Clear does multiple things, would it be possible to observe one of them?

chrisfenner: You could generate a primary key in the storage hierarchy with the same template before and after the clear, and assert the public part changed.

josephlr: Good idea, I like that

chrisfenner:
Added this with the requested test in #305

PR #295, opened 2022-07-21
Implemented and Tested pcrReset.

matt-tsai:
Implemented and tested pcrReset. TPM2_PCR_SetAuthPolicy will need to be implemented for more coverage.

### DIFF
--- a/direct/tpm2/pcr_test.go
+++ b/direct/tpm2/pcr_test.go
@@ -1,15 +1,35 @@
 package tpm2
 
 import (
+	"bytes"
+	"crypto/sha1"
+	"crypto/sha256"
+	"crypto/sha512"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-tpm/direct/structures/tpm"
 	"github.com/google/go-tpm/direct/structures/tpml"
-	"github.com/google/go-tpm/direct/transport/simulator"
-
 	"github.com/google/go-tpm/direct/structures/tpms"
+	"github.com/google/go-tpm/direct/structures/tpmt"
+	"github.com/google/go-tpm/direct/transport/simulator"
 )
+
+var extendsDirect = map[tpm.AlgID][]struct {
+	digest []byte
+}{
+	tpm.AlgSHA1: {
+		{bytes.Repeat([]byte{0x00}, sha1.Size)},
+		{bytes.Repeat([]byte{0x01}, sha1.Size)},
+		{bytes.Repeat([]byte{0x02}, sha1.Size)}},
+	tpm.AlgSHA256: {
+		{bytes.Repeat([]byte{0x00}, sha256.Size)},
+		{bytes.Repeat([]byte{0x01}, sha256.Size)},
+		{bytes.Repeat([]byte{0x02}, sha256.Size)}},
+	tpm.AlgSHA384: {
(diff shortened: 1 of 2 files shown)


direct/tpm2/pcr_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	selection := tpml.PCRSelection{
+		PCRSelections: []tpms.PCRSelection{
+			{
+				Hash:      tpm.AlgSHA1,
+				PCRSelect: PCRs,
+			},
+		},
+	}
```
josephlr, line 33: If we don't use this struct anywhere else, could we inline it into the PCRRead struct below?

direct/tpm2/pcr_test.go
```diff
@@ -0,0 +1,72 @@
...
+		PCRHandle: authHandle,
+	}
+
+	if _, err := pcrReset.Execute(thetpm); err != nil {
+		t.Fatalf("pcrReset failed: %v", err)
+	}
+
+	pcrRead = PCRRead{
+		PCRSelectionIn: selection,
+	}
```
josephlr, line 60: do we need this, can't we just reuse the pcrRead structure we filled out earlier?

josephlr (changes requested):

direct/tpm2/pcr_test.go
matt-tsai: Yes, you are correct. Resolved.

direct/tpm2/pcr_test.go
matt-tsai: Resolved.

direct/tpm2/pcr_test.go
```diff
@@ -0,0 +1,72 @@
...
+	pcrReadRsp, err = pcrRead.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("failed to read PCRs")
+	}
+
+	postResetBuffer := pcrReadRsp.PCRValues.Digests[0].Buffer[:]
+
+	if !cmp.Equal(preResetBuffer, postResetBuffer) {
+		t.Errorf("pcr after reset changed.")
+	}
```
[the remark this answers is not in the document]
matt-tsai: I have rewritten the whole test so it is more clear. I have also sparingly added some comments for clarity.

alexmwu (approved):

direct/tpm2/pcr_test.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+			if _, err := pcrReset.Execute(thetpm); err != nil {
+				t.Fatalf("pcrReset failed: %v", err)
+			}
+			if pcrReadRsp, err = pcrRead.Execute(thetpm); err != nil {
+				t.Fatalf("failed to read PCRs")
+			}
+			postResetPCR16 := pcrReadRsp.PCRValues.Digests[0].Buffer
+
+			if !bytes.Equal(startPCR16, postResetPCR16) {
```
josephlr, line 120: Shouldn't we just be comparing the PCR to all zeros?

josephlr (approved):

direct/tpm2/pcr_test.go
matt-tsai: Yes I agree that would be more clear. I have made the change.

josephlr (approved):

Issue #287, opened 2022-06-24
Implement TPM2_Commit


akakou:
It needs when implementing ECDAA(e.g. https://github.com/google/go-tpm/issues/278#issuecomment-1148113966).

>[tpm2_commit](https://manpages.ubuntu.com/manpages/jammy/man1/tpm2_commit.1.html)(1)  -  Performs the first part of an ECC anonymous signing operation.  The TPM will perform the point multiplications on the  provided  points  and  return  intermediate signing  values.  The signing key is an ECC key.  The key cannot be a sign+decrypt key and must have an anonymous signing scheme.  TPM_ALG_ECDAA  is  the  only  supported  anonymous scheme.

https://manpages.ubuntu.com/manpages/jammy/man1/tpm2_commit.1.html

akakou:
Now I'm implementing the test code of TPM2_Commit, but not work.
Do you have any idea to solve this error?

```
go test -run TestCommit
--- FAIL: TestCommit (0.01s)
    commit_test.go:102: could not commit: TPM_RC_AUTH_MISSING: command requires an authorization session for handle and it is not present.
FAIL
exit status 1
FAIL    github.com/google/go-tpm/direct/tpm2    0.012s
```

https://github.com/akakou/go-tpm/blob/tpm2-commit/direct/tpm2/commit_test.go

akakou:
Thank you. It's one of the bugs!
But it has never worked still...and panic with the error following:

```
--- FAIL: TestCommit (0.00s)
    commit_test.go:102: could not commit: TPM_RC_SIZE (handle 0): structure is the wrong size
FAIL
exit status 1
FAIL    github.com/google/go-tpm/direct/tpm2    0.008s
```

Do you have any idea?

akakou:
Thank you! You are right and I fixed them.

Specifically, I fixed them as:
1. Set `x=0x1`, `y = 0x2` to P1.
    - It's the generator of ECP.
    - 5.2.7 BN P256 (https://trustedcomputinggroup.org/wp-content/uploads/TCG_Algorithm_Registry_Rev_1.22.pdf)
2. Use annotation to express the size of TPM2B_ECC_POINT

akakou:
But the program calls the same error...

akakou:
I compared worked TPM2_Commit(tpm2-tools) and this library.
Mainly, I compared what binaries each library sent to TPM.


These are the binaries:

go-tpm
[...]

PR #296, opened 2022-07-26
Implement TPM2_Commit

akakou:
#287

### DIFF
--- a/direct/structures/tpm2b/tpm2b.go
+++ b/direct/structures/tpm2b/tpm2b.go
@@ -110,6 +110,10 @@ type Private = internal.TPM2BPrivate
 // See definition in Part 2: Structures, section 13.6.
 type NVPublic = internal.TPM2BNVPublic
 
+// ContextSensitive represents a TPM2B_CONTEXT_SENSITIVE.
+// See definition in Part 2: Structures, section 14.2.
+type ContextSensitive = internal.TPM2BContextSensitive
+
 // CreationData represents a TPM2B_CREATION_DATA.
 // See definition in Part 2: Structures, section 15.2.
 type CreationData = internal.TPM2BCreationData
--- a/direct/structures/tpm2b/tpm2b.go
+++ b/direct/structures/tpm2b/tpm2b.go
@@ -29,6 +29,10 @@ type Timeout = internal.TPM2BTimeout
 // See definition in Part 2: Structures, section 10.4.5.
 type Auth = internal.TPM2BAuth
 
+// TPM2BMaxBuffer represents a TPM2B_MAX_BUFFER.
+// See definition in Part 2: Structures, section 10.4.8.
+type MaxBuffer = internal.TPM2BMaxBuffer
+
 // TPM2BMaxNVBuffer represents a TPM2B_MAX_NV_BUFFER.
 // See definition in Part 2: Structures, section 10.4.9.
 type MaxNVBuffer = internal.TPM2BMaxNVBuffer
--- a/direct/structures/tpm2b/tpm2b.go
+++ b/direct/structures/tpm2b/tpm2b.go
@@ -82,7 +82,7 @@ type PrivateKeyRSA = internal.TPM2BPrivateKeyRSA
 // See definition in Part 2: Structures, section 11.2.5.1.
 type ECCParameter = internal.TPM2BECCParameter
 
-// Private represents a TPM2B_ECC_POINT.
+// ECCPoint represents a TPM2B_ECC_POINT.
 // See definition in Part 2: Structures, section 11.2.5.3
 type ECCPoint = internal.TPM2BECCPoint
 
--- a/direct/structures/tpm2b/tpm2b.go
+++ b/direct/structures/tpm2b/tpm2b.go
@@ -82,6 +82,10 @@ type PrivateKeyRSA = internal.TPM2BPrivateKeyRSA
 // See definition in Part 2: Structures, section 11.2.5.1.
 type ECCParameter = internal.TPM2BECCParameter
 
+// Private represents a TPM2B_ECC_POINT.
+// See definition in Part 2: Structures, section 11.2.5.3
+type ECCPoint = internal.TPM2BECCPoint
+
 // EncryptedSecret represents a TPM2B_ENCRYPTED_SECRET.
 // See definition in Part 2: Structures, section 11.4.33.
 type EncryptedSecret = internal.TPM2BEncryptedSecret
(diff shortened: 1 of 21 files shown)


direct/tpm2/commit_test.go
```diff
@@ -0,0 +1,157 @@
...
+						KDF: tpmt.KDFScheme{
+							Scheme: tpm.AlgNull,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	rspCP, err := primary.Execute(thetpm)
```
matt-tsai, line 68: Perhaps good to FlushContext on the CP as well.

direct/tpm2/tpm2.go
```diff
@@ -654,6 +654,40 @@ type GetSessionAuditDigestResponse struct {
 // Response implements the Response interface.
 func (*GetSessionAuditDigestResponse) Response() tpm.CC { return tpm.CCGetSessionAuditDigest }
 
+// Commit is the input to TPM2_Commit.
+// See definition in Part 3, Commands, section 19.2.
+type Commit struct {
+	SignHandle handle `gotpm:"handle,auth"`
+	P1         tpm2b.ECCPoint
```
matt-tsai, line 661: nit: Add description that is included in the spec.

direct/tpm2/tpm2.go
```diff
@@ -654,6 +654,40 @@ type GetSessionAuditDigestResponse struct {
...
+	return &rsp, nil
+}
+
+// CommitResponse is the response from TPM2_Commit.
+type CommitResponse struct {
+	// the data read
+	K     tpm2b.ECCPoint
+	L     tpm2b.ECCPoint
+	E     tpm2b.ECCPoint
+	Count uint16
```
matt-tsai, line 685: nit: name in spec is Counter.

direct/structures/tpm2b/tpm2b.go
```diff
@@ -82,6 +82,10 @@ type PrivateKeyRSA = internal.TPM2BPrivateKeyRSA
 // See definition in Part 2: Structures, section 11.2.5.1.
 type ECCParameter = internal.TPM2BECCParameter
 
+// Private represents a TPM2B_ECC_POINT.
```
matt-tsai, line 85: Fix comment to be ECCPoint instead of Private.

akakou:
Thank you for your review! 

But I found the bug around parsing the response. 
So first I will fix it and then fix some code that is stated in the comments of the review.

akakou:
It was a misunderstanding, the TPM2_Commit on `go-tpm` has already worked.

> But I found the bug around parsing the response.

akakou:
@matt-tsai 

I fixed some points which you commented on.
Could you re-review and merge it?

direct/tpm2/commit_test.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+		S2: tpm2b.SensitiveData{
+			Buffer: []byte{},
+		},
+		Y2: tpm2b.ECCParameter{
+			Buffer: []byte{},
+		},
+	}
+
+	_, err = commit.Execute(thetpm)
```
matt-tsai, line 150: Testing the output response with some off-tpm validation would be good here.

akakou: Umm...Any good idea to test this output...?

I have just an idea that implements full ECDAA Setup/Join for this, but it's complex and we should avoid it.

matt-tsai: I think @chrisfenner is a better person to discuss this with, however he is out at the moment and will be back next week. In the meantime, I will add to be a reviewer.

direct/tpm2/commit_test.go
```diff
@@ -0,0 +1,160 @@
...
+					ECCDetail: &tpms.ECCParms{
+						Symmetric: tpmt.SymDefObject{
+							Algorithm: tpm.AlgNull,
+						},
+						Scheme: tpmt.ECCScheme{
+							Scheme: tpm.AlgECDAA,
+							Details: tpmu.AsymScheme{
+								ECDAA: &tpms.SigSchemeECDAA{
+									HashAlg: tpm.AlgSHA1,
+									Count:   1,
```
chrisfenner, line 107: I think this count value is ignored when specified on the key. It needs to match whatever Commit() produced, during the sign operation. Can you leave it out, for understandability?

direct/tpm2/commit_test.go
```diff
@@ -0,0 +1,160 @@
...
+				Parameters: tpmu.PublicParms{
+					ECCDetail: &tpms.ECCParms{
+						Symmetric: tpmt.SymDefObject{
+							Algorithm: tpm.AlgNull,
+						},
+						Scheme: tpmt.ECCScheme{
+							Scheme: tpm.AlgECDAA,
+							Details: tpmu.AsymScheme{
+								ECDAA: &tpms.SigSchemeECDAA{
+									HashAlg: tpm.AlgSHA1,
```
chrisfenner, line 106: nit: since test code is example code, can you use SHA256 or better for this test, to decrease the chance that somebody copies this code and uses SHA1 for production? :)

chrisfenner: (Same goes for the NameAlg)

direct/tpm2/commit_test.go
chrisfenner: I'm new to ECDAA signing, but I think you should be able to use TPM2_Sign (which is already implemented) to produce an ECDAA signature using the commit value. You'll have to grab the commit counter from the Commit command and use it on the ECDAA scheme structure in the Sign command.

Bonus points: Call Sign again with the same counter value and expect it to fail :)

No need to validate the signature at this point, unless you really want to. I imagine that belongs as part of a larger body of ECDAA sample code that someone might work on (maybe you or someone else) as part of a separate effort.

direct/tpm2/commit_test.go
```diff
@@ -0,0 +1,160 @@
...
+			Buffer: []byte{},
+		},
+	}
+
+	_, err = commit.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("could not commit: %v", err)
+	}
+
+	flushContextCP := FlushContext{FlushHandle: rspCP.ObjectHandle}
```
chrisfenner, line 155: These deferred flushes should be defined and `defer`ed right after the resources that need to be freed are created (i.e., right after CreatePrimary). However, I think you should be able to delete the parent key from this test entirely and have just one flush for the signing key.

chrisfenner (changes requested):
Thanks for your patience, I was on vacation and then catching up on work after I returned from vacation :)

The functionality looks great, thanks for the change! I have a few notes on the test to make it tighter and leaner but also cover more of an end-to-end scenario.

akakou:
I will update it as your review told me, on next week... (I'm too busy until next Saturday...)

akakou:
@chrisfenner 
Thank you for your review!
I fixed some code you pointed out.

So could you re-review and merge it if it is ok?

chrisfenner (approved):

chrisfenner:
Thank you for the change!

Issue #309, opened 2022-09-08
Add a Compare function


chrisfenner:
Since a lot of the tpmdirect structs have private members, we should create a convenient way to compare two values without hassling with adding enough types into `cmpopts.IgnoreUnexportedFields`.

Issue #292, opened 2022-07-12
Rough edges of tpmdirect to smooth out


chrisfenner:
- [ ] [Marshal](https://github.com/google/go-tpm/blob/50ce9737c31f0789c301c91ff0d246d6f18dd5a6/direct/tpm2/reflect.go#L127) takes interface{} so will panic if you accidentally pass something that is not marshallable.

  - Possible improvements: 

    - Define an interface for types that can be `Marshalled` in a compact way that minimizes code bloat.

- [ ] Type annotations are a bit magic.

  - Someone working on this code might get the `gotpm:` type annotations incorrect or slightly incorrect. It would be interesting to look into ways to make it more obvious what to do when introducing new TPM structures from the spec.

chrisfenner:
Let's add a Go1.18 [type constraint](https://go.dev/blog/intro-generics) for this:

There's tons of types, so we can divide them by category:

```go
type TPMA interface {
  TPMAAlgorithm | TPMAObject | TPMASession | TPMALocality | ...
}
```

```go
type Marshallable  interface {
  TPMA | TPMT | TPMS | ...
}
```

While we're taking a dependency on Go1.18, let's also consider using type constraints for the unions:


```go
// TPMUCapabilities represents a TPMU_CAPABILITIES.
// See definition in Part 2: Structures, section 10.10.1.
type TPMUCapabilities struct {
	Algorithms    *TPMLAlgProperty       `gotpm:"selector=0x00000000"` // TPM_CAP_ALGS
	Handles       *TPMLHandle            `gotpm:"selector=0x00000001"` // TPM_CAP_HANDLES
	Command       *TPMLCCA               `gotpm:"selector=0x00000002"` // TPM_CAP_COMMANDS
	PPCommands    *TPMLCC                `gotpm:"selector=0x00000003"` // TPM_CAP_PP_COMMANDS
	AuditCommands *TPMLCC                `gotpm:"selector=0x00000004"` // TPM_CAP_AUDIT_COMMANDS
	AssignedPCR   *TPMLPCRSelection      `gotpm:"selector=0x00000005"` // TPM_CAP_PCRS
	TPMProperties *TPMLTaggedTPMProperty `gotpm:"selector=0x00000006"` // TPM_CAP_TPM_PROPERTIES
	PCRProperties *TPMLTaggedPCRProperty `gotpm:"selector=0x00000007"` // TPM_CAP_PCR_PROPERTIES
[...]

PR #310, opened 2022-09-08
Use generics to simplify the TPMDirect interface

### LINKED ISSUE #307: support passing []byte as TPM2B
Capturing one more bit of feedback @matt-tsai gave us in his internship report -

Occasionally, someone sending a TPM2B to a TPM would benefit from not having had to deserialize or serialize it (e.g., the 2B came from a protobuf and can just be forwarded to the TPM command). Every TPM2B_X should support being instantiated either with its contents or a flat buffer. Go1.18 generics can help maintain type safety and compiler-time checking here.

chrisfenner:
This is unfortunately a large change, but I think it does a lot for the ergonomics of the TPMDirect API.

This change uses the new Go 1.18 generics to solve a few problems:

- We want people to be able to provide a flat `[]byte` or actual structure when instantiating TPM2Bs
- We want to avoid people directly manipulating pointer values in the TPMUs or having their TPMUs in an invalid state
- We want a nice Marshal and Unmarshal function (and later, to be able to make a nice Compare function, see #309 )

Generics to the rescue. Here's what this PR does:

- Add a new file called `marshalling.go` that handles a lot of the high level marshalling work. `reflect.go` is still the dirty reflection guts of the library
- Embed a new type called `marshalByReflection` into all the structs that can be marshalled by reflection, as a clear hint to the reflection library
- Add a new interface called `UnmarshallableWithHint` - most of the TPMU implement this, and the old `marshalUnion` and `unmarshalUnion` functions are gone now
- Bonus: I noticed using profiling that the `tags` function was allocating several orders of magnitude more memory than the rest of the library, so I rewrote it
- Introduced a generic TPM2B helper that is aliased by the concrete TPM2B types; there are constructors for instantiating TPM2B from data or structured contents.
- TPMU is public, with private fields. Introduced constructors for these with type constraints.

Fixes #307 and #292.


chrisfenner:
Also CCing @akakou - this is another change to the API, I hope it doesn't cause you much trouble and that the changes make your work easier. See the ways the tests change in this PR to see what sorts of modifications your code will probably need

tpm2/structures.go
```diff
@@ -668,111 +689,345 @@ type TPMSACTData struct {
...
+}
+
+type capabilitiesContents interface {
+	Marshallable
+	*TPMLAlgProperty | *TPMLHandle | *TPMLCCA | *TPMLCC | *TPMLPCRSelection | *TPMLTaggedTPMProperty |
+		*TPMLTaggedPCRProperty | *TPMLECCCurve | *TPMLTaggedPolicy | *TPMLACTData
+}
+
+// create implements the UnmarshallableWithHint interface.
+func (u *tpmuCapabilities) create(hint int64) (reflect.Value, error) {
```
chrisfenner, line 794: I'm not 100% happy with this pattern for TPMUs yet. There are few enough TPMU structures that I'm not overly worried about this, but it would be nice to have a helper type or function to make create and get more brief.

Part of the problem is that `reflect.ValueOf` needs to be called on a variable whose type is the actual one we want, and not the interface (e.g., `Marshallable`) so something somewhere needs to instantiate contents on a per-type basis.

It might be possible to create a helper that holds onto a map of selector values to prototype contents and looks it up in the map, but then we have to worry about when that map is allocated (since a zero-valued TPMU needs to be able to return the zero-valued contents of the requested union member without initialization)

tpm2/structures.go
```diff
@@ -929,66 +1372,237 @@ type TPMSAuthCommand struct {
...
+}
+
 // TPMUSymKeyBits represents a TPMU_SYM_KEY_BITS.
 // See definition in Part 2: Structures, section 11.1.3.
-type TPMUSymKeyBits struct {
-	// TODO: The rest of the symmetric algorithms get their own entry
-	// in this union.
-	AES *TPMKeyBits  `gotpm:"selector=0x0006"` // TPM_ALG_AES
-	XOR *TPMIAlgHash `gotpm:"selector=0x000A"` // TPM_ALG_XOR
+func TPMUSymKeyBits[C symKeyBitsContents](selector TPMAlgID, contents C) tpmuSymKeyBits {
```
chrisfenner, line 1431: note that the selector value is provided explicitly in this constructor, this is a little weird.

Ideally, a caller would be able to say "here's something that a TPMUFoo can hold, put it into a TPMUFoo for me" except that the TPMUFoo needs to somehow associate those contents with a selector value for future queries. It can't guess based on the concrete type of what was passed in (because you can't use generics to build function overloading, and because some TPMUs have multiple members with the same type).

tpm2/marshalling.go
```diff
@@ -0,0 +1,237 @@
...
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"reflect"
+)
+
+// maybe represents some type that came out of an unmarshalling process.
```
chrisfenner, line 11: To consider: 2B and U can return pointers

tpm2/marshalling.go
```diff
@@ -0,0 +1,237 @@
...
+	// marshal will deserialize the given value from the given buffer.
+	// Returns an error if there was an unmarshalling error or if there was not
+	// enough data in the buffer.
+	unmarshal(buf *bytes.Buffer) error
+}
+
+// UnmarshallableWithHint represents any TPM type that can be marshalled or unmarshalled,
+// but that requires a selector ("hint") value when unmarshalling. Most TPMU_ are
+// an example of this.
+type UnmarshallableWithHint interface {
```
chrisfenner, line 73: this interface might be hideable

tpm2/marshalling.go
```diff
@@ -0,0 +1,237 @@
...
+	if err := marshal(&buf, reflect.ValueOf(v)); err != nil {
+		panic(fmt.Sprintf("unexpected error marshalling %v: %v", reflect.TypeOf(v).Name(), err))
+	}
+	return buf.Bytes()
+}
+
+// Unmarshal unmarshals the given type from the byte array.
+// Returns an error if the buffer does not contain enough data to satisfy the
+// types, or if the types are not unmarshallable.
+func Unmarshal[T any, P interface {
```
chrisfenner, line 93: Try the obvious thing which Unmarshal[*T Unmarshallable]

tpm2/marshalling.go
```diff
@@ -0,0 +1,237 @@
...
+type UnmarshallableWithHint interface {
+	// create will instantiate and return the corresponding union member.
+	create(hint int64) (reflect.Value, error)
+	// get will return the corresponding union member by copy. If the union is
+	// uninitialized, it will initialize a new zero-valued one.
+	get(hint int64) (reflect.Value, error)
+}
+
+// Marshal will serialize the given values, returning them as a byte slice.
+func Marshal[T Marshallable](v T) []byte {
```
chrisfenner, line 82: This probably doesn't have to be generic, test it out

tpm2/marshalling.go
chrisfenner: Or clean this up with a helper constraint

tpm2/marshalling.go
```diff
@@ -0,0 +1,237 @@
...
+// Marshallable represents any TPM type that can be marshalled.
+type Marshallable interface {
+	// marshal will serialize the given value, appending onto the given buffer.
+	// Returns an error if the value is not marshallable.
+	marshal(buf *bytes.Buffer)
+}
+
+// Unmarshallable represents any TPM type that can be marshalled or unmarshalled.
+type Unmarshallable interface {
+	Marshallable
```
chrisfenner, line 63: Might be good to make it more clear the relationship between Marshallable and Unmarshallable

tpm2/marshalling.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+	}
+	return maybe[T]{
+		value: value.Interface().(*T),
+	}
+}
+
+// marshallableByReflection is a placeholder interface, to hint to the unmarshalling
+// library that it is supposed to use reflection.
+type marshallableByReflection interface {
```
chrisfenner, line 113: Try making this more polymorphic with Marshallable

chrisfenner: Oh, the reason this didn't work out was that the marshalByReflection ends up as the receiver type, and you can't use reflect to deal with the other stuff

tpm2/marshalling.go
chrisfenner, line 93: To consider: weigh the complexity of this solution vs the benefits to type safety here

tpm2/test/audit_test.go
```diff
@@ -25,38 +25,39 @@ func TestAuditSession(t *testing.T) {
...
+				UserWithAuth:         true,
+				AdminWithPolicy:      false,
+				NoDA:                 true,
+				EncryptedDuplication: false,
+				Restricted:           true,
+				Decrypt:              false,
+				SignEncrypt:          true,
+			},
+			Parameters: TPMUPublicParms(
+				TPMAlgECC,
```
chrisfenner, line 45: Little bit irritating to use here, can we avoid the user having to pass the selector

chrisfenner: Consider making the underlying tpmu contain an interface that happens to be implemented (in dummy methods)

tpm2/marshalling.go
chrisfenner: If Marshal/Unmarshal is permitted to return an error we can simplify this quite a bit

chrisfenner: We can consider leaving it generic for future improvements to the constraints

chrisfenner: Could even have it panic vs return an error in that case

chrisfenner: Or bite the bullet and write the placeholder method for all 100 something types

tpm2/marshalling.go
chrisfenner: hidden

tpm2/structures.go
chrisfenner: Haven't thought of a better pattern, and since there are only a few TPMUs, the cost of changing this later is pretty low.

tpm2/marshalling.go
chrisfenner: Can't return nil if we do this.

tpm2/marshalling.go
chrisfenner: Fixed.

tpm2/test/audit_test.go
chrisfenner: Unfortunately, the selector cannot always be inferred from the type of the contents, because some TPU contain multiple different members that have the same type.

chrisfenner:
@alexmwu @josephlr Hi all, sorry this sat for so long. I finally made some time to work on the feedback from our verbal review. I've deleted generics from where they weren't absolutely necessary (e.g., the old `maybe` type is gone, and `Marshal` now has no type constraints). Would you mind taking a look?

tpm2/marshalling.go
```diff
@@ -0,0 +1,191 @@
...
+	"fmt"
+	"io"
+	"reflect"
+)
+
+// Marshallable represents any TPM type that can be marshalled.
+type Marshallable interface {
+	// marshal will serialize the given value, appending onto the given buffer.
+	// Returns an error if the value is not marshallable.
+	marshal(buf *bytes.Buffer)
```
josephlr, line 15: This doesn't seem to return an error.

josephlr (commented):
A few questions:

- Why do we still have the `Startup` and `Shutdown` functions, can't we just use the direct API to call these functions?
- Why does the `Response` need a `TPMCC` as part of its interface?
- Having the Command/Response structs contain private types makes it hard to figure out _how_ to construct our use these types. 
  - Constructing is possible to figure out by searching the docs
  - However, use of the private structs (say to read fields of a union) doesn't seem to show up in godoc.
  - Would it be possible to have the types be public (`TPM2BAttest`) and have public constructors for these types (`NewTPM2BAttest`)

tpm2/marshalling.go
chrisfenner: Good catch. I'll fix this.

To your other questions:

> Why do we still have the Startup and Shutdown functions, can't we just use the direct API to call these functions?

As an adapter so that we can use https://github.com/google/go-tpm-tools/tree/master/simulator (which will call back to Startup and Shutdown as free functions in go-tpm) until there's a version of go-tpm-tools that calls TPMDirect startup and shutdown. I'll add a comment and TODO about this.

> Why does the Response need a TPMCC as part of its interface?

This allows the internal consistency check in the `execute` function to match the correct command and response together. I guess this could be done with generics instead now, WDYT?

> Having the Command/Response structs contain private types makes it hard to figure out how to construct our use these types.

Yeah, this is a great point. I'll make e.g., the `TPM2B` types themselves public and provide public constructors.

josephlr: I was playing around a little and I think the following would work well:

Have a `tpm2b` submodule:
```go
// Package tpm2b ...
//
// Document here how the TPM2B types work and give examples of how to use them
package tpm2b

type TPM2B[T any] struct {
	contents *T
	buffer   []byte
}

func New[T any](t T) TPM2B[T] {
	return TPM2B[T]{contents: &t}
}

func FromBytes[T any](b []byte) TPM2B[T] {
	return TPM2B[T]{buffer: b}
}

func (t *TPM2B[T]) Contents() (*T, error) {
	if value.contents != nil {
		return value.contents, nil
	}
	if value.buffer == nil {
		panic("TPMB had no contents or buffer")
	}
	var result T
	if err := unmarshal(bytes.NewBuffer(value.buffer), reflect.ValueOf(&result).Elem()); err != nil {
		return nil, err
	}
	
	// Cache result
	value.contents = &result
	return value.contents, nil
}

func (t *TPM2B[T]) Bytes() []byte {
	if value.buffer != nil {
		return value.buffer
	}
	if value.contents == nil {
		panic("TPMB had no contents or buffer")
	}
	var temp bytes.Buffer
	marshal(&temp, reflect.ValueOf(value.contents))
	
	// Cache Bytes
	value.buffer = temp.Bytes()
	return value.buffer
}

func (value *tpm2b[T]) marshal(buf *bytes.Buffer) {
	b := value.Bytes()
	binary.Write(buf, binary.BigEndian, uint16(len(b)))
	buf.Write(b)
}

func (value *tpm2b[T]) unmarshal(buf *bytes.Buffer) error {
	var size uint16
	binary.Read(buf, binary.BigEndian, &size)
	value.contents = nil
	value.buffer = make([]byte, size)
	_, err := io.ReadAtLeast(buf, value.buffer)
	return err
}
```

Then in the main `tpm2` module will have just the typedefs:
```go
type TPM2BECCPoint = tpm2b.TPM2B[TPMSECCPoint]
```

> Good catch. I'll fix this.

I think _not_ having `Marshall`/`marshal` return an error makes the library easier to use.

josephlr:
For the commands and responses, is there a reason that the following doesn't work:

We have a generic interface for commands:
```go
type Command[Response any] interface {
	CommandCode() TPMCC
	Execute(t transport.TPM, s ...Session) (Response, error)
}
```
and we have value recievers for the Commands and return the responses by value:
```go
type CreateLoaded struct { /* ... */ }
type CreateLoadedResponse struct { /* ... */ }

func (CreateLoaded) CommandCode() TPMCC { return TPMCCCreateLoaded }
func (cmd CreateLoaded) Execute(t transport.TPM, s ...Session) (CreateLoadedResponse, error) {
	var rsp CreateLoadedResponse
	return rsp, execute(t, cmd, &rsp, s...)
}
```

We have internal `execute` and `executeImpl` functions similar to what is already in this PR:
```go
// Typechecks that we implemented the Command interface correctly
func execute[C Command[R], R any](t transport.TPM, cmd C, rsp *R, s ...Session) error {
	return executeImpl(t, cmd.CommandCode(), cmd, rsp, s...)
}

func executeImpl(t transport.TPM, cc TPMCC, cmd any, rsp any, s ...Session) error {
	// Actual implementation of command execution
}
```

josephlr:
Another thing, for the `TPML` types, is there a reason we can't just do:
```go
type TPMLPCRSelection []TPMSPCRSelection
```
and have the reflection code figure out that its a list? Or if that doesn't work, just tag it as a list in the parent structure:
```go
type TPMSQuoteInfo struct {
	marshalByReflection
	// information on algID, PCR selected and digest
	PCRSelect TPMLPCRSelection `gotpm:"list"`
	// digest of the selected PCR using the hash of the signing key
	PCRDigest TPM2BDigest
}
```

It seems like we could just determine what to do based on the type alone though (i.e. we shouldn't need the tags). As if the type to marshalled/unmarshalled is an array, either:
  - it's a `[]byte` (or an alias thereof), and it gets a 16-bit size prefix
  - it's some other `[]T`, and it gets a 32-bit size prefix

josephlr:
The above TPML change does cause one issue, it makes things like:
```go
type capabilitiesContents interface {
	Marshallable
	*TPMLAlgProperty | *TPMLHandle | *TPMLCCA | *TPMLCC | *TPMLPCRSelection | *TPMLTaggedTPMProperty |
		*TPMLTaggedPCRProperty | *TPMLECCCurve | *TPMLTaggedPolicy | *TPMLACTData
}
```
no longer work as something like `*TPMLCC` (which would be a `*[]uint32`) cannot implement the Marshallable interface as it is a builtin type.

The solution seems to be to remove the `Marshallable` constraint from `capabilitiesContents`. This means that the reflection routine that marshals/unmarshals `tpmuCapabilities` can't just call the `Marshal` method, but this seems fine.

josephlr:
One final ergonomics issue around the `TPMU` types.

Similar to the TPM2B types, the actual `tpmu` type that's in the relevant structures is not exported, so it's accessor methods are not visible in documentation. We might want to make the type public.

For exampe, if we have `TPMU_FOO` :
```go
type TPMUFoo struct {
	selector TPMAlgID
	contents any
}

type FooContents interface {
	TPMSBar | TPMSBaz
}

func NewTPMUFoo[C FooContents](contents C) TPMUFoo {
	sel := /* ideally computed from C, if not possible, pass selector type in constructor */
	return TPMUFoo{
		selector: sel,
		contents: &contents,
	}
}

// Accessor methods (nil if unoccupied):
func (u *TPMUFoo) Bar() *TPMSBar {
	// ...
}
func (u *TPMUFoo) Baz() *TPMSBaz {
	// ...
}
```

chrisfenner:
> The reason I would prefer the Commands/Responses being passed by value is that them being pointer inputs/outputs was a little surprising when I was trying to write some code. For example, it would be nice if the following worked:

Great suggestion! This required making `Marshallable` and `marshallableWithHint` take receiver by value, and `Unmarshallable` and `unmarshallableWithHint` by pointer. I thought this would not work as I wasn't aware Go's pointer decay would support this (since the Unmarshallable interfaces include the Marshallable ones) but it turns out it works just fine. This simplifies the Command interface a bit, and even a little of the reflection code. I did end up having to meet you halfway, though. Commands' `Execute` will return the Response by pointer, i.e., `(*FooResponse, error)` so that we can return a null instead of garbage value if the command fails. However

```go
rsp, err := (tpm2.GetCapability{Capability: tpm2.TPMCapAlgs, PropertyCount: 0xFFFF}).Execute(s)
```

now works (note you have to put parens around the `GetCapability` constructor, the compiler didn't like it without them).

chrisfenner:
I'm in the middle of incorporating these excellent ideas. ❤️ = done, 👍 = "I think I can do this" and 👀 = "I have to try it but I like it." I'll ping when ready for another round of review!

josephlr:
> Great suggestion! This required making `Marshallable` and `marshallableWithHint` take receiver by value, and `Unmarshallable` and `unmarshallableWithHint` by pointe

I'm confused why taking/returning the Command and Response structures by pointer or value affects the receiver types for `Marshallable` and `Unmarshallable`. The commands and responses don't seem to implement those interfaces, so it seems like they won't necessarily have to be related.

> i.e., `(*FooResponse, error)` so that we can return a null instead of garbage value if the command fail

It initially seemed to me that return the zero value for `FooResponse` would be fine. Is the issue just one of safety, were we want the user's program to crash if they forget to check the error? Or is the issue about if we partially Unmarshal the response into `FooResponse`, leading to an object that is partially valid and partially invalid?

chrisfenner:
> I'm confused why taking/returning the Command and Response structures by pointer or value affects the receiver types

The subtle problem: `Command` structs contain `Marshallable` structs inside them. Go reflection doesn't let you take the address of a member field of a struct that you got by value. So if you have to take the address of the member to use `Marshal` on it, you have to pass the whole containing struct by pointer as well.

> It initially seemed to me that return the zero value for FooResponse would be fine.

I think this would harm users more than it helped them: maybe the user was expecting the zero value for the response. I think we should require the user to check the error (and crash obviously if they didn't).

chrisfenner:
> Another thing, for the TPML types, is there a reason we can't just do:

Unfortunately this generally means the TPML types can't be marshalled by users. I'd like users to be able to marshal any TPM type. I think the drawback to keeping it this way is just a bit more boilerplate in the structures code for TPML, so keeping it for now. WDYT?

chrisfenner:
@josephlr Thanks for your patience and for the great feedback. I've incorporated your comments except for the idea about making `TPMLSomething` a `[]Something`, because I'd like for all types to be usable with `Marshal` directly.

BTW, the solution to TPM2B differs from yours in a couple key ways:

- Can't be a submodule due to bidirectional dependency with tpm2
- Name changes since it unfortunately can't be submoduled
- The type parameters on TPM2B are a little complicated because it needs to capture the concrete type of both T (Marshallable) and *T (Unmarshallable) but thankfully this weirdness is completely hidden from users due to type inference.

chrisfenner:
I'd like to mirror this into Google's internal repo as part of next steps for tpmdirect. Since this is a separate branch, it seems OK to me to go ahead and merge. Thanks for the great feedback so far!

Let's discuss any other things that need to be fixed as part of the path to 1.0 release.

josephlr:
> I'd like to mirror this into Google's internal repo as part of next steps for tpmdirect. Since this is a separate branch, it seems OK to me to go ahead and merge. Thanks for the great feedback so far!
> 
> Let's discuss any other things that need to be fixed as part of the path to 1.0 release.

That seems reasonable further changes (if any) will likely be small, and can have their own prs. I think the main test now is to actually use this library to implement some of the medium complexity TPM functionality, and see what works and what doesnt

PR #311, opened 2022-09-27
Add TPMS_ID_OBJECT structure

akakou:


tpm2/structures.go
```diff
@@ -1501,9 +1501,21 @@ type TPMSCreationData struct {
...
 }
 
+// TPMSIDObject represents a TPMS_ID_OBJECT.
+// See definition in Part 2: Structures, section 12.4.2.
+type TPMSIDObject struct {
+	// HMAC using the nameAlg of the storage key on the target TPM
+	IntegrityHMAC TPM2BDigest
+	// credential protector information returned if name matches the referenced object
+	// All of the encIdentity is encrypted, including the size field.
+	EncIdentity []byte
```
chrisfenner, line 1511: This is a strange structure, so it is one of the exceptional cases that can't just get reflection applied to it, I don't think :)

This "encrypted sized" structure makes this not possible to unmarshal (since we don't know the encrypted size of EncIdentity). That's why TPMS_ID_OBJECT in the spec says:

> This structure is used for sizing the TPM2B_ID_OBJECT.

It also says that TPM2B_ID_OBJECT contains a buffer that happens to be limited to the size of a TPMS_ID_OBJECT:

> `credential[size]{:sizeof{TPMS_ID_OBJECT}}`

So I would recommend making `TPM2BIDObject` be a boring TPM2BData, and adding a factory function that makes it easier to assemble a `TPM2BIDObject`.

chrisfenner (changes requested):

tpm2/structures.go
akakou: @chrisfenner 

I'm sorry to reply so late; I had written my graduation thesis.

I agree with your opinion, and I will implement it.

To implement it, I have a question.
Which files should I implement it...?
Do you have an idea?

PR #321, opened 2023-03-30
add github actions tests

ericchiang:
Was chatting with @brandonweeks who mentioned go-tpm uses a different CI and expressed interest in github actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,16 @@
+on: [push, pull_request]
+name: Test
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-version: [1.19.x, 1.20.x]
+        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


chrisfenner:
@josephlr @alexmwu do either of you have context on why the current CI uses Cirrus instead of GH Actions?

alexmwu:
> @josephlr @alexmwu do either of you have context on why the current CI uses Cirrus instead of GH Actions?

Joe introduced this to both go-tpm-tools and go-tpm: https://github.com/google/go-tpm-tools/pull/39, https://github.com/google/go-tpm/pull/126.

go-tpm-tools has since migrated to GH Actions (https://github.com/google/go-tpm-tools/pull/135), so I don't see any special reason to use Cirrus CI. I don't have additional context and would just be guessing :).

brandonweeks (approved):

PR #332, opened 2023-06-21
rely on go-tpm-tools @ 4639ecce2abad383ae6c5cbbc0eba5ba37abb05a

chrisfenner:
This change updates go-tpm to use https://github.com/google/go-tpm-tools/commit/4639ecce2abad383ae6c5cbbc0eba5ba37abb05a whose TPM simulator helper now uses the `Startup` command via its legacy path, so the tests should all compile and pass now.


chrisfenner:
Discussed this change with @alexmwu offline, merging now to avoid another 1-day turnaround due to time zone differences

PR #345, opened 2023-09-13
Introduce helper for formatting PCR selection bitmasks

chrisfenner:
This change introduces PCClientCompatible.PCRs(), a function that converts a variadic list of PCR indices (as ints) into a PCR selection bitmask. Because of the vagaries of TPM:

1. That the minimum size of a PCR selection bitmask is not 0, but related to the minimum number of PCRs specified by the profile
2. That the PC Client Platform TPM Profile specification mandates a minimum but not a maximum number of implementation PCR,

this change creates an interface that could be implemented for other TPM profiles that specify different amounts of PCRs. The vast majority of on-market TPMs will just work with PCClientCompatible.PCRs, even if they implement more than 24 PCRs.

PCRs() can panic if given invalid values; this is to allow it to be inlined into the definition of a structure that needs a PCR selection.

### DIFF
--- a/tpm2/pcrs.go
+++ b/tpm2/pcrs.go
@@ -1,0 +1,61 @@
+package tpm2
+
+import "fmt"
+
+// PCRSelectionFormatter is a Platform TPM Profile-specific interface for
+// formatting TPM PCR selections.
+type PCRSelectionFormatter interface {
+	// PCRs returns the TPM PCR selection bitmask associated with the given PCR indices.
+	// May panic if passed invalid PCR indices (e.g., negative values).
+	PCRs(pcrs ...int) []byte
+}
+
+// PCClientCompatible is a PCRSelectionFormatter that formats PCR selections
+// suitable for use in PC Client PTP-compatible TPMs (the vast majority):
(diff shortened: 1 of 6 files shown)


tpm2/pcrs.go
```diff
@@ -0,0 +1,63 @@
...
+	if selectionSize < (pcClientMinimumPCRCount / 8) {
+		selectionSize = (pcClientMinimumPCRCount / 8)
+	}
+
+	// Allocate a byte array to store the bitfield, that has at least
+	// enough bits to store our selections.
+	selection := make([]byte, selectionSize)
+	for _, pcr := range pcrs {
+		// Panic if negative PCRs are selected.
+		if pcr < 0 {
```
alexmwu, line 47: Could `pcr` be a uint?

alexmwu (approved):
Great quality-of-life improvement!

tpm2/pcrs.go
chrisfenner: Done. I had initially used `int` since functions like `len` return `int` but I think uint is way better (no panics now)!

PR #351, opened 2024-02-16
Add TPM2_HMAC_Start implementation

nckrss:
- See definition in Part 3, Commands, section 17.2

### DIFF
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -639,11 +639,11 @@ const (
 // Context Handle Values come from table 211
 const (
 	// an ordinary transient object
-	TPMIDHSSavedTransient TPMIDHSaved = 0x80000000
+	TPMIDHSavedTransient TPMIDHSaved = 0x80000000
 	// a sequence object
-	TPMIDHSSavedSequence TPMIDHSaved = 0x80000001
+	TPMIDHSavedSequence TPMIDHSaved = 0x80000001
 	// a transient object with the stClear attribute SET
-	TPMIDHSSavedTransientClear TPMIDHSaved = 0x80000002
+	TPMIDHSavedTransientClear TPMIDHSaved = 0x80000002
 )
 
 // TPMHandle represents a TPM_HANDLE.
--- a/tpm2/constants.go
+++ b/tpm2/constants.go
@@ -634,6 +634,18 @@ const (
 	TPMHTAC            TPMHT = 0x90
 )
 
+// Saved Context transient object handles.
+// See definition in Part 2: Structures, section 14.6.2
+// Context Handle Values come from table 211
+const (
+	// an ordinary transient object
+	TPMIDHSSavedTransient TPMIDHSaved = 0x80000000
+	// a sequence object
+	TPMIDHSSavedSequence TPMIDHSaved = 0x80000001
+	// a transient object with the stClear attribute SET
+	TPMIDHSSavedTransientClear TPMIDHSaved = 0x80000002
+)
+
 // TPMHandle represents a TPM_HANDLE.
 // See definition in Part 2: Structures, section 7.1.
 type TPMHandle uint32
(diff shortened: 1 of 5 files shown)


tpm2/constants.go
```diff
@@ -634,6 +634,18 @@ const (
 	TPMHTAC            TPMHT = 0x90
 )
 
+// Saved Context transient object handles.
+// See definition in Part 2: Structures, section 14.6.2
+// Context Handle Values come from table 211
+const (
+	// an ordinary transient object
+	TPMIDHSSavedTransient TPMIDHSaved = 0x80000000
```
chrisfenner, line 642: nit: These symbols don't have a name in the spec. I like your suggestions except for the extra S: could you name them `TPMIDHSavedTransient`, `TPMIDHSavedSequence`, etc.

tpm2/structures.go
```diff
@@ -107,8 +107,8 @@ func (h TPMHandle) HandleValue() uint32 {
 // only PCR, session, and permanent values have known constant Names.
 // See definition in part 1: Architecture, section 16.
 func (h TPMHandle) KnownName() *TPM2BName {
-	switch (byte)(h >> 24) {
-	case 0x00, 0x02, 0x03, 0x40:
+	switch (TPMHT)(h >> 24) {
+	case TPMHTPCR, TPMHTHMACSession, TPMHTPolicySession, TPMHTPermanent, TPMHTTransient:
```
chrisfenner, line 111: Are you sure it's right to add `TPMHTTransient` here? A transient object's name is not just the big-endian byte array of its handle, unlike with PCR, HMAC and Policy sessions, and Permanent entities.

(Thank you for the cleanup though, putting the actual symbols here!)

chrisfenner (changes requested):
I have only some very minor requests while the Cirrus CI re-runs. Thank you for sending this change!

tpm2/constants.go
nckrss: Updated in https://github.com/google/go-tpm/pull/351/commits/f3f60777432027152671c9a5dad0934124ace348

tpm2/structures.go
nckrss: Thanks for calling this out, I mis-interpreted Part 2 Structures section 14.6.2 _savedHandle_.

Based on the table from Part Architecture section 16 Table 3,  I changed the handling in https://github.com/google/go-tpm/pull/351/commits/0d435558a7e4bfdd75116dece576f6601fd1f1df to address the case of a sequence object and to allow for Name computation of Transient and Persistent Objects in future implementations.

I first thought of changing the  `AuthHandle{}` `KnownName()` method because passing in an empty `TPM2BName.Name`<sup>1</sup> would default to `TPM2BName.Handle.KnownName()` being called, which will always be `nil` because the _Name_ associated with a _sequenceHandle_ is an empty buffer<sup>1</sup>.  The `nil` _Name_ value would then error in `cmdNames`.

Rather than modify the `AuthHandle` struct method I opted to extend the _Name_ computation since it is defined for both Transient and Persistent Objects in Part Architecture section 16 Table 3.

Concerns or suggestions?

<sup>1</sup> Part 1: Architecture, section 32.4.5

nckrss:
I updated the `Dockerfile` to use `golang:1.21` because using `golang:latest` introduces incompatible changes between the linter v1.52.2 and go 1.22.  Updating the linter to v1.56.2 (latest) raises new lint errors not in scope for this PR.  If you're open to updating dependencies, I can open a separate PR to update the dependencies and fix lint errors.

chrisfenner (approved):

chrisfenner:
Thanks @nckrss for the change!

chrisfenner:
> I updated the `Dockerfile` to use `golang:1.21` because using `golang:latest` introduces incompatible changes between the linter v1.52.2 and go 1.22. Updating the linter to v1.56.2 (latest) raises new lint errors not in scope for this PR. If you're open to updating dependencies, I can open a separate PR to update the dependencies and fix lint errors.

If you would like to do this, I would love to review the PR! Thanks in advance!

nckrss:
> If you would like to do this, I would love to review the PR! Thanks in advance!

Opened #353

PR #357, opened 2024-04-01
feat: add HierarchyChangeAuth command

novag:
see definition in Part 3, Commands, section 24.8


tpm2/test/hierarchy_change_auth_test.go
```diff
@@ -0,0 +1,75 @@
...
+			},
+		}
+
+		_, err := hca.Execute(thetpm)
+		if err != nil {
+			t.Fatalf("failed HierarchyChangeAuthWithAuth: %v", err)
+		}
+	})
+
+	t.Run("Clear", func(t *testing.T) {
```
chrisfenner, line 65: Hmm, what aspect of HierarchyChangeAuth is this subtest testing?

tpm2/test/hierarchy_change_auth_test.go
```diff
@@ -0,0 +1,75 @@
...
+		hca := HierarchyChangeAuth{
+			AuthHandle: TPMRHOwner,
+			NewAuth: TPM2BAuth{
+				Buffer: newAuthKey,
+			},
+		}
+
+		_, err := hca.Execute(thetpm)
+		if err == nil {
+			t.Fatal("failed HierarchyChangeAuthWithoutAuth: Expected HierarchyChangeAuth to fail")
```
chrisfenner, line 44: optional: you could assert a particular error (might be `TPM_RC_AUTH`)

tpm2/test/hierarchy_change_auth_test.go
```diff
@@ -0,0 +1,75 @@
...
+		hca := HierarchyChangeAuth{
+			AuthHandle: TPMRHOwner,
+			NewAuth: TPM2BAuth{
+				Buffer: authKey,
+			},
+		}
+
+		_, err := hca.Execute(thetpm)
+		if err != nil {
+			t.Fatalf("failed HierarchyChangeAuth: %v", err)
```
chrisfenner, line 30: nit: Some of these `t.Fatal` cases could be `t.Error` since they aren't must-succeed setup steps. Not a big deal to me either way

chrisfenner (changes requested):
Thanks for the change! I had a couple of minor comments, otherwise looks ready to go.

tpm2/test/hierarchy_change_auth_test.go
novag: You are right, this step is unnecessary for this test.

tpm2/test/hierarchy_change_auth_test.go
novag: Can you clarify that? In which case would it be okay if this test fails?

novag (commented):
Thanks! In which cases would you want the t.Fatal to be a t.Error?

tpm2/test/hierarchy_change_auth_test.go
chrisfenner: Sure! Basically, https://google.github.io/styleguide/go/decisions#keep-going:

> Tests should keep going for as long as possible, even after a failure, in order to print out all of the failed checks in a single run. This way, a developer who is fixing the failing test doesn’t have to re-run the test after fixing each bug to find the next bug.
>
> Prefer calling t.Error over t.Fatal for reporting a mismatch. When comparing several different properties of a function’s output, use t.Error for each of those comparisons.
>
> Calling t.Fatal is primarily useful for reporting an unexpected error condition, when subsequent comparison failures are not going to be meaningful.

The cases in this PR are somewhat borderline since the tests are nice and simple.

chrisfenner:
To be more precise, I think lines 31, 45, and 62 in `change_auth_test.go` could all be `t.Errorf`. There aren't currently more test assertions after these lines so it's not a big deal. Either `Fatalf` or `Errorf` will fail the test, but `Errorf` is preferred unless the failure means the test should abort (https://google.github.io/styleguide/go/decisions#keep-going).

chrisfenner (approved):

chrisfenner:
Thank you @novag for the change and for the updates in response to my minor feedback items!

PR #359, opened 2024-06-12
add PolicyAuthValue

salrashid123:
Adds support for `PolicyAuthVaue`


salrashid123:
@chrisfenner for review per discussion

note the test cases are end-to-end vs comparing just the policy values

chrisfenner (approved):

PR #360, opened 2024-06-25
add EncryptDecrypt2 support

salrashid123:
* added encryptdecrypt support
* test case covers chained blocks as an example

### DIFF
--- a/tpm2/test/symmetric_encryption_test.go
+++ b/tpm2/test/symmetric_encryption_test.go
@@ -25,26 +25,8 @@ func TestAESEncryption(t *testing.T) {
 		}
 	})
 
-	createPrimaryCmd := CreatePrimary{
+	primary, err := CreatePrimary{
 		PrimaryHandle: TPMRHOwner,
-		InPublic:      New2B(RSASRKTemplate),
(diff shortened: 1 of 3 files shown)


tpm2/test/symmetric_encryption_test.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		t.Fatalf("could not connect to TPM simulator: %v", err)
+	}
+	t.Cleanup(func() {
+		if err := theTpm.Close(); err != nil {
+			t.Errorf("%v", err)
+		}
+	})
+
+	createPrimaryCmd := CreatePrimary{
```
chrisfenner, line 28: Could you simplify this test by making the AES key a primary key?

chrisfenner (changes requested):
Small change requested for the test

tpm2/test/symmetric_encryption_test.go
salrashid123: done;  (just a habit to make make primary and then key)

chrisfenner (approved):

PR #96, opened 2019-04-27
Add a wrapper that connects to TPM 1.2 emulators.

tmroeder:
This wrapper follows the style of communication required by the IBM SW
TPM 1.2 over Unix domain sockets: it connects, writes a request, reads a
response, then closes the socket. Each following write/read pair do the
same thing.

PR #364, opened 2024-08-01
tpmutil: separate OpenTPM into emulation and non-emulation

leongross:
To prohibit the inclusion of the "net" package, separate the build of `tpmutil` via build tag `emu` into a filesystem-based and a socket-based version.

**EDIT**: This PR is part of enabling `tinygo` support for u-root.
`tinygo` currently lacks sufficient linux network support, so we need to remove unnecessary networking features, e.g. in [vboot](https://github.com/u-root/u-root/blob/main/cmds/exp/vboot/vboot.go#L16).

### DIFF
--- a/tpmutil/run_emu.go
+++ b/tpmutil/run_emu.go
@@ -1,0 +1,100 @@
+// Copyright (c) 2018-2024, Google LLC All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !windows && emu
+
+package tpmutil
+
+import (
+	"fmt"
+	"io"
+	"net"
+	"os"
+)
+
+// In emulation scenarios, the file is a // Unix domain socket, then it
+// opens a connection to the socket.
+func OpenTPM(path string) (io.ReadWriteCloser, error) {
+	// If it's a regular file, then open it
+	fi, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+
+	if fi.Mode()&os.ModeSocket == 0 {
+		return nil, fmt.Errorf("unsupported TPM file mode %s", fi.Mode().String())
+	}
+
+	return NewEmulatorReadWriteCloser(path), nil
+}
+
+// dialer abstracts the net.Dial call so test code can provide its own net.Conn
+// implementation.
+type dialer func(network, path string) (net.Conn, error)
+
+// EmulatorReadWriteCloser manages connections with a TPM emulator over a Unix
+// domain socket. These emulators often operate in a write/read/disconnect
+// sequence, so the Write method always connects, and the Read method always
+// closes. EmulatorReadWriteCloser is not thread safe.
+type EmulatorReadWriteCloser struct {
+	path   string
+	conn   net.Conn
+	dialer dialer
+}
+
+// NewEmulatorReadWriteCloser stores information about a Unix domain socket to
+// write to and read from.
+func NewEmulatorReadWriteCloser(path string) *EmulatorReadWriteCloser {
+	return &EmulatorReadWriteCloser{
+		path:   path,
+		dialer: net.Dial,
+	}
+}
+
+// Read implements io.Reader by reading from the Unix domain socket and closing
+// it.
+func (erw *EmulatorReadWriteCloser) Read(p []byte) (int, error) {
+	// Read is always the second operation in a Write/Read sequence.
+	if erw.conn == nil {
+		return 0, fmt.Errorf("must call Write then Read in an alternating sequence")
+	}
+	n, err := erw.conn.Read(p)
+	erw.conn.Close()
+	erw.conn = nil
+	return n, err
+}
+
+// Write implements io.Writer by connecting to the Unix domain socket and
+// writing.
+func (erw *EmulatorReadWriteCloser) Write(p []byte) (int, error) {
+	if erw.conn != nil {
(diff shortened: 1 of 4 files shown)


tpmutil/run_emu.go
```diff
@@ -0,0 +1,103 @@
+//go:build !windows && emu
```
rminnich, line 1: generally this kind of tag is not put before copyright, please see other u-root files.

tpmutil/run_emu.go
```diff
@@ -0,0 +1,103 @@
...
+// opens a connection to the socket.
+func OpenTPM(path string) (io.ReadWriteCloser, error) {
+	// If it's a regular file, then open it
+	var rwc io.ReadWriteCloser
+	fi, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+
+	if fi.Mode()&os.ModeSocket != 0 {
```
rminnich, line 36: it is more commont to do:
if fi.Mode()&os.ModeSocket == 0 {
return nil, fmt.Errorf. ...
}
rc = ...

tpmutil/run_emu.go
```diff
@@ -0,0 +1,103 @@
...
+
+// dialer abstracts the net.Dial call so test code can provide its own net.Conn
+// implementation.
+type dialer func(network, path string) (net.Conn, error)
+
+// EmulatorReadWriteCloser manages connections with a TPM emulator over a Unix
+// domain socket. These emulators often operate in a write/read/disconnect
+// sequence, so the Write method always connects, and the Read method always
+// closes. EmulatorReadWriteCloser is not thread safe.
+type EmulatorReadWriteCloser struct {
```
rminnich, line 53: what I'm not understanding: why do you need all this when you can use net.DialUnix for the test case? This seems like un-needed work?

rminnich: e.g in a file with the emu tag
var dial = net.DialUnix
and in another file with ! emu
var dial = however you dial

something like that. Can that work?

leongross: I am not sure about the design choices introducing this code via https://github.com/google/go-tpm/pull/96 but I think it should introduce abstractions for the Read and Write operations to ensure a correct communication with the TPM emulator.

leongross: @tmroeder

chrisfenner:
Hi @leongross, first my apologies for slow SLO on PR review here.

I confess, I don't really understand the purpose of this change. What aspect of go-tpm is pulling in `net` in your use case? I feel like a build tag that disables some compilation is less ideal than just factoring packages better.

leongross:
I dug up this "problem" while enabling builds of u-root commands using `tinygo`.
Tinygo does not (yet) support networking on linux systems, s.t. APIs like `net.Dial` cannot be used. While building the before mentioned `vboot` command the build failed and I investigated, that this was due to this `go-tpm` package requiring the net package.

The function `OpenTPM` checks the type of the TPM (linux device aka file, or Unix Domain Socket for emulation) at runtime. Hence there is no way to build this package depending on if the type of supported TPM interface (emulation or non-emulation) is known and should be set at build time.

I want to move this decision of the opening mode from runtime to build time.
There could even be something like only file-based, only uds-based, and as is, decide at the runtime, so every use case is covered.

An alternative to build-tags might be several `OpenTPM` functions so that the dead code elimination could take care of keeping the net package out of the way. 

```go
func OpenTPM(path string) (io.ReadWriteCloser, error)       // old API for runtime decision
func OpenTPMFile(path string) (io.ReadWriteCloser, error)
func OpenTPMUds(path string) (io.ReadWriteCloser, error)
```

leongross:
any feedback on my proposals? @chrisfenner @rminnich @josephlr

leongross:
@chrisfenner @rminnich @josephlr any thoughts on that?

chrisfenner:
Agh, sorry for the delay @leongross. Let me try to find some time to take a look this week. I need to tinker with this to fully understand why the DCE can't help us yet...

chrisfenner:
First off, sorry for the extensive delays here. We've had trouble with our CI in this repo, and I've been busy with my day-to-day work.

I'm hesitant to add a(nother) build tag here, but I see the problem you're running into.

1. We have this pretty awkward `OpenTPM` stack, even for the new AP:
   * https://github.com/google/go-tpm/blob/ec702090749bc96310fcf40e7c567db1c5feea5b/tpm2/transport/open_other.go#L20
   * calls https://github.com/google/go-tpm/blob/ec702090749bc96310fcf40e7c567db1c5feea5b/legacy/tpm2/open_other.go#L34
   * calls https://github.com/google/go-tpm/blob/ec702090749bc96310fcf40e7c567db1c5feea5b/tpmutil/run_other.go#L29

   tpmutil's `OpenTPM` performs a run-time check of the file mode flags to decide whether to use `net.Dial` to open the UDS, or just open the file. Thus, the dead code eliminator cannot delete either path.

2. Regardless of whether you are using the legacy or the new API, calling `tpmutil` directly or not, you're affected. Also this call stack is weird.

3. I don't prefer adding a(nother) build tag here because it really just bandaids the problem you're having.

4. Making any change to `transport.OpenTPM` will probably break someone.

Here's my suggestion, @leongross please let me know what you think.

Let's add some new transport packages under `tpm2/transport`, one for each type of TPM someone might want to connect to (Linux device file, Linux UDS, Windows, TCP). This requires the caller to be very explicit about which TPM they want to use, rather than relying on some heuristic under the hood and linking to everything.

Normally I'd suggest that you make this change to unblock yourself, but I'm not willing to make that suggestion after taking 6 weeks to really get to this PR, it's not fair to you. So, I will send a PR and add you as a reviewer.

leongross:
@chrisfenner thank you for your extensive explanation and the work you are putting in here.
The architectural change you proposed for the transport packages makes a lot of sense looking at its current state. Making the transport type explicit the entire stack down should solve the problem. As soon as you have something worked out, I will review it as quickly as possible.

chrisfenner:
I sent https://github.com/google/go-tpm/pull/369, PTAL @leongross. I'd especially like to know if tinygo happily compiles a program that uses `linuxtpm.Open` :)

leongross:
I will have a look at it @chrisfenner.

leongross:
This PR can be closed since #369 implemented all the necessary changes.

PR #365, opened 2024-08-05
add TPM2_PolicyDuplicationSelect

salrashid123:
Adds `TPM2_PolicyDuplicationSelect` as described in

[pg 247 23.15 TPM2_PolicyDuplicationSelect](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf)

Specifically, this allows for direct control [where the duplication can occur](https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_duplicate.1.md#example-4-exporting-an-hmac-key-for-a-remote-tpm-and-restrict-it-from-further-exports)

the test cases included here are basic.  i can add an end-to-end gist where a key with this policy is created and imported on a second TPM (todo is  the -ve test where its on a 3rd unauthorized TPM )

@chrisfenner

### DIFF
--- a/tpm2/test/policy_test.go
+++ b/tpm2/test/policy_test.go
@@ -1007,12 +1007,76 @@ func TestPolicyAuthValue(t *testing.T) {
 
 }
 
-func TestPolicyDuplicatonSelect(t *testing.T) {
+func TestPolicyDuplicationSelectUpdate(t *testing.T) {
 	thetpm, err := simulator.OpenSimulator()
 	if err != nil {
 		t.Fatalf("could not connect to TPM simulator: %v", err)
 	}
 	defer thetpm.Close()
+
+	ek, ekcleanup := primaryRSAEK(t, thetpm)
+	defer ekcleanup()
+
+	pk, pkcleanup := primaryRSASRK(t, thetpm)
+	defer pkcleanup()
+
+	k, err := CreateLoaded{
+		ParentHandle: AuthHandle{
+			Handle: pk.Handle,
+			Name:   pk.Name,
+			Auth:   PasswordAuth(nil),
+		},
+		InPublic: New2BTemplate(&TPMTPublic{
+			Type:    TPMAlgRSA,
+			NameAlg: TPMAlgSHA256,
+			ObjectAttributes: TPMAObject{
+				FixedTPM:            false,
+				FixedParent:         false,
+				SensitiveDataOrigin: true,
+				UserWithAuth:        true,
+				SignEncrypt:         true,
+			},
+			Parameters: NewTPMUPublicParms(
+				TPMAlgRSA,
+				&TPMSRSAParms{
+					Scheme: TPMTRSAScheme{
+						Scheme: TPMAlgRSASSA,
+						Details: NewTPMUAsymScheme(
+							TPMAlgRSASSA,
+							&TPMSSigSchemeRSASSA{
+								HashAlg: TPMAlgSHA256,
+							},
+						),
+					},
+					KeyBits: 2048,
+				},
+			),
+		}),
+	}.Execute(thetpm)
+	if err != nil {
+		t.Fatalf("error creating key %v", err)
+	}
+	defer func() {
+		t.Helper()
+		_, err := FlushContext{
+			FlushHandle: k.ObjectHandle,
+		}.Execute(thetpm)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


tpm2/test/policy_test.go
```diff
@@ -981,3 +1006,160 @@ func TestPolicyAuthValue(t *testing.T) {
 	}
 
 }
+
+func TestPolicyDuplicatonSelect(t *testing.T) {
```
chrisfenner, line 1010: nit: typo (should be `TestPolicyDuplicationSelect`)

More substantive : this test only checks that the Update function calculates the right hash. Maybe it should really be called `TestPolicyDuplicationSelectUpdate`?

tpm2/test/policy_test.go
```diff
@@ -981,3 +1006,160 @@ func TestPolicyAuthValue(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			ek, ekcleanup := primaryRSAEK(t, thetpm)
+			defer ekcleanup()
+
+			pk, pkcleanup := primaryRSASRK(t, thetpm)
+			defer pkcleanup()
+
+			if tt.IncludeObject {
```
chrisfenner, line 1037: This test could be a bit more readable if `if tt.IncludeObject` were used to manipulate only `pav`'s `ObjectName` and `IncludeObject` fields rather than kicking off a bunch of other calls.

Since the gist of this test case is "make sure that multiple flavors of `PolicyDuplicationSelect` all compute the right policy hash", we should pretty much do the same setup in all cases, and just do a minor tweak on `pav` before execution depending on the test case.

salrashid123: ok, i think i understand; ptal the latest commit.

tpm2/test/policy_test.go
```diff
@@ -981,3 +1006,156 @@ func TestPolicyAuthValue(t *testing.T) {
...
+				}
+
+				pdr, err = PolicyGetDigest{
+					PolicySession: sess.Handle(),
+				}.Execute(thetpm)
+				if err != nil {
+					t.Fatalf("error executing PolicyGetDigest: %v", pdr)
+				}
+
+			}
```
chrisfenner, line 1142: I still don't quite follow why we need a lot of duplicated code between `if` and `else` here (or even, why we need an if statement at all). What I meant to suggest in https://github.com/google/go-tpm/pull/365/files#r1704689386 is the sequence:

* Create trial session
* Create PolicyDuplicationSelect object where `IncludeObject` is equal to `tt.IncludeObject`
* Run that policy and get the hash
* Create a policy calculator
* Feed the PolicyDuplicationSelect object created above into the calculator
* Check that the hash is the same

chrisfenner (changes requested):
Sorry for the poor response SLO, still have a concern about the design of the test.

tpm2/test/policy_test.go
salrashid123: the reason for the if statements is i wanted to explictly omit the null value ` TPM2BName{}` when sending in the `objectname` and includeobject values...thats more natural use to me.  

this is a testcase so not a big deal so added in this

```golang
	tests := []struct {
		name          string
		objectName    TPM2BName
		includeObject bool
	}{
		{"IncludeObjectFalse", TPM2BName{}, false},
		{"IncludeObjectTrue", k.Name, true},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			pds := PolicyDuplicationSelect{
				PolicySession: sess.Handle(),
				NewParentName: ek.Name,
				ObjectName:    tt.objectName,
				IncludeObject: tt.includeObject,
			}
```

chrisfenner: Yeah, not a big deal either way but thank you for making this test case more concise!

salrashid123:
@chrisfenner thanks; can someone trigger the tests?  (they do pass but it seems there was a timeout/resource issue with the test vm provisioning)

chrisfenner:
Yeah I re-triggered them a while ago... maybe they need to be re-rerun. I'll override this check if it's failing and preventing checkin today.

chrisfenner:
cirrus CI is having a bad day, will force merge this

PR #366, opened 2024-08-22
Add TPMFW (firmware) handles and FirmwareLimited attribute to TPMAObject 

destinideinde:
Add TPMFW (firmware) handles to go_tpm from updated TCG specs at https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-2-Structures.pdf#page83 and add FirmwareLimited attribute to TPMAObject and allow to the validation to accept keys with and without the bit set

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -4,7 +4,6 @@ package tpm2
 import (
 	"bytes"
 	"encoding/binary"
-
 	"github.com/google/go-tpm/tpm2/transport"
 )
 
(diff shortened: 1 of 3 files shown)


tpm2/tpm2.go
```diff
@@ -4,8 +4,10 @@ package tpm2
 import (
 	"bytes"
 	"encoding/binary"
+	"io"
```
chrisfenner, line 7: I'm not sure this import is needed - does `goimports` add it?

destinideinde: This was due to an erroneous pull and merge, I believe it is fixed now?

chrisfenner (approved):

tpm2/tpm2.go
chrisfenner: Perfect, thank you!

chrisfenner:
CI seems a little broken. This is a very small, safe change so I will merge it and investigate the CI failures separately.

PR #369, opened 2024-09-18
Create individual packages for Windows and Linux TPM transport

chrisfenner:
https://github.com/google/go-tpm/pull/364 called to attention some long-standing technical debt around TPM transport. In particular, the stack looks like:

(Linux or Windows) `OpenTPM` function
calls the legacy `OpenTPM` function
calls the tpmutil `OpenTPM` function

At the bottom of the stack, tpmutil does some runtime introspection to see what type of TPM it wants to open (e.g., on Linux, the device could be either a device file or a socket). This runtime support is convenient, but also breaks dead-code elimination (for example, tinygo will fail to compile the UDS support code, and users have no way of leaving that out without patches).

In principle, we've found within Google that "open my TPM" should be as un-smart as possible, to avoid awkward edge cases (for example, what happens if the logic finds two different TPMs on the system; which should it prefer; should it invisibly succeed and surprise the user?). Instead, the preferred pattern is to require the user to explicitly say which TPM they are trying to open.

This change introduces 3 packages as a replacement for `transport.OpenTPM` (which this change marks as now Deprecated):

`transport/linuxtpm.Open(path)` opens Linux device TPMs (e.g., /dev/tpm0 or /dev/tpmrm0)
`transport/linuxudstpm.Open(path)` opens Linux Unix Domain Socket TPMs `transport/windowstpm.Open()` opens the TPM from TBS.dll

Intentionally, the now-deprecated `transport.OpenTPM` is not touched. This would create an import cycle.

TODO: Manual testing on Linux and Windows TPMs, since our testing uses the TPM simulator.

### DIFF
--- a/tpm2/transport/linuxtpm/linuxtpm.go
+++ b/tpm2/transport/linuxtpm/linuxtpm.go
@@ -1,0 +1,36 @@
+//go:build !windows
+
+// Package linuxtpm provides access to a physical TPM device via the device file.
+package linuxtpm
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/google/go-tpm/tpm2/transport"
+)
+
+var (
+	// ErrFileIsNotDevice indicates that the TPM file mode was not a device.
+	ErrFileIsNotDevice = errors.New("TPM file is not a device")
+)
+
+// Open opens the TPM device file at the given path.
+func Open(path string) (transport.TPMCloser, error) {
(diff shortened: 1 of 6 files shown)


alexmwu (approved):
Really nice clean up!

leongross (approved):
I tested it out in said u-root commands from PR #364 with tinygo and the prior issues seem to be resolved. Thanks a lot for taking care of the improved implementation. This will help us a lot!

PR #371, opened 2024-09-20
Fix KnownName() and Empty Buffer sequence handle Names

### LINKED ISSUE #370: Invalid handling of KnownName() and Empty Buffer responses
KnownName() has a case for `TPMHTTransient` that conflates transient handles with sequence object handle values of a TPMS_CONTEXT.  Undefined behavior will result when more than one transient handle exists.

The TPM specification also defines an Empty Buffer as the Name for sequence objects (Part 1: Architecture, section 32.4.5), which is a special case that must be handled for `TPM2_SequenceUpdate` and `TPM2_SequenceComplete` commands.  Users of the library will receive command errors of the form:

```text
SequenceComplete failed: missing Name for 'SequenceHandle' parameter
```

This error should not occur because `Name` for these commands is the Empty Buffer.

nckrss:
This PR resolves a bug introduced to KnownName handling and introduces a struct tag for commands that allow for the `Name` to be the Empty Buffer, namely TPM2_SequenceUpdate and TPM2_SequenceComplete.

- [x] Add tests for multiple transient objects in an HMAC sequence
- [x] Handle special case for sequence object handle Name fields
- [x] Relax HmacStart.Handle type from AuthHandle to handle
- [x] Fix prior RSA test that silently succeeded due to KnownName() bug

Fixes #370

### DIFF
--- a/tpm2/reflect.go
+++ b/tpm2/reflect.go
@@ -815,8 +815,10 @@ func cmdNames[R any](cmd Command[R, *R]) ([]TPM2BName, error) {
 		}
 
 		// Special case: handles with an empty name buffer (anonymous:anon)
-		// The Name of a sequence object is an Empty Buffer
-		// See part 1: Architecture, section 32.4.5
+		// See part 1: Architecture, section 32.4.5:
+		// The Name of a sequence object is an Empty Buffer (sized array with no
+		// data; indicated by a size field of zero followed by an array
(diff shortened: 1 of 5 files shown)


chrisfenner:
Thank you so much for the bug report and the accompanying fix, @nckrss!

chrisfenner (approved):

PR #381, opened 2024-12-16
Improve crypto & constants

AlexandreEXFO:


tpm2/constants.go
```diff
@@ -63,15 +83,20 @@ type TPMECCCurve uint16
...
+	TPMECCNone              TPMECCCurve = 0x0000
+	TPMECCNistP192          TPMECCCurve = 0x0001
+	TPMECCNistP224          TPMECCCurve = 0x0002
+	TPMECCNistP256          TPMECCCurve = 0x0003
+	TPMECCNistP384          TPMECCCurve = 0x0004
+	TPMECCNistP521          TPMECCCurve = 0x0005
+	TPMECCBNP256            TPMECCCurve = 0x0010
+	TPMECCBNP638            TPMECCCurve = 0x0011
+	TPMECCSM2P256           TPMECCCurve = 0x0020
+	TPMECCBrainpool_P256_R1 TPMECCCurve = 0x0030
```
chrisfenner, line 95: nit: Current style is to omit underscores from the names (even though that can lead to initialisms being next to each other) in accordance with Google's general [Go style guide](https://google.github.io/styleguide/go/decisions#naming)

chrisfenner (changes requested):
Thanks for this work! I have just one nit to pick about naming, then we can merge. I am also working on some testing for this code that uses TPM2_Duplicate to check that we get valid/compatible private and public keys out of this code. I'll integrate my changes atop yours.

chrisfenner (approved):

PR #382, opened 2024-12-16
Test Pub() and Priv() by exporting the key and signing/verifying from software

chrisfenner:
This PR enhances the crypto_test.go unit-test by adding a test that creates a TPM key, exports it, and then checks that TPM and software can mutually sign/verify one another.

### DIFF
--- a/tpm2/crypto_test.go
+++ b/tpm2/crypto_test.go
@@ -526,7 +526,7 @@ func TestRoundTrip(t *testing.T) {
 				FlushHandle: load.ObjectHandle,
 			}.Execute(tpm)
 
-			pub, priv := duplicateAndImport(t, tpm, load.ObjectHandle)
+			pub, priv := duplicateToSoftware(t, tpm, load.ObjectHandle)
 			checkSignVerify(t, tpm, NamedHandle{
 				Handle: load.ObjectHandle,
 				Name:   load.Name,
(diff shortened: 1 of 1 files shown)


chrisfenner:
+CC @AlexandreEXFO in case you have comments on this one.

AlexandreEXFO:
@chrisfenner : There are errors in the error messages of the TestRoundTrip function since there are RSA and ECC keys. Otherwise, it's good.

By curiosity, do you plan to make a new version soon with the changes? I would like to stop depending on a fork for a project that uses this library.

chrisfenner:
> @chrisfenner : There are errors in the error messages of the TestRoundTrip function since there are RSA and ECC keys. Otherwise, it's good.
> 
> By curiosity, do you plan to make a new version soon with the changes? I would like to stop depending on a fork for a project that uses this library.

Thank you! I will kick off another release after this PR goes in.

tpm2/crypto_test.go
```diff
@@ -368,3 +373,365 @@ func TestPub(t *testing.T) {
...
+			}, tc.alg, pub, priv)
+		})
+	}
+}
+
+// Uses the `Pub` and `Priv` helpers to import the public and private key.
+// Note that in practice, one could just use `crypto.PrivateKey.Public()` to
+// derive the public key from the private key. This helper uses the `Pub`
+// routine to validate that function specifically while we test the rest.
+func duplicateAndImport(t *testing.T, tpm transport.TPM, h TPMHandle) (crypto.PublicKey, crypto.PrivateKey) {
```
jkl73, line 542: Seems like this function duplicates and exports the key, but it never "import" the key back to the TPM, maybe need to updat e the function name?

jkl73 (approved):

tpm2/crypto_test.go
chrisfenner: Oh yeah, I'll clarify. It "imports" it to software :)

chrisfenner:
> > @chrisfenner : There are errors in the error messages of the TestRoundTrip function since there are RSA and ECC keys. Otherwise, it's good.
> > By curiosity, do you plan to make a new version soon with the changes? I would like to stop depending on a fork for a project that uses this library.
> 
> Thank you! I will kick off another release after this PR goes in.

https://github.com/google/go-tpm/releases/tag/v0.9.3

PR #383, opened 2024-12-20
Tolerate differences in RSA private key libraries

chrisfenner:
`GOEXPERIMENT=boringcrypto go test ./... --test.run TestPriv/valid_rsa` fails occasionally due to the fact that the boringcrypto implementation uses the Carmichael function (lambda(N)) instead of Euler's totient function (phi(N)). Sometimes we get lucky (where phi(N) == lambda(N)) and sometimes we don't.

Unfortunately, the TPM RSA private key structure only contains a prime, for compactness. This means that when importing TPM RSA private keys, we have to compute D from the factorization of N ourselves (see crypto.go). This test generates a key in software, converts it into a TPM private key structure, and then checks that the import function returns the same key. Unfortunately, the software key-generation algorithm may disagree with crypto.go's technique (which is currently to use phi(N)).

This change introduces a helper function to the test for comparing RSA private keys by public exponent, modulus, and primes only.

### DIFF
--- a/tpm2/crypto_test.go
+++ b/tpm2/crypto_test.go
@@ -11,8 +11,6 @@ import (
 	"reflect"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/go-tpm/tpm2/transport"
 	"github.com/google/go-tpm/tpm2/transport/simulator"
 )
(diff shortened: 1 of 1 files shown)


rsc:
Not sure you need the import of cmp to handle slice order here. You passed Primes[0] into the TPM struct and then used that field as Primes[0] on the way out too. If they came out reversed that would be a surprise indeed!

chrisfenner:
> Not sure you need the import of cmp to handle slice order here. You passed Primes[0] into the TPM struct and then used that field as Primes[0] on the way out too. If they came out reversed that would be a surprise indeed!
> 
> Otherwise LGTM.

Ha, of course. Too paranoid. I've simplified this, thanks!

alexmwu (approved):

PR #384, opened 2024-12-20
Improve TPMUSymKeyBits and TPMUSymMode

AlexandreEXFO:


tpm2/structures.go
```diff
@@ -1491,7 +1506,21 @@ func NewTPMUSymMode[C SymModeContents](selector TPMAlgID, contents C) TPMUSymMod
...
+		value := u.contents.(*boxed[TPMIAlgSymMode]).unbox()
+		return value, nil
+	default:
+		return nil, fmt.Errorf("did not contain sym (selector value was %v)", u.selector)
+	}
+}
+
 // AES returns the 'aes' member of the union.
+//
+// Deprecated: AES exists for historical compatibility
```
chrisfenner, line 1522: The structure does indeed have members for each algorithm, see https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-2-Structures.pdf#page=150 (and see https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-2-Structures.pdf#page=32 for an explanation of the algorithm tokens). `sym` is there as a convenience within the reference implementation and IMHO probably shouldn't even be mentioned in the spec. Also, I'm expecting that a future version of the spec flattens these macros, because they're hard to understand.

tpm2/structures.go
```diff
@@ -1446,7 +1461,7 @@ type SymModeContents interface {
 // create implements the unmarshallableWithHint interface.
 func (u *TPMUSymMode) create(hint int64) (reflect.Value, error) {
 	switch TPMAlgID(hint) {
-	case TPMAlgAES:
+	case TPMAlgTDES, TPMAlgAES, TPMAlgSM4, TPMAlgCamellia:
```
chrisfenner, line 1464: TDES is deprecated (https://csrc.nist.gov/news/2023/nist-to-withdraw-sp-800-67-rev-2) so can we just not implement support for it?

chrisfenner (changes requested):
Thanks for this change! I have a couple comments.

tpm2/structures.go
AlexandreEXFO: Oh! I really didn't understand the specification like that the first time I look at it (it was really not clear), but looking at it a second time, I actually understand.

tpm2/structures.go
chrisfenner: This is a fair point.

Can we mark the relevant fields as [Deprecated](https://go.dev/wiki/Deprecated) to discourage new uses?

Thanks
Chris

AlexandreEXFO: Yes, sure.

chrisfenner (approved):

chrisfenner:
Sorry for the delayed review @AlexandreEXFO! Thanks for this contribution!

PR #395, opened 2025-04-19
Add generalized support for Restricted Decryption operations (ActivateCredential, Import, StartAuthSession)

chrisfenner:
Asymmetric restricted decryption keys (e.g., EK/SRK) can be either RSA or ECC and support the following flows:

- Key attestation with ActivateCredential
- Object unwrap with Import
- Salted session establishment with StartAuthSession

This change splits apart the algorithm-specific asymmetric crypto, and the parent-algorithm-agnostic protocols. This PR names the first part "Labeled Key Encapsulation" because it is very much like a KEM with an additional label (that depends on which protocol is being used). The advantage of this splitup, aside from aiding understandability, is that new algorithms for PQC can be added to the "Labeled KEM" side without touching the protocols.

This change adds helper functions that can be used to create attestation credentials/challenges as well as object import blobs, and refactors the existing salted session code to leverage the same Labeled KEM code.

This change vendors the test vectors from https://github.com/chrisfenner/tpm-test-vectors, which provide comprehensive coverage for this crypto protocol code.

This change introduces a .gitallowed file, because `git-secrets` was yelling at me about the JSON files containing the string "KEY" :)


chrisfenner:
In parallel to this, I started working on test vectors. I've got some test vectors now, so I'm going to refactor the implementation a bit to make use of them.

alexmwu (approved):
Neat! The new "Table 27: Protection Values" table saved me a lot of effort, nice job on the spec updates!

PR #396, opened 2025-04-24
Check for padding bytes in EK cert

dwaynebradley:
Some TPM chips "pad" the EK cert that is stored in the TPM to completely fill up the NV index. Most TPM chips do NOT do this. If we try to parse the certificate with this extra padding, the `x509.ParseCertificate` function will return an "trailing bytes" error.

Before trying to parse the certificate, we need to identify how many bytes should be stripped from bytes that were read from the TPM. To do this, we can use the ASN.1 `Unmarshal` function to get the "padding" bytes. From there, we can use the length of the padding bytes and remove those bytes from the bytes we send to the `x509.ParseCertificate` function so it will parse correctly.

### DIFF
--- a/examples/tpm2-ekcert/main.go
+++ b/examples/tpm2-ekcert/main.go
@@ -6,6 +6,7 @@ package main
 import (
 	"crypto"
 	"crypto/x509"
+	"encoding/asn1"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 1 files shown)


chrisfenner:
Hi @dwaynebradley! Thanks for the suggestion.

go-tpm is more about support for software that interfaces with the [TPM 2.0 standard](https://trustedcomputinggroup.org/resource/tpm-library-specification/), whereas a helper function for cleaning up x.509 certs (that happen to be stored in TPM NV) might be better suited for https://github.com/google/go-tpm-tools.

PR #402, opened 2025-05-20
Allow policy session to be extended more programatically

### LINKED ISSUE #401: Consider an update to PolicyCommand or a new interface for working with policy sessions
The PolicyCommand interface is really useful for calculating the correct policy digest to attach to an object but it's a bit harder to use for creating actual policy sessions and writing generic code to work with policies.

Would it be possible to add a method like SetPolicySession(Session) so it would be easier to write more generic code for extending policy sessions?

friedrichsenm:
This PR updates the PolicyCommand interface to add ExecutePolicyInSession. This new method extends the given policy session with the policy command and returns the response.

Closes #401

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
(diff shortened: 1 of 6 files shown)


friedrichsenm:
Looks like an account I didn't intend for might have gotten attached to the commit. I might have to close this and reopen again after making sure I didn't goof on the commit history.

PR #403, opened 2025-05-20
Allow policy session to be extended more programatically

friedrichsenm:
This PR updates the PolicyCommand interface to add ExecutePolicyInSession. This new method extends the given policy session with the policy command and returns the response.

### DIFF
--- a/tpm2/tpm2.go
+++ b/tpm2/tpm2.go
@@ -69,10 +69,13 @@ type Command[R any, PR *R] interface {
 }
 
 // PolicyCommand is a TPM command that can be part of a TPM policy.
-type PolicyCommand interface {
+type PolicyCommand[R any, PR *R] interface {
 	// Update updates the given policy hash according to the command
 	// parameters.
 	Update(policy *PolicyCalculator) error
+	// ExecutePolicyInSession executes the given policy command,
+	// using the given session as PolicySession.
+	ExecutePolicyInSession(t transport.TPM, s Session) (PR, error)
 }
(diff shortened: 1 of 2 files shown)


friedrichsenm:
It looked like most of the policy test were testing the functionality of the policy commands themselves so I'm not sure if the one test I modified is considered meaningful enough. I can add a separate test to just test this interface (perhaps with a trial session or something) if you think that is appropriate.

tpm2/tpm2.go
```diff
@@ -69,10 +69,13 @@ type Command[R any, PR *R] interface {
 }
 
 // PolicyCommand is a TPM command that can be part of a TPM policy.
-type PolicyCommand interface {
+type PolicyCommand[R any, PR *R] interface {
```
chrisfenner, line 72: Couple options come to mind here:

1. We can toss the (non-error) response from the policy command, assuming that users who want to consume the output of the Policy command are not doing so using this interface.
2. We could make this interface function just return any, error and leave it to the caller to do the type assertion if they wish to consume the data.

What do you think? Do either of these options break your use-case?

friedrichsenm:
@chrisfenner, I found an `Update` method that was missing the `error` return parameter and added that as a second commit. I can put that in a separate PR if you'd prefer.

chrisfenner:
Hi @friedrichsenm, sorry for losing track of this. If you're still interested in the change would you mind rebasing to resolve the merge conflicts?

PR #408, opened 2025-07-29
Fix PolicySecret implementation of the PolicyCommand interface

hennk:
`PolicySecret` now correctly implements the `PolicyCommand` interface, which requires to return an `error`


chrisfenner (approved):
Thank you for the fix!

PR #409, opened 2025-07-29
Add TPM2_GetTime Implementation

qwwqwwq:
See section 18.7 of the [spec](https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-3-Commands.pdf)

This is a useful command as it allows you to "tick stamp" arbitrary data, associating it with both with a signing key and a particular TPM clock state. Separately calling the TPM2_ReadClock/TPM2_Sign commands can not accomplish this.

The command code for TPM2_GetTime is enumerated in [constants.go](https://github.com/google/go-tpm/blob/6217e7f085d7021a2ff7a6a8feaea5c3063fd295/tpm2/constants.go#L152) but the actual command implementation does not exist, so this PR adds it.

### DIFF
--- a/tpm2/test/get_time_test.go
+++ b/tpm2/test/get_time_test.go
@@ -1,0 +1,100 @@
+package tpm2test
+
+import (
+	"bytes"
+	"testing"
+
+	. "github.com/google/go-tpm/tpm2"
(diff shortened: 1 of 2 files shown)


chrisfenner (approved):
Thanks for adding this!

Issue #414, opened 2025-09-17
[Question] How to produce an audit log to remotely validate an audit session?


loicsikidi:
Hi,

I'm currently working on *audit session* in order to achieve the following goals:

1. **remotely attest that a TPM really executed a sequence of commands w/ success**
2. **(bonus) get the Response of each audited command**

After reading a bunch of resources:

* [A Practical Guide to TPM 2.0: Chapter 16 Auditing TPM Commands](https://library.oapen.org/viewer/web/viewer.html?file=/bitstream/handle/20.500.12657/28157/1001837.pdf?sequence=1&isAllowed=y&page=263)
* *TCG TPM 2.0 Part 1: Architecture v184*
  * [section 18 Audit Session](https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-2.0-Library-Part-1-Version-184_pub.pdf#page=148)
  * [section 33 Command Audit](https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-2.0-Library-Part-1-Version-184_pub.pdf#page=224)

I've unsterstood the following: we have an untrusted component[^1] (outside the TPM) which maintain an audit log w/ TPM command & response parameters and  `TPM2_GetSessionAuditDigest` allows to get an attestation about the digest of the audit session.

Theoretically, from the *untrusted audit log* a remote party is able to recompute a digest and check if it matches the one produced by `TPM2_GetSessionAuditDigest` to know if it can be trusted.

https://github.com/google/go-tpm/blob/bf120203f23af1e4291a336cb93a7a79e2a0832f/tpm2/audit.go#L17-L53

`CommandAudit` & `AuditCommand` functions seems to provide this logic but not in a way that is suitable for my use case.

Ideally would like to obtain this kind of output:

```
hash(commandA)<DELIMITER>marshal(responseA)
hash(commandB)<DELIMITER>marshal(responseB)
hash(commandC)<DELIMITER>marshal(responseC)
...
```

1. It would allow to keep command privacy[^2]
1. It would allow the remote party to unmarshal the response to access some values (e.g. *GetCapability, etc.*)
1. It would allow to reproduce the hash w/ the same logic implemented in `AuditCommand` (i.e. *auditRPHash*)

It doesn't seems that Command or/and Response to be marshallable which prevent me to achieve my goals.

Does my request make sense or I miss something obvious?

Thank you in advance for your support!

[^1]: an application
[...]

PR #415, opened 2025-09-22
feat: add marshal/unmarshall to command/response

loicsikidi:
a naive implementation of Command/Response marshalling  in order to address #414 

I'm open to feedback if there a better way to provide this feature.


tpm2/marshalling.go
```diff
@@ -126,3 +126,85 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+	cmd, err = UnmarshalCommand[C, R](cmdData)
+	if err != nil {
+		return cmd, rsp, fmt.Errorf("unmarshalling command: %w", err)
+	}
+	rsp, err = UnmarshalResponse[R](rspData)
+	if err != nil {
+		return cmd, rsp, fmt.Errorf("unmarshalling response: %w", err)
+	}
+	return cmd, rsp, nil
+}
```
chrisfenner: Nit: is it that much easier to call these commands vs. MarshalCommand+MarshalResponse, or UnmarshalCommand+UnmarshalResponse? Not sure they are needed.

tpm2/marshalling.go
```diff
@@ -126,3 +126,85 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+
+// MarshalCommand marshals a TPM command.
+func MarshalCommand[C Command[R, *R], R any](cmd C) ([]byte, error) {
+	var buf bytes.Buffer
+	params := taggedMembers(reflect.ValueOf(cmd), "handle", true)
```
chrisfenner: Consider using `cmdParameters` for this?

tpm2/marshalling.go
```diff
@@ -126,3 +126,85 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+	for i := range len(params) {
+		if err := marshalParameter(&buf, cmd, i); err != nil {
+			return nil, fmt.Errorf("marshalling command's parameter: %w", err)
+		}
+	}
+	return buf.Bytes(), nil
```
chrisfenner, line 172: In #414 we discussed how the audit verifier, who needs to verify `cpHash` (the command parameter hash), needs not only the command parameters, but also the names. Instead of returning just `parameters`, should we consider returning a buffer that directly hashes to `cpHash`? That buffer is ( commandCode || name1 [ || name2 [ || name3 ]] || parms ).

See CPHash, which I think is similar to what you need (except obviously that it returns the digest when you want the preimage): https://github.com/google/go-tpm/blob/bf120203f23af1e4291a336cb93a7a79e2a0832f/tpm2/sessions.go#L56

tpm2/marshalling.go
```diff
@@ -126,3 +126,85 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+}
+
+// UnmarshalCommand unmarshals a TPM command.
+func UnmarshalCommand[C Command[R, *R], R any](data []byte) (C, error) {
```
chrisfenner, line 187: Open: What should this function do if the command parameters were encrypted with a session?

tpm2/marshalling.go
```diff
@@ -126,3 +126,85 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+	for i, parameter := range parameters {
+		if err := marshal(&buf, parameter); err != nil {
+			return nil, fmt.Errorf("marshalling response parameter %d: %w", i, err)
+		}
+	}
+	return buf.Bytes(), nil
+}
+
+// UnmarshalResponse unmarshals a TPM response.
+func UnmarshalResponse[R any](data []byte) (*R, error) {
```
chrisfenner, line 344: Open: What should this function do if the response parameters were encrypted with a session?

tpm2/marshalling.go
```diff
@@ -126,3 +126,85 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+}
+
+// UnmarshalCommand unmarshals a TPM command.
+func UnmarshalCommand[C Command[R, *R], R any](data []byte) (C, error) {
+	var cmd C
+	if data == nil {
+		return cmd, fmt.Errorf("data cannot be nil")
+	}
+	buf := bytes.NewBuffer(data)
+	params := taggedMembers(reflect.ValueOf(cmd), "handle", true)
```
chrisfenner: Could the reflection-heavy parts of this code be turned into an (internal) helper that lives in https://github.com/google/go-tpm/blob/main/tpm2/reflect.go? We try to keep the reflection magic contained to within there.

Basically, this helper would be the inverse of https://github.com/google/go-tpm/blob/bf120203f23af1e4291a336cb93a7a79e2a0832f/tpm2/reflect.go#L859

chrisfenner: (Same for the response structure)

tpm2/marshalling.go
chrisfenner: And possibly similarly for the MarshalResponse.

tpm2/marshalling.go
chrisfenner: If we chose the marshaled ( commandCode || name1 [ || name2 [ || name3 ]] || parms ) as the thing to be returned by this function, it would be neat to provide an Unmarshal function that takes the whole thing and returns (TPM_CC, TPM2B_Name[], Command, error). Since the CC is there, we can easily look up how many names to expect and fail fast & obvious when there is a mismatch.

tpm2/marshalling.go
loicsikidi: done in `MarshalCommand` :heavy_check_mark:

tpm2/marshalling.go
loicsikidi: Indeed I didn't implemented your last discussion :smile:.

Finally, I changed my approach and I’m using a struct as a pivot to represent the preimage in binary format.

*Reason: during unmarshal, I want to make my life easier when determining the size of each `name`.*

https://github.com/loicsikidi/go-tpm/blob/59f5c5a4f4ff30dba3f4b4a73885aba07aa695cc/tpm2/marshalling.go#L36-L44

Is it ok for you?

tpm2/marshalling.go
loicsikidi: Could we address this topic in another PR?

---

PS: if I had to implement this, intuitively I would use this signature (but I don't if it's a good idea):

```go
func UnmarshalCommand[C Command[R, *R], R any](data []byte, s ...Session) (C, error) {
```

Knowing the following:

https://github.com/google/go-tpm/blob/bf120203f23af1e4291a336cb93a7a79e2a0832f/tpm2/reflect.go#L859

tpm2/marshalling.go
loicsikidi: same answer as above

tpm2/marshalling.go
loicsikidi: I've moved all reflection in reflect.go

Tell me what you think about the new implementation of [UnmarshalCommand](https://github.com/loicsikidi/go-tpm/blob/59f5c5a4f4ff30dba3f4b4a73885aba07aa695cc/tpm2/marshalling.go#L230) and [UnmarshalResponse](https://github.com/loicsikidi/go-tpm/blob/59f5c5a4f4ff30dba3f4b4a73885aba07aa695cc/tpm2/marshalling.go#L275)

tpm2/marshalling.go
```diff
@@ -126,3 +166,119 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+		return 0, nil, nil, fmt.Errorf("unmarshalling CommandPreimage: %w", err)
+	}
+
+	return preimage.CommandCode, preimage.Names, preimage.Parameters.Buffer, nil
+}
+
+// UnmarshalCommand unmarshals a serialized [CommandPreimage] back into a TPM command.
+// The data should be the output from [MarshalCommand].
+//
+// Note: command produced from this function is not meant to be executed directly on a TPM,
```
loicsikidi: Question: @chrisfenner do you agree with this statement?

loicsikidi:
@chrisfenner once this PR will be approved it will be required to **squash** because latest commit messages won't be useful.

tpm2/marshalling.go
```diff
@@ -22,6 +23,45 @@ type marshallableWithHint interface {
 	get(hint int64) (reflect.Value, error)
 }
 
+// CommandPreimage represents a structured preimage of cpHash for a TPM command.
+// This structure is marshaled to bytes using [Marshal] for storage/transmission
```
chrisfenner: Is it necessary for this structure to support the reflection based marshalling? That system is intended for structures that need to be 1:1 for some TPM structure.

It looks like we use `binary.Write` directly in the implementation. So I think we can drop the embed of `marshalByReflection`.

tpm2/marshalling.go
```diff
@@ -22,6 +23,45 @@ type marshallableWithHint interface {
...
+	CommandCode TPMCC
+	// Names are the names of the handles referenced by the command
+	Names []TPM2BName `gotpm:"list"`
+	// Parameters are the marshaled command parameters
+	Parameters TPM2BData
+}
+
+// ToCPHashPreimage converts the CommandPreimage to the raw buffer format
+// used to compute cpHash according to TPM 2.0 spec.
+func (cp *CommandPreimage) ToCPHashPreimage() []byte {
```
chrisfenner: Why not use the raw preimage of cphash/rphash for the serialized format, so that a caller can directly compute those hashes from the serialized byte arrays rather than unmarshalling and remarshalling again with `ToCPHashPreimage`?

If it's a matter of encoding how many handles each command/response uses, it might be possible to write a helper that uses `cmdNames` and just counts the size.

tpm2/marshalling.go
chrisfenner: Yeah, because session keys will be missing. I guess someone could rehydrate their command with this and then add the sessions on top, which could be neat. But it seems orthogonal to your use-case, so I like how you've written this comment already (somebody else can enhance the feature to make it easier to tape-on sessions later if needed) :)

tpm2/marshalling.go
chrisfenner: Absolutely, "do it later" or even "don't do it" are fine answers :)

Since it might be hard for us to introspect a serialized command structure to see if it's encrypted, it might be sufficient to just write (in a comment) "command/response parameters that were encrypted with a session are not supported".

tpm2/marshalling.go
chrisfenner: Works for me, we can resolve them both the same way (I propose by commenting explicitly that it is not supported to do that)

tpm2/reflect.go
```diff
@@ -854,6 +854,187 @@ func marshalParameter[R any](buf *bytes.Buffer, cmd Command[R, *R], i int) error
 	return marshal(buf, parm)
 }
 
+// unmarshalParameter will deserialize the given parameter of the command from the buffer.
+// Returns an error if the value is not unmarshallable or if there's insufficient data.
+func unmarshalParameter[C Command[R, *R], R any](buf *bytes.Buffer, cmd *C, i int) error {
```
chrisfenner, line 894: It seems like this function duplicates a lot of code from `unmarshalStruct`. Can we deduplicate?

tpm2/reflect.go
```diff
@@ -854,6 +854,187 @@ func marshalParameter[R any](buf *bytes.Buffer, cmd Command[R, *R], i int) error
...
+	return unmarshal(buf, parm)
+}
+
+// populateHandlesFromNames populates the handle fields of a command with NamedHandles
+// created from the provided names.
+//
+// Supports different handle types based on struct tags:
+//   - gotpm:"handle" → NamedHandle (without handle value)
+//   - gotpm:"handle,auth" → AuthHandle (without handle value and nil Auth)
+func populateHandlesFromNames[C Command[R, *R], R any](cmd *C, names []TPM2BName) error {
```
chrisfenner, line 910: Instead of trying to recover the bare handle values from the Names (which is rarely even possible) why not populate the struct with a `tpm2.handle` interface implementor struct that gives you the name (implements `KnownName()`) but not the actual handle value (e.g., return a fixed invalid handle value if `HandleValue()` gets called, bail out from `cmdHandles` if encountered)

loicsikidi: ok I'll refactor this part :rocket:

tpm2/reflect.go
loicsikidi: ok I'll try to refactor this part :rocket:

tpm2/marshalling.go
loicsikidi: I've made a refactoring to migrate to a `binary.Write` implementation but I think it's adds a bit of complexity.

Tell me what you think

tpm2/marshalling.go
loicsikidi: Yeah you right, I'll remove this function

tpm2/marshalling.go
loicsikidi: I hope to address *encrypted session* in the future

tpm2/marshalling.go
```diff
@@ -27,39 +27,71 @@ type marshallableWithHint interface {
...
-func (cp *CommandPreimage) ToCPHashPreimage() []byte {
-	var buf bytes.Buffer
-
-	// Write command code (4 bytes, big endian)
-	binary.Write(&buf, binary.BigEndian, cp.CommandCode)
-
-	// Write names (raw buffers without size prefix)
+// Marshal converts the CommandPreimage to its byte representation.
+//
+// Format:
```
chrisfenner: Are you sure about this? Per https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-2.0-Library-Part-1-Version-184_pub.pdf#page=103:

Should be CommandCode, followed by each Name, followed by the parameters.

Just to check we are on the same page here, I think `MarshalCommand` (and `MarshalResponse`) should return the actual bytes that should be hashed, rather than some other format. I prefer this because it will make the caller's code pretty obvious.

```go
cmdData, err := MarshalCommand(myCmd)
// handle err
cpHash := sha256.Sum256(cmdData)
// do something interesting with cpHash
```

Note that this will require some smarts on the part of the library in order to properly do `UnmarshalCommand` and `UnmarshalResponse`. i.e., we have to look up the command code to see how many handles (names) to expect for the given command/response, as you do for the check on line 278 on the current commit.

tpm2/marshalling.go
chrisfenner: Thanks for doing it this way - I agree that this code is more verbose, but it makes it more clear what is happening (and it's thus easier to align it with the spec, see my comment below)

tpm2/marshalling.go
```diff
@@ -22,6 +23,77 @@ type marshallableWithHint interface {
 	get(hint int64) (reflect.Value, error)
 }
 
+// CommandPreimage represents a structured preimage of cpHash for a TPM command.
+// This structure is marshaled to bytes using [Marshal] for storage/transmission
+// and can be converted to the raw cpHash preimage format for hashing.
+//
+// See definition in Part 1: Architecture, section 16.7.
+type CommandPreimage struct {
```
chrisfenner: I don't think exporting this type is necessary, since it's an implementation detail of [Marshal, Unmarshal][Command, Response].

tpm2/marshalling.go
```diff
@@ -126,3 +198,125 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+	buf := bytes.NewBuffer(params)
+	if err := unmarshalCmdParameters(buf, &cmd, nil); err != nil {
+		return cmd, err
+	}
+	return cmd, nil
+}
+
+// MarshalResponse marshals a TPM response.
+func MarshalResponse[R any](rsp *R) ([]byte, error) {
+	return marshalRspParameters(rsp, nil)
```
chrisfenner: (In the same vein of my above comment) Should `MarshalResponse` return the preimage of `rpHash`? If so, per https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-2.0-Library-Part-1-Version-184_pub.pdf#page=104 that is responseCode || commandCode || parameters. Note that responseCode is always 0 because rpHash is not computed for failed commands.

tpm2/reflect.go
chrisfenner: Thanks so much for doing this!

tpm2/reflect.go
chrisfenner: Thanks for doing this!

chrisfenner (changes requested):
Thanks for your patience on my review, @loicsikidi -- your reply came in while I was at a conference, and then I have been out sick the past week. I think we are very close here, just a couple more things to make sure that this satisfies your original use case of making it as easy as possible for users to compute cpHash/rpHash for audit- and audit-adjacent TPM activities.

tpm2/marshalling.go
loicsikidi: Hi @chrisfenner,

Before starting the implementation, I would like to have your advice first.

```go
cmdData, err := MarshalCommand(myCmd)
// handle err
cpHash := sha256.Sum256(cmdData)
// do something interesting with cpHash
```

I agree w/ this spec! 

But when I first try to do `cmd, err := UnmarshalCommand(cmdData)`, I came accross the following problematic: *how to know the size of each name?*

* Does the name's size (bit length) is always the same or it can differ?
* If it differs, what should be rule to determine the size?
   * Historically, I tried to address this w/ 2 strategies:
      * 1st: implement `CommandPreimage` as a *marshalByReflection* in order to benefit Marshal/Unmarshal functions. In addition, I've added a method to `CommandPreimage` to return the cpHash value.
      * 2d (current implementation): add some fields (during the marshaling) to store each field's size

> we have to look up the command code to see how many handles (names) to expect for the given command/response

Thanks to this I know the number of names I should expect.

Ty in advance for your help :pray:

chrisfenner: Great question, and yes this is tricky. However a TPM Name is one of the two formats (https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-2.0-Library-Part-1-Version-184_pub.pdf#page=87)

1. For PCRs, HMAC Sessions, Policy Sessions, and Permanent Values: the Name is a 32 bit handle.
2. For all other Names, it is (nameAlg || H_nameAlg).

Case 2 is easy if you have a lookup table of hash algorithm IDs and fail-out any time an unrecognized algorithm is encountered.
Case 1 can be handled for leading bytes 0x0000... (PCRs), 0x02... (HMAC Sessions), 0x03... (Policy Sessions), and 0x40... (Permanent Values). Note that all Algorithm IDs [defined by TCG](https://trustedcomputinggroup.org/wp-content/uploads/TCG-Algorithm-Registry-Version-2.0_pub.pdf) today are within the range 0x0001 to 0x00B3

I don't think there are any situations where it is ambiguous, although I'll send a note to the workgroup to make sure this is/stays the case.

loicsikidi: Ok I'll implement something following this spec. Thanks for the hint!

tpm2/marshalling.go
loicsikidi: Thanks to [this comment](https://github.com/google/go-tpm/pull/415#discussion_r2491222656) I'm able to avoid this struct.

It have been removed :heavy_check_mark:

tpm2/marshalling.go
loicsikidi: Yes you are right, this design is more consistent.

tpm2/marshalling.go
```diff
@@ -126,3 +127,251 @@ func (b *boxed[T]) unmarshal(buf *bytes.Buffer) error {
...
+// MarshalResponse marshals a TPM response into its raw rpHash preimage format.
+// The returned bytes can be directly hashed to compute rpHash.
+//
+// Example:
+//
+//	rspData, _ := MarshalResponse(myCmd, myRsp)
+//	rpHash := sha256.Sum256(rspData)
+//
+// Note: Encrypted response parameters (via sessions) are not currently supported.
+func MarshalResponse[C Command[R, *R], R any](cmd C, rsp *R) ([]byte, error) {
```
loicsikidi, line 306: @chrisfenner 

I need to get the Command Code as an input. 

I think that it's more natural to pass the command but maybe that `TPMCC` is better.

WDYT?

chrisfenner: I think what you've written is perfect. Very natural. Thanks!

PR #434, opened 2026-05-20
policy: Implement Clone() for the PolicyCalculator

Foxboron:
When implementing nested policies, like systemd-pcrlock, it's practical to clone the current PolicyCalculator to transfer the state of the previous policy without mutation.

### DIFF
--- a/tpm2/policy.go
+++ b/tpm2/policy.go
@@ -27,6 +27,15 @@ func NewPolicyCalculator(alg TPMIAlgHash) (*PolicyCalculator, error) {
 	}, nil
 }
 
+// Clone copies the internal state of the policy to a new calculator
+func (p *PolicyCalculator) Clone() *PolicyCalculator {
+	return &PolicyCalculator{
+		alg:   p.alg,
+		hash:  p.hash,
+		state: bytes.Clone(p.state),
+	}
+}
+
 // Reset resets the internal state of the policy hash to all 0x00.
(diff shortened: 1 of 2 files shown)


tpm2/policy.go
```diff
@@ -27,6 +27,15 @@ func NewPolicyCalculator(alg TPMIAlgHash) (*PolicyCalculator, error) {
 	}, nil
 }
 
+// Clone copies the internal state of the policy to a new calculator
+func (p *PolicyCalculator) Clone() *PolicyCalculator {
+	return &PolicyCalculator{
+		alg:   p.alg,
+		hash:  p.hash,
+		state: p.state,
```
josephlr: I think this needs to be:
```suggestion
		state: bytes.Clone(p.state),
```
So that we ensure any state modifications to a Clone doesn't affect the original. 

`p.state` is usually re-assigned in the update methods, but the clone is cheap and makes things less brittle.

josephlr (changes requested):
I like this change, thank you.

Could we add a small test to `tpm2/test/policy_test.go` to check that `Clone` works as expected? I can be a very simple test-case.

Foxboron:
> Could we add a small test to tpm2/test/policy_test.go to check that Clone works as expected? I can be a very simple test-case.

Of course. I didn't spend time on a test because the change was simple and I was not completely sure if this was a good idea to begin with :)

PR #435, opened 2026-05-21
fix(tpm2): honor `gotpm:"check"` tag in the unmarshaller

evilgensec:
## Summary

`TPMSAttest.Magic` (tpm2/structures.go:1318) is declared with `gotpm:"check"`. The tag is meant to declare that the field's type implements `Check() error` and that the unmarshaller must reject a blob whose value fails the check. `TPMGenerated.Check()` (tpm2/structures.go:75-82) enforces `g == TPMGeneratedValue` (`0xff544347`).

`unmarshalStructField` in `tpm2/reflect.go` processes every other `gotpm:` tag (`skip`, `list`, `sized`, `sized8`, `tag`, `nullable`, `optional`, `handle`, `auth`, `bit`, `anon`) but never inspects `check`. Result: `TPMGenerated.Check()` has zero call sites anywhere in the package, the Magic field is the only field carrying the tag, and `Unmarshal[TPMSAttest]` happily returns a populated struct for any 4-byte Magic value.

A verifier following the pattern in `tpm2/test/certify_test.go` (lines 138-167) — call `.Contents()` to unmarshal the attest, marshal it back to compute the digest, verify the signature with the trusted AK, read `Attested.Quote().PCRDigest` and trust the result — therefore accepts forged "quotes" whose Magic is anything, as long as the surrounding signature verifies. The Magic field is precisely the boundary between "TPM-generated attestation" and "arbitrary bytes signed by any key the verifier trusts". Real-world signing oracles include non-restricted signing keys on the same TPM (TPM2_Sign with a non-restricted scheme does NOT prepend `TPM_GENERATED_VALUE`), vTPM-as-a-service back-ends, previously-leaked AKs reused for ad-hoc Sign, and cross-tenant signing endpoints.

Compare to the legacy package, which gets this right: `legacy/tpm2/structures.go:667-679` (`DecodeAttestationData`) explicitly rejects `ad.Magic != 0xff544347`.

## Fix

Wire the `check` tag into `unmarshalStructField`. After the field's bytes are read, if the field type implements `Check() error`, call it; return the error.

## Test plan

- [x] `go build ./tpm2/` clean.
- [x] New `tpm2/check_tag_test.go` adds a regression matrix for `Unmarshal[TPMSAttest]` that rejects `Magic=0x00000000`, `0xdeadbeef`, and the byte-swapped lookalike `0x47ff5443`, and that still accepts the legitimate `0xff544347`. The tests don't import the TPM simulator dep.
- Local PoC against `main` (commit `ab627a9`) before the fix: a hand-built 74-byte TPMSAttest with `Magic=0`, `Type=TPM_ST_ATTEST_QUOTE`, attacker-chosen 32-byte PCRDigest, attacker ECDSA signature over the blob verifies and the verifier reads `parsed.Attested.Quote().PCRDigest` as the attacker-chosen value. After this PR, `Unmarshal[TPMSAttest]` returns `check tag failed for field 'Magic' of struct 'TPMSAttest': TPM_GENERATED value should be 0xff544347, was 0x0` and the verifier short-circuits.

### DIFF
--- a/tpm2/check_tag_test.go
+++ b/tpm2/check_tag_test.go
@@ -1,0 +1,92 @@
+// Copyright 2026 The Go-TPM Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+
+package tpm2
+
+import (
+	"bytes"
+	"encoding/binary"
+	"strings"
+	"testing"
+)
+
+// buildAttestBlob constructs a minimal TPMSAttest blob with an attacker-chosen
+// Magic value. The remaining fields are populated with zeros so that the blob
+// is otherwise unmarshallable (TPMSAttest is a fixed-shape struct apart from
+// its Attested union, for which we select TPM_ST_ATTEST_QUOTE and supply an
+// empty PCRSelect + empty PCRDigest).
+func buildAttestBlob(t *testing.T, magic uint32) []byte {
+	t.Helper()
+	buf := new(bytes.Buffer)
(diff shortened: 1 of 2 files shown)


evilgensec:
Gentle bump on this one. Has anyone had a chance to look yet? Happy to rebase or adjust if that helps.

PR #436, opened 2026-05-28
Add testhelper.Open with TCP support

crocomegan:
This pull request adds TCP support to the simulator.

Summary of changes:
1. Adds flag support (`cmd_addr` and `plat_addr`) to specify command and platform TCP ports.
2. Updates `OpenSimulator()` to check for these flags. If both are provided, it initializes a TCP transport connection instead of using the default local simulator.


josephlr:
Here's the idea we discussed earlier about how to handle spawning the simulator per-test:

https://github.com/josephlr/go-tpm/commit/49ab7069baed67d1217207495c30ca5b5e997b94

It uses the `testhelper` package to simplify some things. It's just a rough idea, but running:
```bash
go test ./tpm2/test --tpm-sim-path ~/dev/tcg/TPM/build/Simulator/Simulator
```

Passes all the tests except for the ones we expect it to fail.

josephlr (approved):
Nice Job on this!

PR #444, opened 2026-08-04
Add support for Google EC-over-MTD transport

jhand2:
Many of Google's TPMs support the EC command protocol over a Linux MTD device. The RoT's mailbox is exposed as a flash device over MTD.

Add support for sending TPM commands via this EC-over-MTD interface. Additionally, refactor the generic EC code out of the ipmi package so that it can be reused between IPMI and MTD.


tpm2/transport/googleipmi/go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	github.com/vtolstov/go-ioctl v0.0.0-20151206205506-6be9cced4810 // indirect
 	golang.org/x/sys v0.41.0 // indirect
 )
+
+replace github.com/google/go-tpm => ../../../
```
jhand2: Looking at this closer, the other packages in transport don't have a go.mod file. I wonder why were creating a new module for googleipmi instead of using the root module. cc @anamendes23

tpm2/transport/googlemtd/mtd.go
```diff
@@ -0,0 +1,181 @@
...
+
+// Close closes the mtd connection to the Titan.
+func (d *Dispatcher) Close() error {
+	return d.mtd.Close()
+}
+
+// lock performs a file lock on the mtd file.
+func (d *Dispatcher) lock() error {
+	fd := int(d.mtd.Fd())
+	if err := syscall.Flock(fd, syscall.LOCK_EX); err != nil {
```
jhand2, line 119: Google mtd is only supported on Linux (mtd is a linux-specific construct). Need to look into whether I should just disable Windows CI for this package.

tpm2/transport/googleipmi/go.mod
anamendes23: The go.mod setup was requested by the maintainer, so that clients depending only on `googleipmi` don't have to import all other packages.

jhand2: Ah thanks, that makes sense. Updated googleec and googlemtd to also follow this pattern.

PR #445, opened 2026-09-17
Add StirRandom, MAC, MACStart and DictionaryAttackLockReset to tpm2

NiklasKoehler:
Four commands from Part 3 that the direct API did not have yet. The command codes were already in constants.go, so this is the structs, the Execute methods and tests, following the existing pattern.

- StirRandom (16.2)
- MAC (15.6) and MACStart (17.2.3), with a TPMIAlgMACScheme alias for TPMI_ALG_MAC_SCHEME (Part 2, 9.36). MAC gives CMAC on a symmetric signing key, the only MAC the TPM offers that is not an HMAC.
- DictionaryAttackLockReset (25.2)

Tests are in tpm2/test. The MAC test creates an AES-128 key with the RFC 4493 key value under an SRK and checks all four CMAC vectors from the RFC, then feeds the 64-byte message through MACStart, SequenceUpdate and SequenceComplete and compares with the single-shot result. StirRandom is checked for success on 32 bytes and rejection of 129 (MAX_SYM_DATA is 128). The lockout reset is a success check against the simulator.


tpm2/test/da_lock_reset_test.go
```diff
@@ -0,0 +1,23 @@
+package tpm2test
+
+import (
+	"testing"
+
+	. "github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpm2/transport/testhelper"
+)
+
+func TestDictionaryAttackLockReset(t *testing.T) {
```
chrisfenner, line 10: if it's not too much trouble, could this test deliberately put the TPM into lockout, reset it, then make sure the TPM isn't in lockout mode anymore?

chrisfenner (changes requested):
Thank you especially for including test vectors from RFC 4493, that's greatness. One small ask on testing DA lock just to be reasonably sure it's working. Other than that good to go. Thanks so much for the contribution!

PR #446, opened 2026-09-17
tpm2: trim only trailing zeros from auth values

NiklasKoehler:
hmacKeyFromAuthValue truncated the auth value at every zero byte it met while scanning backwards, so an auth value with a zero anywhere in it was cut down to the bytes before its first zero. HMAC sessions on such objects then failed with TPM_RC_AUTH_FAIL and bumped the DA counter, which for a random 32-byte auth happens about one time in eight.

Part 1, 19.6.5 only removes trailing zeros. Rewrite the loop to stop at the last non-zero byte and add a table test.


chrisfenner (approved):
Thanks for this fix!

PR #447, opened 2026-09-20
examples: format TPM handles as hexadecimal values

HNO3Miracle:
Format SRK and object handles as `0x%x` in the seal/unseal example's cleanup errors, matching the successful handle messages in the same example. `%q` formats these integer handles as quoted characters rather than TPM handle values.

Validation: `go test ./examples/tpm2-seal-unseal` with Go 1.26.3 (compilation and vet; this example has no tests). No TPM hardware test was run.


chrisfenner (approved):
Thanks for this cleanup!

